PR #23, opened 2012-12-04
Reuse connection when using requests library.

jpuderer:
When using the requests library, use a session to allow the connection
to be reused.  this greatly reduces overhead and latency when making
multiple requests.


kyleconroy:
This is still a good idea, but will need to wait for #96 to land. Closing in favor of that work.

PR #38, opened 2013-07-25
Custom request handler

amilstead:
Finding myself needing to implement a custom request handler for the stripe API, and I'd rather not monkey patch.

The infrastructure I work in requires that all IO be done in a very specific way, using a wrapped handler.

I can't imagine others wouldn't find this useful at some point.

Not to mention being able to apply different request handler functions from standard packages.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -29,7 +29,6 @@ logger = logging.getLogger('stripe')
 ## Configuration variables
 
 api_key = None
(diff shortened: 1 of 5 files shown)


amilstead:
@kyleconroy  I'll check into it when I get some time -- sorry it's been a while since I've looked at this PR. =)

kyleconroy:
Understandable! I'm going to close this PR until you get a chance to update it :)

PR #95, opened 2014-07-18
Configure GAE urlfetch timeout

### LINKED ISSUE #40: Stripe Timeout on App Engine Is Too Short When Being Called from TaskQueue
Currently, on Google App Engine, the stripe client times out if a request takes more than 55 seconds.  The reason for this (see code below), is because Google App Engine has a 60 second timeout for normal requests.  However, App Engine backends and TaskQueues are not subject to this deadline.

It is a common usage scenario to do a lot of heavy lifting for billing on a backend (e.g., running reports, generating invoices, etc.).  Often the 55 second timeout is too short and the request times out.  

Please increase this timeout, or better yet, detect when running in a TaskQueue and increase it only in that situation.  Thank you.

stripe.__init__.py:427
```
    # GAE requests time out after 60 seconds, so make sure we leave
    # some time for the application to handle a slow Stripe
    args['deadline'] = 55
```

kyleconroy:
Fix #40. To configure the deadline, set the `deadline` property on the UrlFetchClient.

```python
import stripe
import stripe.http_client

http_client.UrlFetchClient.deadline = 999

...
```

r? @metcalf @wangjohn

### DIFF
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -141,6 +141,9 @@ class RequestsClient(HTTPClient):
 
 
 class UrlFetchClient(HTTPClient):
+    # GAE requests time out after 60 seconds, so make sure we leave
+    # some time for the application to handle a slow Stripe
+    deadline = 55
     name = 'urlfetch'
 
     def request(self, method, url, headers, post_data=None):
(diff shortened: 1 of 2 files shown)


kyleconroy:
We never got to consensus here, closing

PR #118, opened 2015-01-12
[WIP] Unexpose BitcoinTransaction routes

maxwells:
Removed interface to query BitcoinTransactions -- only it should only be a child property of BitcoinReceiver


kyleconroy:
Sweet, looks good

PR #126, opened 2015-02-18
Add support for creating and listing accounts

kyleconroy:
I had to do some gross things here to fix nested updating. Instead of a complete refactor, I just took what we were doing for the metadata key and applied it to all keys. 

I need to make sure that this doesn't try to pass along keys for expanded parameters.


kyleconroy:
cc @bkrausz

bkrausz:
It's like christmas (both in terms of PRs and the red/green you got going there). LGTM!

PR #128, opened 2015-02-19
Add support for the Stripe-Account header

kyleconroy:
cc @bkrausz

### DIFF
--- a/stripe/test/test_requestor.py
+++ b/stripe/test/test_requestor.py
@@ -290,6 +290,24 @@ class APIRequestorRequestTests(StripeUnitTestCase):
         self.check_call('get', headers=APIHeaderMatcher(
             extra={'Stripe-Version': 'fooversion'}, request_method='get'))
 
+    def test_uses_instance_account(self):
+        account = 'acct_foo'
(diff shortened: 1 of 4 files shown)


kyleconroy:
Alright, ready for review! r? @bkrausz

stripe/test/test_requestor.py
```diff
@@ -290,6 +290,24 @@ def test_passes_api_version(self):
         self.check_call('get', headers=APIHeaderMatcher(
             extra={'Stripe-Version': 'fooversion'}, request_method='get'))
 
+    def test_uses_instance_account(self):
```
bkrausz: Maybe a test that these chain properly from an object to a related object?

bkrausz:
Generally LGTM. Mild FUD that we're adding another property in the internals instead of building an object, but given Python's named-param-friendly world I can't think of too strong an argument against that.

stripe/test/test_requestor.py
kyleconroy: Yep, will do.

kyleconroy:
I actually started out using an object instead of another property, but I found that this ended up being more straight forward.

PR #129, opened 2015-02-19
Remove sys import hack

kyleconroy:
I'm a bit worried that this change will mean we're just testing the latest version of stripe on PyPI.

### DIFF
--- a/stripe/test/test_integration.py
+++ b/stripe/test/test_integration.py
@@ -2,12 +2,9 @@
 import os
 import sys
 import unittest
-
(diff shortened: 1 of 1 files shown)


jpellerin:
:+1: 

tox installs the package in a virtualenv, there's no need for path hacks. If you're at all worried about it still, you could assert the import path and/or version in a test.

kyleconroy:
Thanks! I'm not super familiar with tox, so I wasn't sure. Thanks @jpellerin

PR #131, opened 2015-02-23
setup.py url should point to URL of package

frewsxcv:
https://docs.python.org/2/distutils/setupscript.html#additional-meta-data

"home page for the package"


kyleconroy:
Thanks!

PR #132, opened 2015-02-26
bitcoin receiver test consistency

maxwells:
made bitcoin receiver creation in test consistent with backend (email is actually required)

r? @kyleconroy


kyleconroy:
3rd times a charm :8ball:

PR #133, opened 2015-02-26
Basic cleanup

thoas:
Hi,

I'm have done some basic cleaning :)

* Removed unnecessary else clause when exiting the context (raising an
error, returning a value)
* Removed parenthesis on hasattr

### DIFF
--- a/stripe/api_requestor.py
+++ b/stripe/api_requestor.py
@@ -142,14 +142,16 @@ class APIRequestor(object):
         if rcode in [400, 404]:
             raise error.InvalidRequestError(
                 err.get('message'), err.get('param'), rbody, rcode, resp)
-        elif rcode == 401:
+
+        if rcode == 401:
             raise error.AuthenticationError(
                 err.get('message'), rbody, rcode, resp)
-        elif rcode == 402:
+
+        if rcode == 402:
             raise error.CardError(err.get('message'), err.get('param'),
                                   err.get('code'), rbody, rcode, resp)
-        else:
-            raise error.APIError(err.get('message'), rbody, rcode, resp)
+
+        raise error.APIError(err.get('message'), rbody, rcode, resp)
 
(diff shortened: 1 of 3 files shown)


danielchatfield:
Personally, I prefer the original. Whilst it is certainly true that in the following example the `elif` could just be an `if` the `elif` provides more information at very little cost, you can immediately tell that only one of those code paths can be taken without having to read all the other if statements.

```python
def handle_api_error(self, rbody, rcode, resp):
        try:
            err = resp['error']
        except (KeyError, TypeError):
            raise error.APIError(
                "Invalid response object from API: %r (HTTP response code "
                "was %d)" % (rbody, rcode),
                rbody, rcode, resp)

        if rcode in [400, 404]:
            raise error.InvalidRequestError(
                err.get('message'), err.get('param'), rbody, rcode, resp)
        elif rcode == 401:
            raise error.AuthenticationError(
                err.get('message'), rbody, rcode, resp)
        elif rcode == 402:
            raise error.CardError(err.get('message'), err.get('param'),
                                  err.get('code'), rbody, rcode, resp)
        else:
            raise error.APIError(err.get('message'), rbody, rcode, resp)
```

In addition to this, whilst I would make no great effort to try and change code of the form:
```python
def example():
   #...
    if some_condition:
        return 1
    return 4
```
into the form:
```python
def example():
    #...
    if some_condition:
        return 1
    else:
        return 4
```
I believe the second is better for the same reason. "return 4" can only be executed if the if clause is false and syntax that expresses this lets a reader infer this without having to read the body of the if clause is, in my opinion, preferable.

kyleconroy:
First off, @thoas thanks for taking the time to write this code and submit a pull request.

In this case, I'm going to agree with @danielchatfield and say that the changes you've made here are stylistic. Either form is fine, so I'm going to stick with the forms we currently have.

PR #136, opened 2015-03-22
Support arrays of StripeObjects

bkrausz:
r? @kyleconroy 

Because the iteration syntax between lists and dicts are different, I couldn't globally make the list-is-now-a-dict change. Hardcoding it to `additional_owners` for now until we can figure out a better solution.

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -38,6 +38,30 @@ def populate_headers(idempotency_key):
     return None
 
 
+def _compute_diff(current, previous):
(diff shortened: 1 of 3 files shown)


kyleconroy:
r? @bkrausz 

Some notes about this implementation: You can't delete a single item from the list. I believe that's what you intended. I'm going to add a few more tests, but I think this is good to go.
I'll also rebase this before I merge it in.

PR #137, opened 2015-03-25
Balance.retrieve support for managed accounts.

mrmicjam:
Makes sure the "stripe_account" parameter to the Balance.retrieve query gets passed all the way to the API call


kyleconroy:
Thanks!

PR #140, opened 2015-04-08
Fix saving managed accounts

### LINKED ISSUE #139: Managed Account .save() fails with TypeError
This problem occurs in Python 2.7.6, using either stripe package v1.22.0 or v1.22.1, and API Version 2015-04-07 (latest) .

When trying to `save()` a `<class 'stripe.resource.Account'>` Managed Account instance, I am met with the following error:

>>> acct = stripe.Account.retrieve('acct_[Managed Account ID]')
>>> acct.support_phone = "2013331111"
>>> acct.save()
  File "<input>", line 1, in <module>
  File "C:\Python27\Lib\site-packages\stripe\resource.py", line 344, in save
    updated_params = self.serialize(None)
  File "C:\Python27\Lib\site-packages\stripe\resource.py", line 223, in serialize
    params[k] = v.serialize(previous.get(k, None))
  File "C:\Python27\Lib\site-packages\stripe\resource.py", line 227, in serialize
    params[k] = _serialize_list(v, previous.get(k, None))
  File "C:\Python27\Lib\site-packages\stripe\resource.py", line 55, in _serialize_list
    for i, v in enumerate(array):
TypeError: 'NoneType' object is not iterable

I even tried running `account.save()` before updating any fields, and I still got this error. It looks like `None` is being passed in place of the array by default, and so the Managed Account instance is unable to be saved/updated/overwritten - which leads me to believe this is a bug.

kyleconroy:
r? @bkrausz 

Fixes #139

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -49,6 +49,7 @@ def _compute_diff(current, previous):
 
 
 def _serialize_list(array, previous):
+    array = array or []
     previous = previous or []
     params = {}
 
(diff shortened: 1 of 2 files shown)


bkrausz:
:+1: :sheep:

PR #141, opened 2015-04-18
Ultra verbose logging should be in DEBUG

xealot:
Log output of the entire HTTP response should be in DEBUG not in INFO. Putting everything in INFO gives me no ability to allow stripe to log without completely clobbering the logs in other ways.


kyleconroy:
@xealot +1 I think this is a great change. Not sure why the Travis build is failing. I'm rerunning the build, which should hopefully fix the issues.

xealot:
Thanks!

PR #142, opened 2015-04-20
Suggest adding contributors via script

xealot:
You can cancel or ignore this PR if it doesn't fit with your internal policies but I thought I would offer it anyway. Just an update to more accurately reflect the contributors of your project generated from the git history.


kyleconroy:
Sorry for leaving this hanging around! As you found out, the contributors file hasn't been updated in ages. Since it will always be out of date, I'd rather just use [the graph that GitHub provides](https://github.com/stripe/stripe-python/graphs/contributors).

PR #144, opened 2015-05-20
Add a BankAccount object

trenton42:
Currently, bank accounts that were added as a source to Customers cannot be deleted in the same way as Cards. This patch takes care of that.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -15,8 +15,9 @@ verify_ssl_certs = True
 
 # Resource
 from stripe.resource import (  # noqa
-    Account, Balance, BalanceTransaction, Card, Charge, Customer, Invoice,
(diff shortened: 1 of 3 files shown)


kyleconroy:
@trenton42 Thanks for the code. Could you also add a test for the accounts case? I'm thinking something like

```python
def test_customer_delete_bank_account(self):
    source = stripe.BankAccount.construct_from({
        'account': 'acct_123',
        'id': 'ba_source',
    }, 'api_key')
    source.reload()

    self.requestor_mock.request.assert_called_with(
        'get',
        '/v1/accounts/acct_123/bank_accounts/ba_source',
        {},
        None
    )
```

trenton42:
@kyleconroy Certainly! I missed that last night :disappointed:

trenton42:
@kyleconroy All set. Poor travis seems to be overwhelmed recently.

kyleconroy:
@trenton42 thanks!

Issue #151, opened 2015-06-13
saving BankAccount object after failed verify call results in InvalidRequestError


catskul:
Attempting to save a BankAccount object after attempting validation results in `InvalidRequestError`.

```
In [20]: account.verify(amounts=[1,1])

...InvalidRequestError: You have tried to verify this bank account 21 times. To continue trying to verify the account, please reach out to us directly.

In [21]: account.save()

InvalidRequestError: A parameter provided in the URL (customer) was repeated as a GET or POST parameter. You can only provide this information as a portion of the URL.
```

Seems to be related to this old bug and fix in Ruby:

https://github.com/stripe/stripe-ruby/issues/52
https://github.com/stripe/stripe-ruby/commit/6a57264758edcec12fa23e9a290e9fec3d2d50eb

I'm able to get around it by using the same workaround listed in linked ruby issue: Namely `thing.refresh()` before `thing.save()`

I'm guessing the `verify` call adds an attribute to the object which is not handled by the `if` chain in `serialize`

catskul:
experiment implies it's a unicode problem on the `customer` and `id` fields:

```
In [60]: before_items
Out[60]: 
[(u'status', u'new'),
 (u'routing_number', u'110000000'),
 (u'bank_name', u'STRIPE TEST BANK'),
 (u'country', u'US'),
 (u'object', u'bank_account'),
 (u'currency', u'usd'),
 (u'fingerprint', u'hL63yGLH3uPk1iVr'),
 (u'last4', u'1116'),
 (u'customer', 'cus_6Mp7N1PVjCgzvS'),
 (u'id', 'ba_16CdgTBGLd1HVBcija3Py1Ug'),
 (u'metadata',
  <StripeObject at 0x7f7234117740> JSON: {
  "error_message": "You have tried to verify this bank account 17 times. To continue trying to verify the account, please reach out to us directly.", 
  "score": "3"
})]

In [61]: after_items
Out[61]: 
[(u'status', u'new'),
 (u'routing_number', u'110000000'),
 (u'bank_name', u'STRIPE TEST BANK'),
 (u'country', u'US'),
 (u'object', u'bank_account'),
 (u'currency', u'usd'),
 (u'fingerprint', u'hL63yGLH3uPk1iVr'),
 (u'last4', u'1116'),
 (u'customer', u'cus_6Mp7N1PVjCgzvS'),
 (u'id', u'ba_16CdgTBGLd1HVBcija3Py1Ug'),
[...]

PR #152, opened 2015-06-13
Fix mutation side effect in Card and BankAccount

catskul:
fix accidental 'self' reference causing mutation side effect on methods meant to be a functions

was causing `InvalidRequestError: A parameter provided in the URL (customer) was repeated as a GET or POST parameter. You can only provide this information as a portion of the URL.`

fixes: https://github.com/stripe/stripe-python/issues/151


kyleconroy:
@catskul great catch!

PR #156, opened 2015-06-19
Payment and BankAccount support improvements

catskul:
* Add in Payment stub. 
* Add BankAcccount and Payment to class lookup
* Add verification method to BankAccount


kyleconroy:
@catskul thanks! Any chance we can squash these commits into a manageable number? I'd prefer having 1-4 instead of 17. Thanks!

catskul:
I played with it a bit, but had problems with interactive rebase with the merge commits.  My git foo isn't strong enough to squash this correctly

jackdent:
@catskul I'd suggest running `git reset --soft cddb9acca5fd4ecdeacf72b259ff17cdaa2dc69d && git reset .`. This will remove all of your commits, and unstage your changes. If you run `git status` at this point, you'll be able to see the files that have changes. You can then run `git commit -p` to create a new commit, and interactively pick which changes you want to add. Run this command for as many commits as you think you need (I would suggest one commit for the Payment resource, one for the verify method, and one for styling). Run `git status` again to check there are no further changes to commit, and then `git push -f origin HEAD` to push the changes up to your repo.

catskul:
Unfortunately I think that would destroy authorship on Lenny Pratt's commits. I'll see if I can find another way.

jackdent:
Oops, didn't realise there were multiple committers. My bad.

catskul:
32c89da and bbcefef look to be identical and an artifact of what bitbuckets PR system. I've seen this before, and after the merge, the commits that were duplicated disappeared strangely.

In any case I don't think the merge commits can be squashed easily as there were merge conflicts that had to happen in some of the cases. It may be doomed to be a messy merge (though slightly less messy than it seems here)

kyleconroy:
@catskul Thanks again for the PR, but since it no longer merges, I went ahead and added bank account verification. Also the `payment` type has been deprecated, so I didn't add a `Payment` class. Sorry for the run-around here, but you should now be able to verify bank accounts!

PR #158, opened 2015-07-16
Added support for verify bank account endpoint

jackdent:
Covered in https://github.com/stripe/stripe-python/pull/156?

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -465,6 +465,12 @@ class BankAccount(UpdateableAPIResource, DeletableAPIResource):
 
         return "%s/%s/%s/%s" % (base, owner_extn, class_base, extn)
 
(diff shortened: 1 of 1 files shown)


kyleconroy:
So this and #156 are different. This `verify` function will make calls to the following endpoints:

```
/v1/accounts/acct_xxx/external_sources/src_xxx/verify
/v1/customers/cust_xxx/sources/src_xxx/verify
```

#156 will only work if the bank account is attached to a customer, and the URL is different.

```python
/v1/customers/cust_xxx/bank_accounts/src_xxx/verify
```

Which one is correct?

jackdent:
We have scheduled an API review session for later today to discuss changes to how bank accounts are handled. I'll update the PRs based on the result of that conversation.

kyleconroy:
:+1:

PR #159, opened 2015-07-16
Add top-level dispute methods.

cupcait:


cupcait:
r? @metcalf cc @kyleconroy 
(note: I don't have a huge amount of experience in python, so feel free to nitpick!)

stripe/test/test_resources.py
```diff
@@ -823,6 +823,66 @@ def test_create_with_source_param(self):
...
+        stripe.Dispute.retrieve('dp_test_id')
+
+        self.requestor_mock.request.assert_called_with(
+            'get',
+            '/v1/disputes/dp_test_id',
+            {},
+            None
+        )
+
+    def test_update_dispute(self):
```
metcalf: Updating with evidence seems like it would be the more common case, might be better to test that instead?

bin/stripe
```diff
@@ -159,6 +159,14 @@ class CustomerClient(CreateableAPIResourceClient, UpdateableAPIResourceClient,
     url = self.instance_url() + '/subscription'
     self.logged_curl('delete', url, params)
 
+class DisputeClient(CreateableAPIResourceClient, ListableAPIResourceClient,
```
metcalf: This file is deprecated so I'd be inclined to just stop adding new things to it unless @kyleconroy has strong feels otherwise.

stripe/resource.py
```diff
@@ -525,6 +525,15 @@ def mark_as_safe(self, idempotency_key=None):
         return self
 
 
+class Dispute(CreateableAPIResource, ListableAPIResource,
+              UpdateableAPIResource):
+    def close(self, idempotency_key=None):
```
metcalf: Other classes in the file have a line break before the first method. Seems like PEP8 doesn't care.

metcalf:
Made a few minor comments but generally looks good to me!

stripe/resource.py
```diff
@@ -17,7 +17,7 @@ def convert_to_stripe_object(resp, api_key, account):
              'fee_refund': ApplicationFeeRefund,
              'bitcoin_receiver': BitcoinReceiver,
              'bitcoin_transaction': BitcoinTransaction,
-             'transfer_reversal': Reversal}
+             'transfer_reversal': Reversal, 'dispute': Dispute}
```
metcalf: You asked me to turn on nit-picky mode... I wish this list and all the classes in this file were sorted in the same order (preferably alphabetical). That's my fault for not doing it in the first place though, not worth polluting this diff.

bin/stripe
kyleconroy: I have strong feels. Strong feels that we should stop updating this file :)

PR #163, opened 2015-07-29
api endpoint for subscription is updated.

goldsoft1206:
While I test multiple subscription for a customer, I found that subscription endpoint is not correct, so I'm updating it. 
Please refer this document - https://stripe.com/docs/api/curl#create_subscription


kyleconroy:
The `update_subscription` and `cancel_subscription` are deprecated methods that existed before the `/subscrptions` endpoint. Instead, you should just be using the `subscriptions` resource on the `customer` object.

I've now added depreciation warnings 9fae10aa153c477c8ce50d53dd18077d3dd9d6a0

PR #169, opened 2015-08-24
Allow for the iteration of ListObjects.

mastermatt:
More pythonic `for .. in` loops can now
loop the the contents of the `data` list
and continue on using the cursor-based
pagination model.

```python
for charge in stripe.Charge.all():
    my_func(charge)
```

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -316,6 +316,22 @@ class ListObject(StripeObject):
 
         return self.request('get', url, params)
 
+    def __iter__(self):
+        page = self
+        params = dict(self._retrieve_params)  # copy so not to mutate originals
+
+        while True:
+            last_id = None
+            for item in page.data:
+                last_id = item.get('id')
+                yield item
+
+            if not page.has_more or not last_id:
+                return
+
+            params['starting_after'] = last_id
+            page = self.all(**params)
+
 
 class SingletonAPIResource(APIResource):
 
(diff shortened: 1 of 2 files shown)


stripe/resource.py
```diff
@@ -316,6 +316,22 @@ def retrieve(self, id, **params):
...
         return self.request('get', url, params)
 
+    def __iter__(self):
+        page = self
+        params = dict(self._retrieve_params)  # copy so not to mutate originals
+
+        while True:
+            last_id = None
+            for item in page.data:
+                last_id = item.get('id')
```
kyleconroy: We can't assume that all list objects will have an `id`. To be extra safe, we should use `.get('id', None)`

kyleconroy: Whoops, ignore me :)

stripe/test/test_resources.py
```diff
@@ -208,6 +208,60 @@ def test_retrieve(self):
         self.assertResponse(res)
 
 
+class ListObjectTests(StripeApiTestCase):
```
kyleconroy: We'll need to add tests for edge cases. Specifically, we need to cover the case when `has_more` is `True`. We also should test the case when we fetch an additional page and no results are returned.

mastermatt: I didn't bother to go down that road until I received feedback. I wanted to ensure the core devs were ok with the __iter__ recalling the endpoint. It could be reasoned that iterating the list object should only go through the single page.

kyleconroy: Another option would be to add an `iter()` method to list objects, making it clearer that it will return all objects. Thoughts?

kyleconroy:
We may not want to make the pagination on by default. We could do something like this?

```python
stripe.auto_paginate = True

for charge in stripe.Charge.all(limit=3 paginate=True):
    print(charge)
```

Issue #165, opened 2015-08-11
`StripeError` should implement __unicode__


alexmic:
It seems that the `StripeError` class expects error messages to be in ASCII which fails when formatting errors with unicode characters like "Amount must be at least 300 öre". 

Test case:

```python
>>> from stripe.error import StripeError
>>> e = StripeError(u'Amount must be at least 300 \xf6re')
>>> unicode(e)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/opt/virtualenv/tictail/order-service/local/lib/python2.7/site-packages/stripe/error.py", line 22, in __str__
    msg = super(StripeError, self).__str__()
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 28: ordinal not in range(128)
>>> unicode(e, 'utf-8')
Traceback (most recent call last):
  File "<console>", line 1, in <module>
TypeError: coercing to Unicode: need string or buffer, StripeError found
>>> str(e)
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/opt/virtualenv/tictail/order-service/local/lib/python2.7/site-packages/stripe/error.py", line 22, in __str__
    msg = super(StripeError, self).__str__()
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf6' in position 28: ordinal not in range(128)
```

PR #176, opened 2015-09-17
Allow unicode error messages in `StripeError`

alexmic:
This PR fixes https://github.com/stripe/stripe-python/issues/165.

### DIFF
--- a/stripe/error.py
+++ b/stripe/error.py
@@ -1,4 +1,7 @@
 # Exceptions
+import sys
+
+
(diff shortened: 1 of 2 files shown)


kyleconroy:
Thanks for the PR. I'd like to factor out the unicode fix into a mixin ala http://lucumr.pocoo.org/2011/1/22/forwards-compatible-python/#recursion-error-with-str. Could you do that and rebase into one commit? Thanks

PR #179, opened 2015-09-30
Add bank account verification.

kyleconroy:
You can now verify a bank account. Let the API tell us if a source can't be verified.


kyleconroy:
Supercedes https://github.com/stripe/stripe-python/pull/156

PR #180, opened 2015-09-30
Configurable default HTTP client

kyleconroy:
You can now override the HTTP client that Stripe uses with your own client. Note that this client must implement the `HTTPClient` interface defined in `http_client.py`.

### DIFF
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -144,6 +144,12 @@ class RequestsClient(HTTPClient):
 class UrlFetchClient(HTTPClient):
     name = 'urlfetch'
 
+    def __init__(self, verify_ssl_certs=True, deadline=55):
+        self._verify_ssl_certs = verify_ssl_certs
+        # GAE requests time out after 60 seconds, so make sure to default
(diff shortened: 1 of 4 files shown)


kyleconroy:
r? @brandur 

I know Python isn't your strong suit, but wanted to run this by someone.

stripe/http_client.py
```diff
@@ -144,6 +144,12 @@ def _handle_request_error(self, e):
 class UrlFetchClient(HTTPClient):
     name = 'urlfetch'
 
+    def __init__(self, verify_ssl_certs=True, deadline=55):
```
brandur: Do you think it might be worth just putting `deadline` in the existing constructor for `HTTPClient`? Avoids duplicating the `verify_ssl_certs` setting logic which is kind of nice.

brandur:
@kyleconroy Left a minor comment, but looks good to me!

stripe/http_client.py
kyleconroy: (discussed IRL, but deadline is specific to App engine urlfetch)

PR #183, opened 2015-10-09
Auto pagination

kyleconroy:
Supersedes #169

This pull request is a WIP.


brandur:
Holy crap, did you add like 500 new tests? That's a new level of awesome.

Looks great so far; let us know when it's ready for review.

kyleconroy:
Lol, I wish I had added all these tests. These tests were all in `test_resources.py` before, so I split them out into separate files.

kyleconroy:
cc @brandur

stripe/test/resources/test_list_object.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def test_class_method_two_pages(self):
+        self.mock_response({
+            'object': 'list',
+            'data': [{'id': 'bar'}],
+            'url': '/v1/charges',
+            'has_more': False,
+        })
+
+        seen = [item['id'] for item in stripe.Charge.paging_iter(limit=25)]
```
brandur: +1 doing a test right away with parameter checks!

One thing maybe worth mentioning: on my first go I ended up special casing `limit` in particular, which ended up allowing other types of parameters (i.e. expansions and filter conditions) to fall through. It doesn't look like this code is in danger of that, but may be worth considering just for some extra certainty.

stripe/resource.py
```diff
@@ -312,9 +312,32 @@ def instance_url(self):
...
         return self.request('get', self['url'], params)
 
+    def all(self, **params):
+        warnings.warn("The `all` method is deprecated and will"
+                      "be removed in future versions. Please use the "
+                      "`list` method instead",
+                      DeprecationWarning)
+        return self.list(**params)
+
+    def paging_iter(self):
```
brandur: Bikeshed: I think somebody at one point pitched the idea of this being named `auto_paging_iter` (@russelldavis?). That might be worth considering just for the sake of consistency with the Ruby bindings.

stripe/resource.py
```diff
@@ -312,9 +312,32 @@ def instance_url(self):
 
 class ListObject(StripeObject):
 
-    def all(self, **params):
+    def list(self, **params):
```
brandur: Hah, almost a little comical how closely this naming schema matches Ruby :) Not a bad thing though!

brandur:
@kyleconroy Left a couple minor comments, but this looks great to me! :ship:

stripe/resource.py
russelldavis: yeah, +1 for consistency

kyleconroy: K, works for me

jpellerin:
(I'm ignoring the tests, since I don't know what's really new there.) Overall I think this looks fine. In isolation I'd probably do something else (either a decorator or a context manager or a thing that operates as both) to wrap the list call in paging, but I'm not sure of the value of that vs consistency with the ruby version. I'm not sure how many people use both, or move from one to the other? So much bikeshedding to be done here. :) :shipit:

kyleconroy:
@jpellerin Can you give some examples of what your alternative interfaces would look like?

```python
with stripe.paging(auto=True):
    for ch in stripe.Charge.list('foo'):
        print(ch)
```

jpellerin:
Yeah, something like that, or `with stripe.paging(pagesize=...):` (I think the auto is implied).

kyleconroy:
I'm not sure that we want to page size to be picked by the client, as they wouldn't know what page size to pick. I'm leaning towards merging this PR with `auto_paging_iter`. We can revisit and provide a different (maybe better) interface at a later date.

PR #185, opened 2015-11-05
Add legacy rate limit error handling

beedub:
The stripe api pre-2015-09-18 returned rate limit errors as 400 with code 'rate_limit'. This change adds support for that legacy behavior.


kyleconroy:
@beedub Thanks! If the tests pass, I'll go ahead and merge.

PR #195, opened 2015-12-16
Make calls to `instance_url` side-effect free.

kyleconroy:
Calling instance_url shouldn't change the state of an object.
Previously, we were setting properties in this method, which lead to
some very strange failure cases. No more!

If we called `save` and the request failed with a 400, the object would
be left in an incossitent state. Since we set attributes when calling
instance_url, these values would be counted as unsaved attributes the
next time save was called.


kyleconroy:
cc @stripe/dev-platform

tetrakai:
Looks good to me :+1:

brandur:
Nice one Kyle!

PR #196, opened 2015-12-19
Handle the case where error messages are None.

### LINKED ISSUE #194: TypeError: coercing to Unicode: need string or buffer, NoneType found
If `self._message` is `None` printing `StripeError` raises `TypeError`. See https://github.com/stripe/stripe-python/blob/master/stripe/error.py#L34.

kyleconroy:
Previously this would have resulted in a TypeError.

Fix #194

### DIFF
--- a/stripe/error.py
+++ b/stripe/error.py
@@ -24,7 +24,8 @@ class StripeError(Exception):
 
     def __unicode__(self):
         if self.request_id is not None:
-            return u"Request {0}: {1}".format(self.request_id, self._message)
+            msg = self._message or "<empty message>"
+            return u"Request {0}: {1}".format(self.request_id, msg)
         else:
             return self._message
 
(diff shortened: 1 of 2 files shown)


stripe/error.py
```diff
@@ -24,7 +24,8 @@ def __init__(self, message=None, http_body=None, http_status=None,
 
     def __unicode__(self):
         if self.request_id is not None:
-            return u"Request {0}: {1}".format(self.request_id, self._message)
+            msg = self._message or ""
```
brandur: Okay, so I'm not sure under exactly which conditions we can expect an empty message, but do you thing it would make sense to use a symbolic stand-in instead of an empty string here? e.g. `"<empty message>"` It might make the error more digestible when encountered.

brandur:
Left a minor comment, but +1. Thanks for patching this!

stripe/error.py
kyleconroy: Yeah, good point.

PR #198, opened 2015-12-24
Simpler if/else

Bekt:
Not a big deal, but more pythonic :+1:


brandur:
Looks good to me. Thanks!

PR #199, opened 2015-12-26
Fixed flake8 errors

olivierbellone:
PR #196 introduced two flake8 errors:

    stripe/error.py:33:9: E731 do not assign a lambda expression, use a def
    stripe/error.py:35:9: E731 do not assign a lambda expression, use a def

This PR changes the code to be flake8-compliant.


brandur:
Looks good to me! It doesn't look like the lambdas have a huge advantage here anyway.

Thanks Olivier!

PR #200, opened 2016-01-05
Catch exceptions with modern syntax.

ticosax:
Because it improves reading of source code.


brandur:
Cool. [Looks like we'd already started using this to some degree anyway](https://github.com/stripe/stripe-python/blob/7c089c7bb24edc96f435936038423c3c8547c490/stripe/test/helper.py#L145), so it makes sense to make this consistent. Thanks!

PR #209, opened 2016-02-26
Add CountrySpec

ngzhian:

PR #210, opened 2016-02-26
Add support for country specs

olivierbellone:
r? @brandur 
cc @stripe/api-libraries 

Tests are failing at the moment because the test account for the bindings is being purged, but they should pass once that's done.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -25,6 +25,7 @@ from stripe.resource import (  # noqa
     BitcoinTransaction,
     Card,
     Charge,
+    CountrySpec,
     Coupon,
     Customer,
     Dispute,
--- a/stripe/test/resources/test_country_specs.py
+++ b/stripe/test/resources/test_country_specs.py
@@ -1,0 +1,23 @@
+import stripe
+from stripe.test.helper import StripeResourceTest
+
+
+class CountrySpecTest(StripeResourceTest):
+
+    def test_country_spec_list(self):
+        stripe.CountrySpec.list()
+        self.requestor_mock.request.assert_called_with(
+            'get',
+            '/v1/country_specs',
+            {}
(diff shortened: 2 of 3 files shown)


olivierbellone:
OMG... Just saw #209.

Welp. Pick whichever you like best :D

stripe/__init__.py
```diff
@@ -26,6 +26,7 @@
     Card,
     Charge,
     Coupon,
+    CountrySpec,
```
brandur: Mind just re-ordering `CountrySpec` to be above `Coupon` here? Thanks!

stripe/test/resources/test_country_specs.py
```diff
@@ -0,0 +1,23 @@
+import stripe
+from stripe.test.helper import StripeResourceTest
+
+
+class CountrySpecTest(StripeResourceTest):
+
+    def test_retrieve_country_spec(self):
```
brandur: Mind just doing two minor tweaks here?

1. Re-order the naming to look like our other specs in the form of `test_<resource>_<action>`. Actually, see [the competing pull for an example](https://github.com/stripe/stripe-python/pull/209/files) :)
2. Put `list_all` in above `retrieve`. This seems to be more consistent with other suites.

brandur:
Haha, crazy timing!

Just because I'll be asking for a few minor amendments either way, let's use yours given that you're on the clock here :)

olivierbellone:
Done!

Looks like the test account is still being purged... There must have been a lot of data!

brandur:
> Looks like the test account is still being purged... There must have been a lot of data!

Hah, yep. We're cautiously optimistic that it'll be done sometime in 2016 ;)

Anyway, changes look good to me! Thanks @olivierbellone!

PR #207, opened 2016-02-23
Use latest mock version

olivierbellone:
Fixes #189 and #205.

This PR updates `setup.py` to use the latest mock version.

It also updates two tests (`ChargeRefundTest.test_non_recursive_save()` and `UpdateableAPIResourceTests.test_save_nothing()`) to reflect the current behavior of the library (see #205 for more context on this).

r? @brandur
cc @stripe/api-libraries

PR #211, opened 2016-02-27
Use latest mock version

olivierbellone:
Missed some files in #207.

r? @brandur


brandur:
Nice! Thanks for looping back and taking care of this.

PR #214, opened 2016-03-14
Add bindings for account.reject

gregsabo:
Adds bindings for the [Account.reject](https://stripe.com/docs/api#reject_account) endpoint.

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -450,9 +450,8 @@ class Account(CreateableAPIResource, ListableAPIResource,
     def reject(self, reason=None, idempotency_key=None):
         url = self.instance_url() + '/reject'
         headers = populate_headers(idempotency_key)
-        params = dict(reason=reason)
         self.refresh_from(
-            self.request('post', url, params, headers)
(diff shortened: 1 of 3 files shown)


stripe/resource.py
```diff
@@ -447,6 +447,15 @@ def instance_url(self):
         extn = urllib.quote_plus(id)
         return "%s/%s" % (base, extn)
 
+    def reject(self, reason=None, idempotency_key=None):
+        url = self.instance_url() + '/reject'
+        headers = populate_headers(idempotency_key)
+        params = dict(reason=reason)
+        self.refresh_from(
+            self.request('post', url, params, headers)
```
kyleconroy: Instead of creating the `params` variable, just create a dictionary like so

```python
self.request('post', url, {"reason": reason}, headers)
```

kyleconroy:
Can we add a test for calling reject without a reason?

gregsabo:
@kyleconroy how are things looking now?

kyleconroy:
:+1:

brandur:
Released as 1.31.0.

PR #215, opened 2016-03-23
Add "CyberTrust Global Root"

brandur:
This allows old OpenSSL versions (0.9.8 and 1.0.0) to successfully
verify Stripe's certificate chain.


brandur:
Released as 1.32.0.

PR #216, opened 2016-03-23
Don't check for request ID on invalid credentials

brandur:
It seems that some middleware tweaks inside the API server have resulted
in a request ID no longer being returned when invalid credentials are
passed in. This patch removes the check given that it's currently
failing our test suite.

~~/cc @kyleconroy Would you mind reviewing this one? Thanks!~~


brandur:
Actually, thinking about this more, it's such a tiny tweak that I'm just going to merge it so that I can get CI on other branches fixed. Kyle, you're off the hook!

kyleconroy:
I'm 100% okay with this code being removed.

PR #218, opened 2016-03-24
Fix ASCII/Unicode handling in Python 2 and 3

### LINKED ISSUE #217: Problem with uploading binary files with Python 3
With Python 3, uploading a binary file will fail:

```python
with open('/path/to/a/file.jpg', 'r') as fp:
    file_upload = stripe.FileUpload.create(
        purpose='dispute_evidence',
        file=fp
    )
```
fails with:
```
  <snip>
  File "/Users/ob/.pyenv/versions/3.5.1/lib/python3.5/site-packages/stripe/multipart_data_generator.py", line 28, in add_params
    self._write_file(value)
  File "/Users/ob/.pyenv/versions/3.5.1/lib/python3.5/site-packages/stripe/multipart_data_generator.py", line 54, in _write_file
    file_contents = f.read(self.chunk_size)
  File "/Users/ob/.pyenv/versions/3.5.1/lib/python3.5/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte
```

Opening the file as binary (e.g. replacing `'r'` by `'rb'` yields a different issue:
```
  <snip>
  File "/Users/ob/.pyenv/versions/3.5.1/lib/python3.5/site-packages/stripe/multipart_data_generator.py", line 28, in add_params
    self._write_file(value)
  File "/Users/ob/.pyenv/versions/3.5.1/lib/python3.5/site-packages/stripe/multipart_data_generator.py", line 57, in _write_file
    self._write(file_contents)
  File "/Users/ob/.pyenv/versions/3.5.1/lib/python3.5/site-packages/stripe/multipart_data_generator.py", line 50, in _write
    self.data.write(bytes(value, 'utf-8'))
TypeError: encoding or errors without a string argument
```

I think we ought to:

1. Update the Python file upload examples to use `mode='rb'` rather than `mode='r'` (since image files are binary files and it makes no sense to apply an encoding to them).

2. Patch [this method](https://github.com/stripe/stripe-python/blob/master/stripe/multipart_data_generator.py#L46-L50) that assumes that Python 3 will always deal with Unicode strings. Something like this should work:
```python
def _write(self, value):
  if sys.version_info < (3, 0) or isinstance(value, bytes):
    self.data.write(value)
  else:
    self.data.write(bytes(value, 'utf-8'))
```

@brandur, any thoughts? Should I proceed with the above?

olivierbellone:
Fixes #217.

This PR ensures that `MultipartDataGenerator` can handle keys and values of all encodings in both Python 2 and 3.

### DIFF
--- a/stripe/multipart_data_generator.py
+++ b/stripe/multipart_data_generator.py
@@ -18,8 +18,11 @@ class MultipartDataGenerator(object):
             self._write(self.param_header())
             self._write(self.line_break)
             if hasattr(value, 'read'):
-                self._write("Content-Disposition: form-data; name=\"%s\"; "
-                            "filename=\"%s\"" % (key, value.name))
+                self._write("Content-Disposition: form-data; name=\"")
+                self._write(key)
+                self._write("\"; filename=\"")
+                self._write(value.name)
+                self._write("\"")
                 self._write(self.line_break)
                 self._write("Content-Type: application/octet-stream")
                 self._write(self.line_break)
@@ -27,8 +30,9 @@ class MultipartDataGenerator(object):
 
                 self._write_file(value)
             else:
-                self._write("Content-Disposition: form-data; name=\"%s\"" %
-                            (key,))
+                self._write("Content-Disposition: form-data; name=\"")
+                self._write(key)
+                self._write("\"")
                 self._write(self.line_break)
                 self._write(self.line_break)
                 self._write(value)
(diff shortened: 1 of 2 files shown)


stripe/multipart_data_generator.py
```diff
@@ -18,8 +18,11 @@ def add_params(self, params):
             self._write(self.param_header())
             self._write(self.line_break)
             if hasattr(value, 'read'):
-                self._write("Content-Disposition: form-data; name=\"%s\"; "
-                            "filename=\"%s\"" % (key, value.name))
+                self._write("Content-Disposition: form-data; name=\"")
+                self._write(key)
+                self._write("\"; filename=\"")
+                self._write(value.name)
+                self._write("\"")
```
olivierbellone: I changed the code to not rely on Python's string formatting to handle the (admittedly improbable) case where the key is a byte array.

brandur:
Amazing work @olivierbellone! Let's get this out.

brandur:
Released as 1.31.1.

PR #222, opened 2016-04-11
Persist parameters with auto_paging_iter()

### LINKED ISSUE #221: auto_paging_iter() does not persist params
From the documentation at https://stripe.com/docs/api/python#auto_pagination:

> To use the auto-pagination feature in Python, simply issue an initial "list" call with the parameters you need, then call auto_paging_iter() on the returned list object to iterate over all objects matching your initial parameters.

With this code example:
```python
customers = stripe.Customer.all(limit=3)
for customer in customers.auto_paging_iter():
  # Do something with customer
```

However using this code the params that are passed in initially do not get saved here: https://github.com/stripe/stripe-python/blob/efe1ecc575ad97d22ad6381ee1f6686f34fc27b1/stripe/resource.py#L391
so the next trip through the params will be empty: https://github.com/stripe/stripe-python/blob/efe1ecc575ad97d22ad6381ee1f6686f34fc27b1/stripe/resource.py#L328

olivierbellone:
This PR fixes #221: `auto_paging_iter()` should now persist parameters.

r? @brandur 
cc @stripe/api-libraries

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -393,7 +393,10 @@ class ListableAPIResource(APIResource):
         requestor = api_requestor.APIRequestor(api_key, account=stripe_account)
         url = cls.class_url()
         response, api_key = requestor.request('get', url, params)
-        return convert_to_stripe_object(response, api_key, stripe_account)
+        stripe_object = convert_to_stripe_object(response, api_key,
+                                                 stripe_account)
+        stripe_object._retrieve_params = params
+        return stripe_object
 
 
 class CreateableAPIResource(APIResource):
--- a/stripe/test/resources/test_list_object.py
+++ b/stripe/test/resources/test_list_object.py
@@ -57,52 +57,67 @@ class ListObjectTests(StripeApiTestCase):
 
 class AutoPagingTests(StripeApiTestCase):
 
-    def test_iter_one_page(self):
-        lo = stripe.resource.ListObject.construct_from({
+    @staticmethod
+    def pageable_model_response(ids, has_more):
+        return {
             'object': 'list',
-            'url': '/my/path',
-            'data': [{'id': 'foo'}],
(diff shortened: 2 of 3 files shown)


stripe/resource.py
```diff
@@ -393,7 +393,9 @@ def list(cls, api_key=None, idempotency_key=None,
         requestor = api_requestor.APIRequestor(api_key, account=stripe_account)
         url = cls.class_url()
         response, api_key = requestor.request('get', url, params)
-        return convert_to_stripe_object(response, api_key, stripe_account)
+        stripe_object = convert_to_stripe_object(response, api_key, stripe_account)
+        stripe_object._retrieve_params = params
```
kyleconroy: While I'm never a fan of assigning to "private" instance variables, I believe this is the best way to fix the current bug.

brandur: Yeah, I was wondering about this one too.

Here's what PEP-8 has to say about the leading underscore:

> _single_leading_underscore: weak "internal use" indicator. E.g. from M import * does not import objects whose name starts with an underscore.

In this case, `_retrieve_params` should be considered a property of `StripeObject`. Because `ListObject` (a descendant of `StripeObject`) already accesses it, we could consider its visibility to be currently "protected" (based on the definition from other OO languages). `ListableAPIResource` is also a descendant of `StripeObject` so we're not really modifying that visibility from what it was before at all.

The idea of protected visibility is also in-line with PEP-8's idea of internal use, so I think we're okay on this here.

brandur:
@olivierbellone Thanks a lot for tackling this fix man! Also, you measurably improved the quality of those tests, so thank-you for taking a look at that too.

Going to get this shipped!

brandur:
Released as 1.32.1.

PR #223, opened 2016-04-12
Fixes FileUpload.list()

olivierbellone:
`FileUpload.list()` (or `.all()`)  fails because it tries to hit `https://api.stripe.com` instead of `https://uploads.stripe.com`. This PR fixes that.

r? @brandur 
cc @stripe/api-libraries


brandur:
Man @olivierbellone, you're on a roll!

+1.

brandur:
Released as 1.32.2.

PR #224, opened 2016-04-12
Allow creation and retrieval of subs, use v1/subs endpoint for subs

jxu577:


kyleconroy:
While where changing these files, we should make sure to add a test for `customer.subscriptions.all()`

kyleconroy:
I'm re-running the build (looks like we hit an API error)

cupcait:
cc @stripe/api-libraries

stripe/test/resources/test_customers.py
```diff
@@ -234,20 +234,38 @@ def test_legacy_delete_subscription(self):
             None
         )
 
+    def test_list_customer_subscriptions(self):
+        customer = stripe.Customer.construct_from({
+            'id': 'cus_foo',
+            'subscriptions': {
+                'object': 'list',
+                'url': '/v1/subscriptions',
```
kyleconroy: This isn't an accurate test case. The customer subscriptions endpoint will continue to return the same URL: `/v1/customers/cus_foo/subscriptions'

kyleconroy:
Once the tests are fixed to update the URLs in `test_customers.py`, this will be good to go.

stripe/test/resources/test_customers.py
jxu577: Actually, the old `customer.subscriptions` methods also call the new URL - this is because the URL is set on the subscription object, which is the same class whether through the customer or by itself. This is why we could not roll this out until all the data had been verified. Otherwise it would have been fine to roll out if only the new methods called the new endpoints.

kyleconroy: I don't believe this is accurate. For operations on a subscription object, you are correct: it will use the new endpoint. However, for the operations on `ListObject` such as `create`, `all`, `retrieve`, it will [use the URL returned in the `url` parameter](https://github.com/stripe/stripe-python/blob/jacqueline-subs-v1/stripe/resource.py#L314). We aren't changing the subscriptions URL that's returned from the `/v1/customers/cus_foo` endpoint. 

{
  "id": "cus_7j89JgbvIey5zw",
  "object": "customer",
  "subscriptions": {
    "object": "list",
    "data": [
      {
        "id": "sub_7j8BvyTb5vfUef",
        "object": "subscription",
        ...
      }
    ],
    "has_more": false,
    "total_count": 1,
    "url": "/v1/customers/cus_7j89JgbvIey5zw/subscriptions"
  }
}
```

kyleconroy:
:+1:

PR #225, opened 2016-04-12
Fix deprecation warnings

olivierbellone:
This PR fixes deprecation warnings:
- updated `resource.py` to use `.list()` instead of `.all()` in some convenience methods
- updated a lot of tests to use `.list()` instead of `.all()` 
- updated a couple tests for legacy methods to catch the warnings and assert their presence

The test suite's output should now be much cleaner! :)


olivierbellone:
r? @brandur 
cc @stripe/api-libraries

brandur:
Amazing! Thanks :)

PR #230, opened 2016-05-16
Add support for Alipay accounts

### LINKED ISSUE #229: AliPay source objects have no `delete()`-method
When an AliPay account is attached to a Customer record, they cannot be deleted using the Python-bindings, as the standard `delete()`-method does not exist.  You are able to delete this object using a basic `curl`, so I'm assuming it's an oversight.

olivierbellone:
Fixes #229.

r? @stripe/api-libraries


brandur:
Awesome :) +1.

PR #235, opened 2016-06-14
Add `modify` class method for updateable resources

kyleconroy:
I'd much rather call this method `update`, but we can't use that name because `StripeObject` already has an `update` method, and I'd rather not make this a breaking change. Note that this PR is not ready merge. There are a few resources that are going to need a custom `modify` method due to URL format.

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -859,13 +859,22 @@ class ApplicationFee(ListableAPIResource):
 
 class ApplicationFeeRefund(UpdateableAPIResource):
 
-    def instance_url(self):
-        token = util.utf8(self.id)
-        fee = util.utf8(self.fee)
+    @classmethod
+    def _build_instance_url(cls, fee, sid):
+        fee = util.utf8(fee)
+        sid = util.utf8(sid)
         base = ApplicationFee.class_url()
         cust_extn = urllib.quote_plus(fee)
-        extn = urllib.quote_plus(token)
+        extn = urllib.quote_plus(sid)
         return "%s/%s/refunds/%s" % (base, cust_extn, extn)
+
+    @classmethod
+    def modify(cls, fee, sid, **params):
+        url = cls._build_instance_url(fee, sid)
+        return cls._modify(url, **params)
+
+    def instance_url(self):
+        return self._build_instance_url(self.fee, self.id)
 
     @classmethod
     def retrieve(cls, id, api_key=None, **params):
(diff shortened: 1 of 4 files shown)


tetrakai:
Is there a reason we're creating a `modify` method here, rather than using the current paradigm of updating a dictionary and re-saving?

brandur:
@kyleconroy Would you mind clarifying the approach to Python here compared to your related pulls in PHP and Ruby? It seems like the other languages, only subscriptions became updatable, but if I'm reading this correctly, it looks like this one takes a more broad approach? Thanks!

kyleconroy:
@brandur My plan is to make all objects updatable, but I wanted to get sign off on the PHP and Ruby changes before going through and updating all the call sites. I actually may do what I did with the Ruby bindings and separate save and update.

kyleconroy:
@brandur thoughts on the latest version?

brandur:
@kyleconroy Nice! Just two notes/questions:

* As with stripe-ruby, I think there's some possibility that you may just want to merge `ModifiableAPIResource` with `UpdateableAPIResource`. As far as I understand it, these are both going to go to the same API action in all cases, and support for one should always imply support for the other.
* Also as with stripe-ruby, it might be worthwhile seeing if can get in one or two "core" tests for modifiable so that it's being tested somewhere more general than the models themselves (i.e. within charge). I imagine that this isn't convention for the Python bindings though, so it's not that big of a deal.

Looking good though! Thanks.

kyleconroy:
@brandur so I've been going through and I don't think we want to merge the two concepts. There are actually many resources that can be saved but can't be updated. This is due to the fact that the instance URL is determined using the state of the object (such as source or bank account).

brandur:
> @brandur so I've been going through and I don't think we want to merge the two concepts. There are actually many resources that can be saved but can't be updated. This is due to the fact that the instance URL is determined using the state of the object (such as source or bank account).

Hrm, fair point on the sources/bank accounts.

One thing that I might suggest there though is that those resources may be the exceptions rather than the norm. It's only the polymorphic resources for which the update logic is a little bit hairy. As such, it might be worth merging the concepts and just blacklisting the use from certain places like sources.

And even for sources/bank accounts, I believe there's a new feature that we should see pretty soon that will allow them to be updated the same as every other resources. (Hit me up over chat if you want me to elaborate on that one.)

kyleconroy:
@brandur alright, this is ready to go as well

kyleconroy:
I'm not sure why the `push` branch isn't building on Travis. The other branch seemed to work just fine.

brandur:
> I'm not sure why the push branch isn't building on Travis. The other branch seemed to work just fine.

Weird! It's totally stuck, haha. Probably not that big of a deal though. Looks like the build is good.

LGTM!

PR #241, opened 2016-08-23
UPDATED Oct 4 - Adds a configurable default timeout to the constructor for the requests http client

alexhughson:


brandur:
This looks quite reasonable. Thanks!

Do you use this by setting `stripe.default_http_client = ...`?

alexhughson:
Yeah, I was following the pattern of the deadline setting for the UrlFetch client.

alexhughson:
I don't understand what is happening with the CI though :(

brandur:
Hm, this line is actually coming from the requests library itself:

```
http_error_msg = u'%s Client Error: %s for url: %s' % (self.status_code, reason, self.url)
```

It was changed pretty recently to include the unicode prefix, so that could be the problem. Requests itself doesn't support 3.2 anymore. From their README:

> Requests officially supports Python 2.6–2.7 & 3.3–3.5, and runs great on PyPy.

It kind of looks like 3.2 was EOLed more than two years ago, so I wonder if we should just drop it.

brandur:
@alexhughson I just merged #242 to drop support for Python 3.2. Can you rebase against master and re-push to get the CI light green? Thanks!

alexhughson:
@brandur Sorry, got distracted by some other stuff, but I did the rebase and CR is running.

alexhughson:
Looks to have passed

brandur:
@alexhughson No worries, and thanks for following up! We have one other major pull coming in that will conflict with this one, but I'll merge it in as soon as it's in mainline.

stripe/http_client.py
```diff
@@ -84,9 +84,11 @@ def request(self, method, url, headers, post_data=None):
 class RequestsClient(HTTPClient):
     name = 'requests'
 
-    def request(self, method, url, headers, post_data=None):
-        kwargs = {}
+    def __init__(self, timeout=80, **kwargs):
+        super(RequestsClient, self).__init__(**kwargs)
+        self._timeout = timeout
 
+    def request(self, method, url, headers, post_data=None, **kwargs):
```
brandur: @alexhughson Actually, just taking one more look at this: would you mind reverting this line to how it was? I'm not sure that this change is necessary.

Could you also try rebasing the changes on master? You'll need to merge a conflict in `RequestsClientTests`, but it's not too bad.

brandur:
LGTM.

Thanks!

brandur:
Released as 1.40.0.

PR #242, opened 2016-08-24
Drop support for Python 3.2

brandur:
So 3.2 is failing our build matrix because our dependencies (like requests) are
using "u" unicode prefixes which were originally dropped from Python 3, but
were then re-added in 3.3 for compatibility reasons.

I made a token attempt to get this fixed, but couldn't quite get there. Given
that Python 3.2's been unsupported for 2+ years now, let's take the easy way
out and drop it.

[1] http://stackoverflow.com/questions/8679985/unicode-strings-syntax-in-python


brandur:
I checked in with @kyleconroy about this change, so I'm going to self-approve this one:

LGTM.

Issue #53, opened 2013-12-12
Thread safety


szs8:
Is this api thread safe? I looked at the flask integration example which exposed the module at the global level. Just want to make sure that it is thread safe.

PR #245, opened 2016-09-09
UPDATE: Stripe API proxy support for requests, pycurl, and urllib2 clients

krassif:
In complex enterprise environment, different proxies may be used for communication across the many enterprise network zones, the Internet. Obviously, the HTTP_PROXY and HTTPS_PROXY environment variables are more trouble than help there.

A clean solution would be introducing API level proxy support that limits the proxy scope to just stripe communication. Everybody else would remain unaffected and could continue using either direct or own, different HTTP proxy configuration.

This pull request introduces the proxy API support described. The new proxy-example.py file demonstrates the proxy API usage.

Please consider prompt review of this pull request. Not having the proxy support will affect our project with an externally set, beyond our control hard deadline.

Happy to address any questions or concerns.

Thank you!


stripe/api_requestor.py
```diff
@@ -69,9 +69,10 @@ def __init__(self, key=None, client=None, api_base=None, account=None):
         self.stripe_account = account
 
         from stripe import verify_ssl_certs as verify
+        from stripe import proxies
 
         self._client = client or stripe.default_http_client or \
-            http_client.new_default_http_client(verify_ssl_certs=verify)
+            http_client.new_default_http_client(verify_ssl_certs=verify, proxies= proxies)
```
olivierbellone: Should be `proxies=proxies` (no space after `=`).

stripe/http_client.py
```diff
@@ -51,13 +51,20 @@
 except ImportError:
     urlfetch = None
 
+# proxy support for the pycurl client
+if sys.version_info >= (3, 0):
+    from urllib.parse import urlparse
+else:
+    from urlparse import urlparse
```
olivierbellone: I think you can just write `from urlparse import urlparse` and `2to3` will automatically adapt the import for Python 3.

stripe/http_client.py
```diff
@@ -51,13 +51,20 @@
...
 
 def new_default_http_client(*args, **kwargs):
     if urlfetch:
+        # no proxy support in urlfetch. for a patch, see:
+        # https://code.google.com/p/googleappengine/issues/detail?id=544
         impl = UrlFetchClient
     elif requests:
         impl = RequestsClient
-    elif pycurl:
+    elif pycurl:
```
olivierbellone: No space after `:`.

stripe/http_client.py
```diff
@@ -73,8 +80,9 @@ def new_default_http_client(*args, **kwargs):
 
 class HTTPClient(object):
 
-    def __init__(self, verify_ssl_certs=True):
+    def __init__(self, verify_ssl_certs=True, proxies= None):
```
olivierbellone: Should be `proxies=None` (no space after `=`).

stripe/http_client.py
```diff
@@ -73,8 +80,9 @@ def new_default_http_client(*args, **kwargs):
...
         self._verify_ssl_certs = verify_ssl_certs
+        self._proxies = proxies.copy() if proxies else proxies
```
olivierbellone: I'd rewrite that as `self._proxies = proxies.copy() if proxies else None`.

stripe/http_client.py
```diff
@@ -144,7 +155,16 @@ def _handle_request_error(self, e):
 class UrlFetchClient(HTTPClient):
     name = 'urlfetch'
 
-    def __init__(self, verify_ssl_certs=True, deadline=55):
+    def __init__(self, verify_ssl_certs=True, proxies= None, deadline=55):
```
olivierbellone: Should be `proxies=None` (no space after `=`).

stripe/http_client.py
```diff
@@ -144,7 +155,16 @@ def _handle_request_error(self, e):
...
+        
+        # no proxy support in urlfetch. for a patch, see:
+        # https://code.google.com/p/googleappengine/issues/detail?id=544        
+        if proxies:
+            raise error.APIConnectionError(
+                "No proxy support in urlfetch library. "
+                "Set stripe.default_http_client to either RequestsClient, "
+                "PycurlClient, or Urllib2Client instance to use a proxy." )
```
olivierbellone: Remove the space before `)`.

stripe/http_client.py
```diff
@@ -187,10 +207,31 @@ def _handle_request_error(self, e, url):
         msg = textwrap.fill(msg) + "\n\n(Network error: " + str(e) + ")"
         raise error.APIConnectionError(msg)
 
-
```
olivierbellone: Reinsert the removed line (need 2 empty lines between each class definition).

stripe/http_client.py
```diff
@@ -187,10 +207,31 @@ def _handle_request_error(self, e, url):
...
 class PycurlClient(HTTPClient):
     name = 'pycurl'
 
+    def __init__(self, verify_ssl_certs=True, proxies= None):
```
olivierbellone: Should be `def __init__(self, verify_ssl_certs=True, proxies=None):`

stripe/http_client.py
```diff
@@ -187,10 +207,31 @@ def _handle_request_error(self, e, url):
...
+        super( PycurlClient, self ).__init__( verify_ssl_certs= verify_ssl_certs, proxies= proxies )
```
olivierbellone: Should be `super(PycurlClient, self).__init__(verify_ssl_certs=verify_ssl_certs, proxies=proxies)`

stripe/http_client.py
```diff
@@ -187,10 +207,31 @@ def _handle_request_error(self, e, url):
...
+        # need to urlparse the proxies, since PyCurl
+        # consumes the proxy url in small pieces
+        if self._proxies:
+            # now that we have the parser, get the proxy url pieces
+            proxies = self._proxies
+            for scheme in proxies:
+                proxies[ scheme ] = urlparse( proxies[ scheme ] )
```
olivierbellone: Should be `proxies[scheme] = urlparse(proxies[scheme])`.

stripe/http_client.py
```diff
@@ -187,10 +207,31 @@ def _handle_request_error(self, e, url):
...
+    
+    def __get__proxy__( self, url ):
```
olivierbellone: Should be `def __get__proxy__(self, url):`.

olivierbellone:
Hi @krassif!

Thanks for the PR. I can definitely see how better proxy support could be useful in some environments.

Your PR seems to have a few PEP8 violations. I noted some of them in comments, but you can find a complete list in the failing Python 2.7 test suite: https://travis-ci.org/stripe/stripe-python/jobs/158882520.

Could you please correct these violations and make sure the test suite passes?

krassif:
Thank you for the prompt review.
Applied the suggested formatting and references.

Let me know if there is any outstanding issues.

stripe/http_client.py
```diff
@@ -144,7 +153,16 @@ def _handle_request_error(self, e):
 class UrlFetchClient(HTTPClient):
     name = 'urlfetch'
 
-    def __init__(self, verify_ssl_certs=True, deadline=55):
+    def __init__(self, verify_ssl_certs=True, proxies=None, deadline=55):
+
+        # no proxy support in urlfetch. for a patch, see:
+        # https://code.google.com/p/googleappengine/issues/detail?id=544
+        if proxies:
+            raise error.APIConnectionError(
```
brandur: Great idea raising an error here! What do you think about making this a `ValueError` instead? I think it may be more appropriate given that this is an invalid option.

stripe/http_client.py
```diff
@@ -51,9 +51,14 @@
...
     urlfetch = None
 
+# proxy support for the pycurl client
+from urlparse import urlparse
+
 
 def new_default_http_client(*args, **kwargs):
     if urlfetch:
+        # no proxy support in urlfetch. for a patch, see:
+        # https://code.google.com/p/googleappengine/issues/detail?id=544
```
brandur: I think we may be able to remove this note given that it's duplicated from within the URL fetch client's constructor.

brandur:
Cool, looks very useful! Just a couple thoughts related to the higher level design:

1. Even when in testmode we should never really see any non-HTTPS made against Stripe. I wonder if we should remove the scheme-based abstraction and just have a single proxy option?
2. Although I don't think it's necessary to be testing the proxy implementations of each underlying HTTP library, I think we should at least be making an effort to make sure that the new code here gets exercised in the test suite, especially since there's some parsing logic involved. Thoughts?

stripe/http_client.py
krassif: Done

stripe/http_client.py
krassif: Done

krassif:
Thanks for the review.

Re 1: I see your point about Stripe being HTTPS only. The underlying libraries however, requests being the first one to cover, allow for both HTTP and HTTPS proxies. I can see use for both connection settings in either unit or deployment tests, where the Stripe state shall not be altered.

Re 2: Totally! Will review the tests identifying the best place to hook and get proxy settings code coverage. If you happen to have specific test in mind, let me know in the meantime.

proxy-example.py
```diff
@@ -0,0 +1,37 @@
...
+        amount=200,
+        currency='usd',
+        card={
+            'number': '4242424242424242',
+            'exp_month': 10,
+            'exp_year': 2019
+        },
+        description='customer@gmail.com'
+    )
+    print( 'Success: %s, %r' % ( c.name, resp, ) )
```
brandur: One thing that might be a good idea is to move this file and `example.py` into their own `examples` directory so we don't have a bunch of extras hanging around at root, but I can handle that after this is merged.

brandur:
> Re 1: I see your point about Stripe being HTTPS only. The underlying libraries however, requests being the first one to cover, allow for both HTTP and HTTPS proxies. I can see use for both connection settings in either unit or deployment tests, where the Stripe state shall not be altered.

Ah, yes I see your point.

I wonder though if we should be trying to optimize the API for the "95% path" with respect to the fact that most people will probably just need to set a single proxy value.

I don't want to go crazy on taking advantage of duck typing, but I could imagine something along the lines of:

``` python
stripe.proxy = "http://<user>:<pass>@<proxy>:<port>"
```

This would set that value for both HTTP and HTTPs. If you really need to pass options for both, you could set a dictionary like so:

``` python
stripe.proxy = {
    "http": "http://<user>:<pass>@<proxy>:<port>",
    "https": "http://<user>:<pass>@<proxy>:<port>"
}
```

> Re 2: Totally! Will review the tests identifying the best place to hook and get proxy settings code coverage. If you happen to have specific test in mind, let me know in the meantime.

Awesome! Thank you.

krassif:
I hope the last commit addresses all the previously raised points and the PR is good to be merged. Let me know. Thanks.

proxy-example.py
```diff
@@ -0,0 +1,37 @@
+import stripe
+
+
+stripe.api_key = 'tGN0bIwXnHdwOa85VABjPdSn8nWY7G7I'
+
+print( "Attempting charge..." )
+
+stripe.proxies = {
```
brandur: @krassif Thanks for making those changes! It looks like one last `proxies` straggler here.

krassif:
Yeap! Thanks for catching this! As you can imagine the file ran slightly different looking on my side. Will update.

stripe/http_client.py
```diff
@@ -199,10 +246,22 @@ def parse_headers(self, data):
         return dict((k.lower(), v) for k, v in dict(headers).iteritems())
 
     def request(self, method, url, headers, post_data=None):
+
```
brandur: Mind just tightening up the whitespace here?

stripe/http_client.py
```diff
@@ -191,6 +215,29 @@ def _handle_request_error(self, e, url):
...
+            verify_ssl_certs=verify_ssl_certs, proxy=proxy)
+        # need to urlparse the proxy, since PyCurl
+        # consumes the proxy url in small pieces
+        if self._proxy:
+            # now that we have the parser, get the proxy url pieces
+            proxy = self._proxy
+            for scheme in proxy:
+                proxy[scheme] = urlparse(proxy[scheme])
+
+    def __get__proxy__(self, url):
```
brandur: Can we move this to be prefixed by only a single underscore and then put it down with the other private methods below the public interface? (i.e. by `_handle_request_error`)? Thanks!

brandur:
@krassif Nice, thank you! One thing: thanks for putting in those tests, but it sort of seems like we might not actually be verifying anything. Do you think we should add some assertions to make sure that the proxy configuration made it in in the right form?

stripe/http_client.py
krassif: Done within the latest commit.

stripe/http_client.py
krassif: Not at all. In.

brandur:
@krassif Awesome! Thank you. Do you think we could do something in the tests to verify that the proxy values are making it into the client objects as we expect?

krassif:
There is check for the requests library:
https://github.com/stripe/stripe-python/pull/245/files#diff-e1dca70032d14face72ead647a5bbe6dR132

Hope this suffice to merge.

krassif:
@brandur Can we have this pull request merged, if there is no outstanding items? Thanks!

brandur:
Hey @krassif, sorry about the delay here.

> There is check for the requests library:

Ah, yes! I saw that you added that. My concern was mainly for the Urllib2 and Pycurl tests, which add a proxy field, but then don't check anything.

I looked into writing a test a bit, but realize that it's quite tricky because adding an opener is purely manipulation on global state, and hard to verify after it's done. As an alternative here, what do you think about removing the proxy support here for `urllib2`? After looking at it in more detail, I don't think it's very good practice because by supplying Stripe with a `proxy` value, you could interfere with state that you've set elsewhere in the application.

Also, thoughts on testing Pycurl? Did that one turn out to be particularly tricky to verify

krassif:
Re Urllib2: I'd follow the certificate verification approach - issuing a warning in case proxy is specified and the library happens to be Urllib2. Personally, there is no need to take the proxy support out and rely on the caller preparing the global scope opener prior to calling into Stripe. If in agreement, will add and commit the warning for review.

Re Pycurl: Can you take the first shot at replacing the current ["pass"](https://github.com/stripe/stripe-python/blob/master/stripe/test/test_http_client.py#L228) with something meaningful? Will follow the lead adding the proxy case.

brandur:
> Re Urllib2: I'd follow the certificate verification approach - issuing a warning in case proxy is specified and the library happens to be Urllib2. Personally, there is no need to take the proxy support out and rely on the caller preparing the global scope opener prior to calling into Stripe. If in agreement, will add and commit the warning for review.

Not sure if I agree on this one! It seems to me that manipulating global scope is _always_ bad, but especially bad when it occurs on initializing a new object because you'd never expect side effects from such an operation. Also, it seems to me that if users do want proxy support for urllib2, it's relatively easily for them to just initialize the opener separately so that it's very obvious what's going on. We could even include that in the examples file to make it easy to figure out. Thoughts?

> Re Pycurl: Can you take the first shot at replacing the current "pass" with something meaningful? Will follow the lead adding the proxy case.

Oops, yeah, the current spider's web of mocking is not exactly intuitive is it? I added some basic assertions in #250 that you should be able to easily build off of to get a working test if you rebase on master.

krassif:
Re Urllib2: Took a second look at the urllib2 api. Seems we do not need to modify the global library state and may be executing the request on a local opener object (second sentence): https://docs.python.org/2/library/urllib2.html#urllib2.install_opener. Will give it a shot. If still flaky, will disable the proxy support for urllib2 with an exception like in the UrlFetch case.

Adding Pycurl proxy test on top of #250..

daspecster:
Hello! Great work on this lib BTW.

Will this allow support of urllib3 or requests for thread safety?

That said, I'm not sure if there are other areas in `stripe-python` that would break thread safety.

http://stackoverflow.com/questions/5825151/are-urllib2-and-httplib-thread-safe

LMK if I should put this in https://github.com/stripe/stripe-python/issues/53.

Thanks again!

krassif:
@daspecster Thank you for the note!
PRs goal is provide proxy API for the already supported http libraries. Re urllib3, like you mentioned, the requests library implemented on top of urllib3 is already in stripe-python. That being said, I'd be happy assisting additional library support.

daspecster:
@krassif, if this works with both urllib2 and urllib3 then kudos!

I have a project that I haven't started yet that may try to thread this so that's the only reason I bring it up.

It's an uncommon case though, as threading would imply generating many requests which would be bad form I think. My threading case is more of an architectural choice.

krassif:
@brandur Just added non-global opener for Urllib2 together with test cases covering Urllib2 non-proxy, string and dictionary specified proxies. In a similar way, PyCurl tests cover non-proxy, and the two supported proxy specs.

The branch has conflict because of stripe/__init__.py. Ready to rebase to the latest master, if this helps the merge. Thanks!

brandur:
> @brandur Just added non-global opener for Urllib2 together with test cases covering Urllib2 non-proxy, string and dictionary specified proxies. In a similar way, PyCurl tests cover non-proxy, and the two supported proxy specs.

@krassif Thanks a lot! This is looking very good to me now. I'm really liking the non-global opener approach for Urllib2.

> The branch has conflict because of stripe/init.py. Ready to rebase to the latest master, if this helps the merge. Thanks!

Yes, that would be great! I'll bring this in post-haste afterwards. Thanks!

krassif:
@brandur Rebased!

brandur:
Thanks again for all the hard work here!

LGTM.

krassif:
Thank you for the support on the pull request.
It' been pleasure!

brandur:
> Thank you for the support on the pull request.

No worries at all! And it's also been great working with you.

Released as 1.39.0.

PR #248, opened 2016-09-16
Add SubscriptionItem to support multiplan subscriptions

jxu577:
Add the SubscriptionItem resource that supports listing, creation, update, retrieval, and deletion through the `v1/subscription_items` endpoint.


kyleconroy:
👍

PR #249, opened 2016-09-27
Encode the items array as a dict

kyleconroy:


stripe/resource.py
```diff
@@ -55,6 +55,16 @@ def convert_to_stripe_object(resp, api_key, account):
         return resp
 
 
+def convert_array_to_dict(arr):
+    if isinstance(arr, list):
+        d = {}
+        for i, value in enumerate(arr):
```
tallowen: More idiomatic python would be:
```python
{"{0}".format(i): value for i, value in enumberate(arr)}
```

stripe/resource.py
```diff
@@ -55,6 +55,16 @@ def convert_to_stripe_object(resp, api_key, account):
...
+            d["{0}".format(i)] = value
```
tallowen: instead of:
```python
"{0}".format(i)
```
Can we do:
```python
str(i)
```

kyleconroy: Yeah, I avoided that because I'm not sure how it works when we use `2to3`. I'm sure it works

stripe/resource.py
kyleconroy: Yeah, I wanted to do that as well (see the first commit). That syntax doesn't work in Python 2.6

tallowen:
lgtm

PR #253, opened 2016-10-10
Return values when creating/modifying subscriptions

### LINKED ISSUE #252: stripe.Subscription.create returning None
Calling `stripe.Subscription.create` is returning `None`, even though a subscription is being created in the system (I can see it in the dashboard).

brandur:
A bug was introduced with subscription items [1] that overrode methods
on the subscription and model and resulted in create/modify no longer
return objects.

This patch addresses that problem by adding return statements.

Fixes #252.

r? @kyleconroy Can you check this one out? Thanks!

[1] https://github.com/stripe/stripe-python/pull/249/commits/48856aea2e35d82d5fa76d0ab4483513039883a0

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -802,13 +802,13 @@ class Subscription(CreateableAPIResource, DeletableAPIResource,
     def modify(cls, sid, **params):
         if "items" in params:
             params["items"] = convert_array_to_dict(params["items"])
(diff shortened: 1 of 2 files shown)


kyleconroy:
Yikes, that was my mistake. Thanks for adding the tests.

👍

brandur:
Released as 1.40.1.

PR #254, opened 2016-10-12
Adds support for sources

olivierbellone:
r? @brandur
cc @stripe/api-libraries

It seems like we forgot to update the Python bindings when we switched to using [sources](https://stripe.com/docs/api#sources) for Bitcoin payments a few weeks ago. This PR fixes that.


brandur:
Awesome!

LGTM.

brandur:
Released as 1.41.0.

PR #256, opened 2016-10-12
Fix lint like problems revealed by lgtm

brandur:
Details:

* Rescue BaseException instead of default Exception so that we don't
  accidentally handle an exit or keyboard interrupt.
* Make sure to call parent constructor of UrlFetchClient.
* Don't mix explicit and implicit return values.
* Remove unnecessary pass statement from OrderReturn.
* Comment empty exception rescue.
* Remove unused logger.

r? @dpetrovics Mind taking a look here? Thanks!


dpetrovics:
Thanks for doing this cleanup, @brandur !

LGTM

PR #265, opened 2016-11-22
flip conditional to prevent NameError on Python 3

ofek:
In Python 3 `unicode` is no longer a built-in function (as it is now `str`). I'm honestly unsure how the tests currently pass, as this most certainly will throw an exception.

Let me know if you would like me to include a test :)

### DIFF
--- a/stripe/util.py
+++ b/stripe/util.py
@@ -46,7 +46,7 @@ if not (json and hasattr(json, 'loads')):
 
 
 def utf8(value):
-    if isinstance(value, unicode) and sys.version_info < (3, 0):
+    if sys.version_info < (3, 0) and isinstance(value, unicode):
         return value.encode('utf-8')
     else:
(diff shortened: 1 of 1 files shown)


ob-stripe:
The code is processed through [2to3](https://docs.python.org/2/library/2to3.html) when running with Python 3, so the function is actually updated like this:

```diff
 def utf8(value):
-    if isinstance(value, unicode) and sys.version_info < (3, 0):
+    if isinstance(value, str) and sys.version_info < (3, 0):
         return value.encode('utf-8')
     else:
         return value
```

That said, your patch makes sense. @brandur-stripe, wdyt?

brandur:
Works for me. Thanks!

And great find on the 2to3 thing Olivier! I read this yesterday from my phone and couldn't think of an obvious explanation off the top of my head.

PR #266, opened 2016-11-23
http_client.py support for Session. Using Session can improve network performance.

NahomAgidew:
http_client.py support for Session. Using Session can improve network performance.

### DIFF
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -6,7 +6,6 @@ import email
 
 from stripe import error, util
 
-
 # - Requests is the preferred HTTP library
 # - Google App Engine has urlfetch
 # - Use Pycurl if it's there (at least it verifies SSL certs)
@@ -76,7 +75,6 @@ def new_default_http_client(*args, **kwargs):
 
 
 class HTTPClient(object):
-
     def __init__(self, verify_ssl_certs=True, proxy=None):
         self._verify_ssl_certs = verify_ssl_certs
         if proxy:
@@ -100,6 +98,7 @@ class RequestsClient(HTTPClient):
     def __init__(self, timeout=80, **kwargs):
         super(RequestsClient, self).__init__(**kwargs)
         self._timeout = timeout
+        self._session = requests.Session()
 
     def request(self, method, url, headers, post_data=None):
         kwargs = {}
@@ -114,12 +113,12 @@ class RequestsClient(HTTPClient):
 
         try:
             try:
-                result = requests.request(method,
-                                          url,
-                                          headers=headers,
-                                          data=post_data,
-                                          timeout=self._timeout,
-                                          **kwargs)
+                result = self._session.request(method,
+                                               url,
+                                               headers=headers,
+                                               data=post_data,
+                                               timeout=self._timeout,
+                                               **kwargs)
             except TypeError as e:
                 raise TypeError(
                     'Warning: It looks like your installed version of the '
@@ -272,7 +271,7 @@ class PycurlClient(HTTPClient):
         curl.setopt(pycurl.CONNECTTIMEOUT, 30)
         curl.setopt(pycurl.TIMEOUT, 80)
         curl.setopt(pycurl.HTTPHEADER, ['%s: %s' % (k, v)
-                    for k, v in headers.iteritems()])
+                                        for k, v in headers.iteritems()])
         if self._verify_ssl_certs:
             curl.setopt(pycurl.CAINFO, os.path.join(
                 os.path.dirname(__file__), 'data/ca-certificates.crt'))


stripe/http_client.py
```diff
@@ -114,12 +114,20 @@ def request(self, method, url, headers, post_data=None):
 
         try:
             try:
-                result = requests.request(method,
-                                          url,
-                                          headers=headers,
-                                          data=post_data,
-                                          timeout=self._timeout,
-                                          **kwargs)
+                if(session is not None):
```
deontologician-stripe: Should remove parens here since they aren't necessary

stripe/http_client.py
```diff
@@ -114,12 +114,20 @@ def request(self, method, url, headers, post_data=None):
...
+                    result = session.request(method,
```
deontologician-stripe: Rather than have two branches that are largely identical, we should probably refactor a bit. Something like:

```py
requestor = session if session is not None else requests
requestor.request(method, ...
```

NahomAgidew: @deontologician-stripe Refactored! I'll be pushing later tonight.

deontologician-stripe:
Looks like this is causing test failures. I think the primary cause is the tests expecting a certain number of parameters being passed to `.request`

NahomAgidew:
@deontologician-stripe Yeah

mastermatt:
If you don't mind two cents from a random guy. This PR seems to add a lot of Requests specific changes to package that tries to be abstract about which client it uses. If you're to the point where you need/want to start managing connection sessions a custom `RequestsClient` or `HTTPClient` set to `stripe.default_http_client` is probably a better way to go anyway. Mainly because there is no good way to take advantage of the perks from using a session from inside the package and instead the calling code must assume this responsibility.

brandur:
@mastermatt Not at all! Thanks!

Wow, so the connection pooling feature of the Requests library is quite confusing. It's not well-documented, but this issue [1] helps us understand that we do need an exterior session to get connection pooling work, but as this issue [2] indicates, you can actually pool connections across sessions as well because it's all just backed by the underlying library.

It seems to me that the right thing to do here might be to just create a session object internal to `HTTPClient` and just use that for every single request. It doesn't even have to leak out to `APIRequestor`. Thoughts @NahomAgidew?

[1] https://github.com/kennethreitz/requests/issues/3201
[2] https://github.com/kennethreitz/requests/issues/3445

NahomAgidew:
@brandur so the method signature inside HTTPClient will be: 
def request(self, session)
instead of
def request(self, method, url, headers, post_data=None)
 right???

brandur:
> @brandur so the method signature inside HTTPClient will be:

Nah, I was thinking that it might be better not to expose a session on the public API at all. What do you think about just creating a new session inside the constructor. i.e.

``` python
  class HTTPClient(object): 
      def __init__(self, verify_ssl_certs=True, proxy=None):
            ...
            self._session = Session()
            ...

      def def request(self, method, url, headers, post_data=None):
            ...
            result = self._session.request(method,
            ...
```

I believe that gains us the performance advantages that you were after and doesn't leak anything new out into the API. What do you think?

NahomAgidew:
@brandur Ahh ok. Seems legit, great idea! Do you want me to do the honors or you will?

mastermatt:
`requests.request` already uses a [session under the hood](https://github.com/kennethreitz/requests/blob/master/requests/api.py#L55-L56) for a single call. The perk of having your client aware of the session is if you wanted to manually manage the session for multiple calls. As is, the session would not get closed after being used. Which might be nice if you're making several API calls in the same process, but it also means those connections would stay open even if a call is made early in a process and never again. 
Also you'd want the session in the `RequestsClient.__init__`  rather than the base `HTTPClient` because this will error if the user was doesn't have `requests` installed.

brandur:
> requests.request already uses a session under the hood for a single call.

Yeah, this turns out now to be super useful though if you're after things like connecting pooling.

> The perk of having your client aware of the session is if you wanted to manually manage the session for multiple calls. As is, the session would not get closed after being used. Which might be nice if you're making several API calls in the same process,

Right, so i think in this case if you did this:

``` python
    stripe.default_http_client = stripe.http_client.RequestsClient( 
        verify_ssl_certs= stripe.verify_ssl_certs,
        proxy= stripe.proxy )

    resp = stripe.Charge.create(
        amount=200,
        currency='usd',
        card={
            'number': '4242424242424242',
            'exp_month': 10,
            'exp_year': 2019
        },
        description='customer@gmail.com'
    )
```

You should get yourself a persistent session that doesn't have to reinitialize a connect for every API request (i.e. because `RequestsClient` internalizes a session).

> but it also means those connections would stay open even if a call is made early in a process and never again.

Generally this isn't really all that negative of a side effect. I believe that the connections will be closed after their keep alive expires and there hasn't been any recent activity.

> Also you'd want the session in the RequestsClient.__init__ rather than the base HTTPClient because this will error if the user was doesn't have requests installed.

Oops, yep. I only looked at the diff when I commented above and recommended the wrong class. Definitely goes in `RequestsClient`.

@NahomAgidew Would you mind making that change and also see what you can do to minimize your patch (i.e. remove whitespace changes and whatnot)?

NahomAgidew:
@brandur No problem.

brandur:
@NahomAgidew Thanks! Hm, it looks like you're still getting a few test failures:

```
======================================================================

ERROR: test_new_default_http_client_pycurl (test_http_client.HttpClientTests)

----------------------------------------------------------------------

  File "/home/travis/build/stripe/stripe-python/stripe/test/test_http_client.py", line 44, in test_new_default_http_client_pycurl

    stripe.http_client.PycurlClient)

  File "/home/travis/build/stripe/stripe-python/stripe/test/test_http_client.py", line 30, in check_default

    inst = stripe.http_client.new_default_http_client()

  File "/home/travis/build/stripe/stripe-python/stripe/http_client.py", line 74, in new_default_http_client

    return impl(*args, **kwargs)

  File "/home/travis/build/stripe/stripe-python/stripe/http_client.py", line 225, in __init__

    verify_ssl_certs=verify_ssl_certs, proxy=proxy)

  File "/home/travis/build/stripe/stripe-python/stripe/http_client.py", line 80, in __init__

    self._session = requests.Session()

AttributeError: 'NoneType' object has no attribute 'Session'
```

NahomAgidew:
@brandur That was the build before I made the appropriate changes. Somehow the build was pending and it built a few hours later I pushed it. 
I haven't yet submitted the new modification. I'll do it tonight.

NahomAgidew:
@brandur Update: Travis-CI build seems to lag the past couple of days cause whenever I make changes and push them it says: "Hang tight, the log cannot be shown until the build has started." So, I have to wait the next day to see my error.

brandur-stripe:
> @brandur Update: Travis-CI build seems to lag the past couple of days cause whenever I make changes and push them it says: "Hang tight, the log cannot be shown until the build has started." So, I have to wait the next day to see my error. Anyway, I have made the changes you requested above and have to see in the morning if they pass.

@NahomAgidew Yeah, unfortunately Travis seems to run into pretty frequent problems around build queues :/ This current one is ridiculous though so I think that something may have gone wrong. Can you try repushing (maybe with a `git amend` and force push if necessary) to see if you can kickstart it?

NahomAgidew:
@brandur I'll try it under a different pull request.

brandur:
Replacing this with #276 which includes test fixes.

PR #269, opened 2016-12-08
Add debug and info logging

deontologician-stripe:
This adds debug/info logging in the client, as well as some tests for whether the env var or global flag should be in force. The remaining part would be documenting how to use the flag, and possibly mentioning how to configure logging directly.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -15,6 +15,9 @@ verify_ssl_certs = True
 proxy = None
 default_http_client = None
 
+# Set to either 'debug' or 'info', controls console logging
+log = None
+
 # Resource
 from stripe.resource import (  # noqa
     Account,
--- a/stripe/api_requestor.py
+++ b/stripe/api_requestor.py
@@ -252,14 +252,19 @@ class APIRequestor(object):
             for key, value in supplied_headers.items():
                 headers[key] = value
 
+        util.log_info('Request to Stripe api', method=method, path=abs_url)
+        util.log_debug(
+            'Post details', post_data=post_data, api_version=api_version)
+
         rbody, rcode, rheaders = self._client.request(
             method, abs_url, headers, post_data)
 
-        util.logger.info('%s %s %d', method.upper(), abs_url, rcode)
-        util.logger.debug(
-            'API request to %s returned (response code, response body) of '
-            '(%d, %r)',
-            abs_url, rcode, rbody)
+        util.log_info(
+            'Stripe API response', path=abs_url, response_code=rcode)
+        util.log_debug('API response body', body=rbody)
+        if 'Request-Id' in rheaders:
+            util.log_debug('Dashboard link for request',
+                           link=util.dashboard_link(rheaders['Request-Id']))
         return rbody, rcode, rheaders, my_api_key
 
     def interpret_response(self, rbody, rcode, rheaders):
@@ -273,6 +278,13 @@ class APIRequestor(object):
                 "(HTTP response code was %d)" % (rbody, rcode),
                 rbody, rcode, rheaders)
         if not (200 <= rcode < 300):
+            util.log_info(
+                'Stripe API error received',
+                error_code=resp.get('error', {}).get('code'),
+                error_type=resp.get('error', {}).get('type'),
+                error_message=resp.get('error', {}).get('message'),
+                error_param=resp.get('error', {}).get('param'),
+            )
             self.handle_api_error(rbody, rcode, resp, rheaders)
         return resp
 
--- a/stripe/test/test_util.py
+++ b/stripe/test/test_util.py
@@ -1,0 +1,139 @@
+from __future__ import print_function
+
+from collections import namedtuple
+
+try:
+    from unittest import mock
+except ImportError:
+    import mock
+
+from stripe import util
+from stripe.test.helper import StripeUnitTestCase
+
+import __builtin__
+
+PRINT_FUNC_STRING = __builtin__.__name__ + '.print'
+
+LogTestCase = namedtuple('LogTestCase', 'env flag should_output')
+FmtTestCase = namedtuple('FmtTestCase', 'props expected')
+
+
+class UtilTests(StripeUnitTestCase):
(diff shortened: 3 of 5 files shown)


deontologician-stripe:
Fixing some test failures...

stripe/util.py
```diff
@@ -1,10 +1,24 @@
+from __future__ import print_function
```
ob-stripe: The project already uses `2to3`, so technically you could remove this import and just use the `print` statement -- it will automatically be converted to the equivalent function in Python 3.

(I would personally prefer that we use `__future__` and/or `six` and not rely on automatic code conversion though.)

stripe/util.py
```diff
@@ -58,3 +72,38 @@ def utf8(value):
...
+
+def log_info(message, *args, **kwargs):
+    if _console_log_level() in ['debug', 'info']:
+        print(message.format(*args, **kwargs))
+    logger.info(message.format(*args, **kwargs))
+
+def _test_or_live_environment():
+    if stripe.api_key is None:
+        return
+    match = re.match(r'sk_(live|test)_', stripe.api_key)
```
ob-stripe: Note that not all API keys follow the `sk_(live|test)_` convention. Some older keys simply start with `sk_` without the `(live|test)_` part, and some yet older keys don't have a prefix at all, they're just random alphanumeric characters.

Unfortunately for those keys the only way of knowing whether they're a test or live key would be to send an API request and check the `livemode` attribute in the response.

stripe/util.py
```diff
@@ -1,10 +1,24 @@
+from __future__ import print_function
+
 import logging
 import sys
 import os
+import stripe
```
ob-stripe: Nit, but per [PEP 8](https://www.python.org/dev/peps/pep-0008/#imports), local imports should come after standard library imports.

brandur:
Hey @deontologician-stripe, I'll take a more detailed look at the code a bit later, but this looks like a great start! Could you include a couple more things here to make review a little easier:

1. Add sample output of what a debug and info log looks like in the PR description.
2. Draft up a section in the README that shows recommendations for usage.

stripe/util.py
deontologician-stripe: Good call, this was a bit needlessly sloppy

stripe/util.py
deontologician-stripe: Ok, I think then I'll make sure it can deal with older keys without barfing, and just do what it does in the case of no api-key and default to test.

stripe/util.py
deontologician-stripe: So the trick here is that I want to mock out `__builtin__.print` in the tests, and to do that I need to use the print function even in python2

stripe/test/test_util.py
```diff
@@ -0,0 +1,132 @@
...
+
+from stripe import util
+from stripe.test.helper import StripeUnitTestCase
+
+import __builtin__
+
+PRINT_FUNC_STRING = __builtin__.__name__ + '.print'
+
+
+class DashboardLinkTests(StripeUnitTestCase):
```
brandur: This is super nitpicky, but I notice that most of the other test suites are kind of modeled after a single class like `UtilTests`. Do you think that we maybe should consider following that pattern here too? I see that you're using an inheritance model here for some of the suites, but I wonder if that could just as easily be refactored to a set of test helpers.

And thanks for getting some util tests started BTW! Crazy that we didn't have any before.

stripe/api_requestor.py
```diff
@@ -252,14 +252,21 @@ def request_raw(self, method, url, params=None, supplied_headers=None):
             for key, value in supplied_headers.items():
                 headers[key] = value
 
+        util.log_info('Stripe {method} request to {path}',
+                      method=method, path=abs_url)
```
deontologician-stripe: Ah! I remember you mentioning this before, but I couldn't find anything on it. This is totally doable, I'll add it.

stripe/test/test_util.py
deontologician-stripe: Yeah, looking at this, the way it is now makes no sense. Originally I had them separate, but after a few refactors it's obvious they should just be in the same class

stripe/__init__.py
```diff
@@ -14,6 +14,7 @@
 verify_ssl_certs = True
 proxy = None
 default_http_client = None
+log = None
```
deontologician-stripe: :+1:

deontologician-stripe:
Ok, I've incorporated everything except for the `logfmt` stuff. I'll finish that up monday

deontologician-stripe:
Ok, simple implementation of logfmt in there, and I re-jiggered some stuff to use it. Should be good to go

?ptal @brandur-stripe 
cc @stripe-internal/dev-experience

brandur:
>  cc @stripe-internal/dev-experience

BTW, I should have mentioned it in Slack, but I don't think you can mention teams across orgs. We have @stripe/api-libraries for this purpose on the public repos.

stripe/api_requestor.py
```diff
@@ -252,14 +252,19 @@ def request_raw(self, method, url, params=None, supplied_headers=None):
...
         rbody, rcode, rheaders = self._client.request(
             method, abs_url, headers, post_data)
 
-        util.logger.info('%s %s %d', method.upper(), abs_url, rcode)
-        util.logger.debug(
-            'API request to %s returned (response code, response body) of '
-            '(%d, %r)',
-            abs_url, rcode, rbody)
+        util.log_info(
+            'Stripe API response', path=abs_url, response_code=rcode)
```
deontologician-stripe: Yeah, that sounds reasonable, will do

deontologician-stripe:
Ok, I added a more specific error log. I had to put it in another section of the code to avoid parsing the json twice, but I don't think it will be significantly more noisy.

dpetrovics-stripe:
Looking good, @deontologician-stripe! Thanks for the great README section as well

brandur:
+1. This looks awesome. I'd say lets bring it in and cut a release. @deontologician-stripe Can you try your hand at that?

deontologician-stripe:
Sounds good :+1:

Issue #274, opened 2016-12-30
Log methods should not use print


mrkevinze:
When verbose logging is enabled, it was surprising that the details are printed to the console via `print` and are not part of the log message itself.

This is the current code in `util.py`:

```
def log_debug(message, **params):
    if _console_log_level() == 'debug':
        print(logfmt(dict(message=message, **params)))
    logger.debug(message, params)


def log_info(message, **params):
    if _console_log_level() in ['debug', 'info']:
        print(logfmt(dict(message=message, **params)))
    logger.info(message, params)
```

This causes problems when the application needs to filter through its logs, because the individual print statements get elevated to severity level error (at least on App Engine). Also, passing `params` to the log method does not change the output log statement, i.e. only the message will be logged and not its details.

My suggested fix is to replace the print statement with a log statement as follows, thereby making the entire method rely on the logging facility:

```
def log_debug(message, **params):
    if _console_log_level() == 'debug':
        logger.debug(message + "\n" + logfmt(dict(**params)))
    else:
        logger.debug(message)


def log_info(message, **params):
    if _console_log_level() in ('debug', 'info'):
        logger.info(message + "\n" + logfmt(dict(**params)))
    else:
        logger.info(message)
```

Python 2.7.13
Stripe 1.44.0

brandur-stripe:
Thanks for the report!

cc @deontologician-stripe Can you take a look at this please?

deontologician-stripe:
It seems like there are two behaviors under discussion here. The first is that `print` and `logger.debug` output different information. I agree this is definitely wrong and I'll fix it.

The second is the behavior that if you use `STRIPE_LOG` or `stripe.log`, you'll always get logs to the console (it uses `print` under the hood). This is intentional, the main reason for it is to always have an easy way to see what is being sent to and from the Stripe API without having to configure logging. If you're in a situation where anything printed to the console is considered an error, you should definitely set up logging explicitly (using `logging.basicConfig()` and a custom handler) and not use the `STRIPE_LOG` or `stripe.log` settings.

Although on second thought, it might make sense to ensure the console logging does go to `stderr` and not `stdout`. That is probably what App Engine is triggering on

deontologician-stripe:
OK, proposed fix up in #275

mrkevinze:
[...]

PR #275, opened 2017-01-03
Log to stderr, logfmt for both logger and print statements

deontologician-stripe:
Proposed fix for #274 

~Failures in travis that aren't showing up locally. Investigating~ Resolved

r? @brandur-stripe

### DIFF
--- a/stripe/util.py
+++ b/stripe/util.py
@@ -87,15 +87,17 @@ def _console_log_level():
 
 
 def log_debug(message, **params):
+    msg = logfmt(dict(message=message, **params))
     if _console_log_level() == 'debug':
-        print(logfmt(dict(message=message, **params)))
-    logger.debug(message, params)
+        print(msg, file=sys.stderr)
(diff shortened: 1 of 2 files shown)


deontologician-stripe:
Turns out changing this exposed a unicode bug the previous change had created but the tests didn't uncover

stripe/util.py
```diff
@@ -116,11 +118,12 @@ def dashboard_link(request_id):
 
 def logfmt(props):
     def fmt(key, val):
-        val = str(val)
+        if not isinstance(val, basestring):
+            val = unicode(val)
```
brandur: Mind annotating these lines with a comment describing why this is done? (I assume it's just to convert non-unicode strings to unicode, but it couldn't hurt because the naming of `basestring` is not intuitive.)

deontologician-stripe:
Yeah, I should have thought of it before, whoops

PR #276, opened 2017-01-06
Use an internal Requests session for speed

brandur:
Improves performance for Requests by using an internal session that allows for connection pooling.

Replaces #266.


brandur:
Released as 1.46.0.

PR #279, opened 2017-02-12
Start testing Python 3.6 and 3.7

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries

Start testing Python 3.6 and 3.7-dev (failures allowed for the latter).

Also use the `pypy3` tagged version rather than `pypy3.3-5.2-alpha1`.


brandur-stripe:
+1. Thanks!

PR #281, opened 2017-02-16
Adds deauthorize method to Account class

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries

This PR adds a `deauthorize()` method to the `Account` class, similar to what already exists in [stripe-ruby](https://github.com/stripe/stripe-ruby/blob/5dd5d2b608af8b597b5298b22eab171d570d6408/lib/stripe/account.rb#L94-L99).

Because the `https://connect.stripe.com/oauth/deauthorize` endpoint doesn't return standard error objects, I had to tweak the error handling code slightly. This isn't ideal, but it works, and it's aligned with what stripe-ruby does.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -10,6 +10,7 @@
 api_key = None
 api_base = 'https://api.stripe.com'
 upload_api_base = 'https://uploads.stripe.com'
+connect_api_base = 'https://connect.stripe.com'
 api_version = None
 verify_ssl_certs = True
 proxy = None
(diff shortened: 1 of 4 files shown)


ob-stripe:
Ok, nevermind, I'm going to close this one. My patch broke broke error handling, and the only solutions I can think of are all too hackish for my taste.

I'll try to revisit this later, probably by adding a separate `connect` module that will properly handle requests to `connect.stripe.com`.

brandur-stripe:
Cool, thanks. Yeah, it's a little unfortunate how we have so many separate hostnames :/

PR #286, opened 2017-02-21
Adds Customer.modify_source and Account.modify_external_account methods

### LINKED ISSUE #282: `modify` should be able to update Sources and External Accounts
Currently, the implementation requires you to write:

```
customer = stripe.Customers.retrieve(customer_id)  # 1st API call
card = customer.sources.retrieve(card_id)  # 2nd API Call
....
card.save()  # 3rd API Call
```

Which is a total of 3 API calls (large bottleneck) just to update a card. Other bindings (e.g. Node, Go), allow you to do this in a single API call:

```
stripe.Customers.update_card(card_id, customer_id=customer_id, **kwargs_to_update)
```

There's Pythonic ways of updating resources similarly, including Django's ORM implementation. I'm not sure what the exact syntax should look like, but it seems extremely inefficient to have to make 3 API calls to modify one resource (and to have to retrieve resources before modifying them)

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries

This PR adds two class methods:
- `Customer.modify_source`
- `Account.modify_external_account`

These methods lets users modify cards and bank accounts that are attached to customers/accounts without having to first retrieve the resources, e.g.:

```python
updated_card = stripe.Customer.modify_source(
    'cus_...',  # customer ID
    'card_...',  # card ID
    # update parameters
    exp_month=12,
    exp_year=2020,
    name='Stripey McStripe'
)
```

Fixes #282.

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -538,6 +538,13 @@ class Account(CreateableAPIResource, ListableAPIResource,
         )
         return self
 
+    @classmethod
+    def modify_external_account(cls, sid, external_account_id, **params):
+        url = "%s/%s/external_accounts/%s" % (
+            cls.class_url(), urllib.quote_plus(util.utf8(sid)),
+            urllib.quote_plus(util.utf8(external_account_id)))
+        return cls._modify(url, **params)
+
 
 class AlipayAccount(UpdateableAPIResource, DeletableAPIResource):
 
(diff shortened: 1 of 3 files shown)


brandur-stripe:
Awesome @ob-stripe!

I think it's a little unusual in our bindings here how haphazardly these class-level helpers are provided (you get them in some places, but probably not everywhere that it's useful) but given that these address a direct need and are relatively in-line with our conventions, it seems reasonable to bring this. Thanks!

brandur-stripe:
Released as 1.48.0.

PR #287, opened 2017-02-21
Add OAuth methods

ob-stripe:
Currently, our docs advise users to send requests to `connect.stripe.com` by their own means. This PR adds support for OAuth directly in the Python bindings.

The design is a little bit awkward and could probably be improved -- I'm open to suggestions :)

To recap:
- I added a new `oauth` module containing a single static `OAuth` class
- the `OAuth` class has 3 static methods:
    - `authorize_url` returns the URL that platforms need to redirect their users to start the OAuth flow (https://stripe.com/docs/connect/reference#get-authorize)
    - `token` is used to turn an authorization code into an access token (https://stripe.com/docs/connect/reference#post-token)
    - `deauthorize` is used to deauthorize a currently connected account (https://stripe.com/docs/connect/reference#post-deauthorize)
- the `client_id` can be set globally or per request (just like API keys)
- since the errors that are returned by `connect.stripe.com` are different from those used by `api.stripe.com`, I created a new `OAuthRequestor` class (derived from `APIRequestor`) that will raise an `OAuthError` if the API returns an error
- because there are now two different types of requestors, I made some test classes abstract, and specified the appropriate class / class name in derived classes
- finally, I added a basic example in `examples/oauth.py` that demonstrates how to use all 3 new methods to connect and disconnect accounts to a platform

@stripe/api-libraries wdyt?

### DIFF
--- a/examples/oauth.py
+++ b/examples/oauth.py
@@ -1,0 +1,54 @@
+import os
+
+import stripe
+from flask import Flask, request, redirect
+
+
+stripe.api_key = os.environ.get('STRIPE_SECRET_KEY')
+stripe.client_id = os.environ.get('STRIPE_CLIENT_ID')
+
+app = Flask(__name__)
+
+
+@app.route('/')
+def index():
+    return '<a href="/authorize">Connect with Stripe</a>'
+
+
+@app.route('/authorize')
+def authorize():
+    url = stripe.OAuth.authorize_url(scope='read_only')
+    return redirect(url)
+
+
+@app.route('/oauth/callback')
+def callback():
+    code = request.args.get('code')
+    try:
+        resp = stripe.OAuth.token(grant_type='authorization_code', code=code)
+    except stripe.error.OAuthError as e:
+        return 'Error: ' + str(e)
+
+    return '''
+<p>Success! Account <code>{stripe_user_id}</code> is connected.</p>
+<p>Click <a href="/deauthorize?stripe_user_id={stripe_user_id}">here</a> to
+disconnect the account.</p>
+'''.format(stripe_user_id=resp['stripe_user_id'])
+
+
+@app.route('/deauthorize')
+def deauthorize():
+    stripe_user_id = request.args.get('stripe_user_id')
+    try:
+        stripe.OAuth.deauthorize(stripe_user_id=stripe_user_id)
+    except stripe.error.OAuthError as e:
+        return 'Error: ' + str(e)
+
+    return '''
+<p>Success! Account <code>{stripe_user_id}</code> is disconnected.</p>
+<p>Click <a href="/">here</a> to restart the OAuth flow.</p>
+'''.format(stripe_user_id=stripe_user_id)
+
+
+if __name__ == '__main__':
+    app.run(port=int(os.environ.get('PORT', 5000)))
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -8,7 +8,9 @@
 # Configuration variables
 
 api_key = None
+client_id = None
 api_base = 'https://api.stripe.com'
+connect_api_base = 'https://connect.stripe.com'
 upload_api_base = 'https://uploads.stripe.com'
 api_version = None
 verify_ssl_certs = True
@@ -53,6 +55,9 @@ from stripe.resource import (  # noqa
     Token,
     Transfer)
 
+# OAuth
+from stripe.oauth import OAuth  # noqa
+
 # Error imports.  Note that we may want to move these out of the root
 # namespace in the future and you should prefer to access them via
 # the fully qualified `stripe.error` module.
@@ -65,6 +70,7 @@ from stripe.error import (  # noqa
     RateLimitError,
     CardError,
     InvalidRequestError,
+    OAuthError,
     StripeError)
 
 # DEPRECATED: These imports will be moved out of the root stripe namespace
--- a/stripe/api_requestor.py
+++ b/stripe/api_requestor.py
@@ -358,3 +358,37 @@ class APIRequestor(object):
     def handle_urllib2_error(self, err, abs_url):
         from stripe.http_client import Urllib2Client
         return self._deprecated_handle_error(Urllib2Client, err)
+
+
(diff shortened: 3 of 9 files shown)


brandur-stripe:
Hey @ob-stripe! Wow man, this is unbelievable work! Sorry for taking so much time to get back to you on this because it's really  obvious that you've done a ton of work here.

So this generally looks amazing: a much needed feature addition, good and correct-looking code, test cases, and even a new example!!

My only comment would be that it might be worth at least considering whether we could approach composition instead of inheritance. We basically introduce three new abstract classes here (the API requestor, the test case, and the API requestor test case), and the code might stay more clear and maintainable if we try to compose things instead. I haven't actually looked in depth to see how that would look though, so lemme know if it's a bad idea.

@kjc-stripe This is a sizable change and I know you have some opinions about the Python SDKs. Would you mind also looking this over?

ob-stripe:
Hey @brandur-stripe, thanks for the feedback :)

To clarify the current design:

- the only difference that **must** be addressed between API requests to `api.stripe.com` and to `connect.stripe.com` is error handling, as the error objects that are returned have different formats. `connect.stripe.com`'s errors are hashes with two attributes: `error` containing an error code and (optionally) `error_description` containing a human-readable description of the error.

  Everything else (authentication, parameter encoding, etc.) is compatible. The Connect docs say that the API key should be passed in a `client_secret` POST parameter [here](https://stripe.com/docs/connect/reference#post-token), but that's just because that's how it is in the OAuth spec. Passing the API key in the `Authorization: Bearer ...` header works just fine.

- there is no new abstract class for `APIRequestor` -- I just added a new `OAuthRequestor` class that directly derives from `APIRequestor` and overrides the `handle_api_error` and `interpret_response` methods to properly handle errors. (Admittedly, maybe I _should_ have created a new `AbstractRequestor` class and make both `APIRequestor` and `OAuthRequestor` derive from it.)

- for tests, I introduced two abstract classes: `AbstractStripeApiTestCase` and `AbstractRequestorRequestTests` and made both the existing `StripeApiTestCase` / `APIRequestorRequestTests` and the new `StripeOAuthTestCase` / `OAuthRequestorRequestTests` derive from them. I did it this way because test classes are instantiated directly by `unittest` and using class attributes was the only way I could think of to parameterize them with the appropriate `*Requestor` classes.

I don't necessarily have anything against composition over inheritance, but I'm having a hard time wrapping my head around how it would apply here. If you can give me a quick example of what you had in mind, I'll be happy to comment on it and/or make some changes to the code :)

brandur-stripe:
Hi @ob-stripe, I seem to have totally dropped the ball on this one. My apologies.

So I think I probably made an assertion that was a little too vague/wrong above, but in general, I think that nesting any kind of OO abstractions too deeply doesn't help things very much, and tends to make code confusing. I'm mostly trying to see what we can do to reduce the number of new classes required, especially when it comes to "abstract" classes in particular.

> there is no new abstract class for APIRequestor -- I just added a new OAuthRequestor class that directly derives from APIRequestor and overrides the handle_api_error and interpret_response methods to properly handle errors. (Admittedly, maybe I should have created a new AbstractRequestor class and make both APIRequestor and OAuthRequestor derive from it.)

I think this is fine as is.

> for tests, I introduced two abstract classes: AbstractStripeApiTestCase and AbstractRequestorRequestTests and made both the existing StripeApiTestCase / APIRequestorRequestTests and the new StripeOAuthTestCase / OAuthRequestorRequestTests derive from them. I did it this way because test classes are instantiated directly by unittest and using class attributes was the only way I could think of to parameterize them with the appropriate *Requestor classes.

Just to reduce the total number of classes, do you think that we could just eliminate the `AbstractStripeApiTestCase` abstraction and just have `OAuthTests` override `REQUESTOR_CLS_NAME` for itself? I think that `StripeOAuthTestCase` is only derived once and we'd only expect it to ever be derived once. Same for `AbstractRequestorRequestTests`.

This seems like a good idea to me because we eliminate some classes and the test class hierarchy maps a little better to the exist `APIRequestor` hierarchy. Thoughts?

ob-stripe:
> This seems like a good idea to me because we eliminate some classes and the test class hierarchy maps a little better to the exist APIRequestor hierarchy. Thoughts?

Yes, this is much better! I updated the PR to get rid of the abstract classes. Now we have:
- [`OAuthRequestor`](https://github.com/stripe/stripe-python/pull/287/files#diff-06e8e6a0633cc19f1fef8fbc334012cfR363) derives from `ApiRequestor`
- [`StripeOAuthTestCase`](https://github.com/stripe/stripe-python/pull/287/files#diff-fc89950d409dc41fcbbbb30f9009d957R233) derives from `StripeApiTestCase` (and overrides [`REQUESTOR_CLS_NAME`](https://github.com/stripe/stripe-python/pull/287/files#diff-fc89950d409dc41fcbbbb30f9009d957R234))
- [`StripeOAuthRequestorTestCase`](https://github.com/stripe/stripe-python/pull/287/files#diff-fc89950d409dc41fcbbbb30f9009d957R210) derives from `StripeRequestorTestCase` (and overrides [`REQUESTOR_CLS`](https://github.com/stripe/stripe-python/pull/287/files#diff-fc89950d409dc41fcbbbb30f9009d957R211))

[`StripeRequestorTestCase`](https://github.com/stripe/stripe-python/pull/287/files#diff-fc89950d409dc41fcbbbb30f9009d957R189) is a new parent class for [`APIRequestorRequestTests`](https://github.com/stripe/stripe-python/pull/287/files#diff-fc89950d409dc41fcbbbb30f9009d957R189) that did not exist previously, that takes care of setting up the requestor and mock client. Previously, `APIRequestorRequestTests` did that itself, but I couldn't make [`OAuthRequestorRequestTests`](https://github.com/stripe/stripe-python/pull/287/files#diff-f44daad033c958c345472ed9e1c63c2fR384) derive from `APIRequestorRequestTests` and simply override `REQUESTOR_CLS` as it would also inherit all the `test_*` methods.

brandur-stripe:
Love it! Thanks @ob-stripe.

brandur-stripe:
Released as 1.51.0.

Issue #206, opened 2016-02-22
Accept dependencies at initialization rather than relying on globals


justanr:
I'm needing to customize some behavior on API calls using the RequestClient however, `RequestsClient` doesn't accept a Session object and instead uses `requests.request`.

The other issue I'm having is that I then need to get this configured Client into the method calls. However, almost every API Resource simply creates a new instance of `APIRequestor` instead of accepting one at construction. But almost every time this is created, it's with things known at initialization.

I realize that there is `stripe.default_http_client` but I feel like that's a partial fix. And also of #184 and #192

brandur:
+1. This library could use some serious refactoring for better modularity.

@justanr I'm tagging this as "future" for now since unfortunately I'm not sure anyone will be taking a look immediately. Sorry about the trouble.

NahomAgidew:
@brandur Done!

ofek:
@justanr The functionality you requested was implemented in #288. Is that sufficient to remedy this issue?

justanr:
Probably, I'll pull it down and play with it. But this issue can be closed. Thanks!

PR #288, opened 2017-03-02
Allow reuse of an APIRequestor

ofek:
Addresses #206.

@brandur-stripe @ob-stripe Let me know if this is an acceptable change, then I'll work on tests.

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -99,8 +99,7 @@ def _serialize_list(array, previous):
 
 
 class StripeObject(dict):
-    def __init__(self, id=None, api_key=None, stripe_account=None,
-                 requestor=None, **params):
+    def __init__(self, id=None, api_key=None, stripe_account=None, **params):
         super(StripeObject, self).__init__()
 
         self._unsaved_values = set()
@@ -111,7 +110,6 @@ class StripeObject(dict):
 
         object.__setattr__(self, 'api_key', api_key)
         object.__setattr__(self, 'stripe_account', stripe_account)
-        object.__setattr__(self, 'requestor', requestor)
 
         if id:
             self['id'] = id
@@ -224,7 +222,7 @@ class StripeObject(dict):
     def request(self, method, url, params=None, headers=None):
         if params is None:
             params = self._retrieve_params
-        requestor = self.requestor or api_requestor.APIRequestor(
+        requestor = api_requestor.APIRequestor(
             key=self.api_key, api_base=self.api_base(),
             account=self.stripe_account)
         response, api_key = requestor.request(method, url, params, headers)
@@ -445,12 +443,10 @@ class ListableAPIResource(APIResource):
 
     @classmethod
     def list(cls, api_key=None, idempotency_key=None,
-             stripe_account=None, requestor=None, **params):
-        requestor = requestor or api_requestor.APIRequestor(
-            api_key,
-            api_base=cls.api_base(),
-            account=stripe_account
-        )
+             stripe_account=None, **params):
+        requestor = api_requestor.APIRequestor(api_key,
+                                               api_base=cls.api_base(),
+                                               account=stripe_account)
         url = cls.class_url()
         response, api_key = requestor.request('get', url, params)
         stripe_object = convert_to_stripe_object(response, api_key,
@@ -463,10 +459,8 @@ class CreateableAPIResource(APIResource):
 
     @classmethod
     def create(cls, api_key=None, idempotency_key=None,
-               stripe_account=None, requestor=None, **params):
-        requestor = requestor or api_requestor.APIRequestor(
-            api_key, account=stripe_account
-        )
+               stripe_account=None, **params):
+        requestor = api_requestor.APIRequestor(api_key, account=stripe_account)
         url = cls.class_url()
         headers = populate_headers(idempotency_key)
         response, api_key = requestor.request('post', url, params, headers)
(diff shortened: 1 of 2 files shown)


stripe/resource.py
```diff
@@ -114,6 +115,8 @@ def __init__(self, id=None, api_key=None, stripe_account=None, **params):
         if id:
             self['id'] = id
 
+        self.requestor = requestor
```
ob-stripe: You'd need to replace this with `object.__setattr__(self, 'requestor', requestor)`, or rename the variable to `_requestor`. Otherwise it's going to be treated as an API resource key (cf. https://github.com/ofek/stripe-python/blob/07ab2648a7f006893fd3337f5f939f0d3f89befc/stripe/resource.py#L127).

ob-stripe:
Hey @ofek, thanks a lot for submitting this PR.

Personally, I'm not a big fan of having a `requestor` argument in every method. We already have two different "classes" of arguments:
- normal arguments, sent as URL-encoded parameters in API requests
- header arguments (`api_key`, `stripe_account` and `idempotency_key`), sent in special HTTP headers in API requests

This would introduce a third class of arguments for a purely internal mechanism.

This isn't aimed specifically at your PR: given the existing design of the library, I'd probably have taken a similar approach. Overall, the change should be peaceful (I don't think we have plans to add a parameter named `requestor` in any API request ;) ) once you fix the issue where `requestor` is treated as an API resource key (I left an inline comment about that).

@brandur-stripe wdyt?

ofek:
@ob-stripe I made your requested change and now all tests pass but the flake command on py27 is failing https://travis-ci.org/stripe/stripe-python/jobs/207003997#L481

ofek:
With that change, now all tests pass.

brandur-stripe:
Hi @ofek, great work here! Thanks for putting it together. My apologies for the delayed response here.

So I'm inclined to agree with @ob-stripe above: I think this is absolutely useful and a good idea, but I'm not crazy about plumbing a new `requestor` through in a bunch of methods that already have some pretty messy overloaded arguments.

I did something similar to this recently [in stripe-ruby](https://github.com/stripe/stripe-ruby#configuring-a-client) whereby instead of mixing in more method parameters, we instead use a block to initiate a request under a specific client (i.e. "session" in the parlance of Requests):

``` ruby
conn = Faraday.new
client = Stripe::StripeClient.new(conn)
client.request do
  charge, resp = Stripe::Charge.retrieve(
    "ch_18atAXCdGbJFKhCuBAa4532Z",
  )
end
puts resp.request_id
```

It works by putting an active client to thread-local storage and then using that instead of the default client if it's available.

What do you think about trying something like this for stripe-python? It doesn't have the exact same block syntax that Ruby has, but I think a custom method that can be passed to a `with` block could be used to a similar effect. [Some information on how this might be accomplished here](http://preshing.com/20110920/the-python-with-statement-by-example/).

ofek:
@brandur-stripe Thanks for the advice!

Although I do see what you are hinting at, how would that allow users to use reuse an APIRequestor across Resource calls without the extra argument?

brandur-stripe:
> Although I do see what you are hinting at, how would that allow users to reuse an APIRequestor across Resource calls without the extra argument?

So I was thinking something like this:

``` python
with stripe.client(session):
    stripe.Charge.create(...)
```

Calling `stripe.client` puts `session` into some kind of [thread-local storage](https://docs.python.org/2/library/threading.html#threading.local) which is then accessed and used by `stripe.Charge.create` instead of the default session.

I'm still not positive this is a good idea, but it's an idea. Thoughts?

brandur-stripe:
And we'd probably want that to look even closer to the Ruby version by wrapping session up into our own construct so that we could technically use it beyond just the "requests" library:

``` python
session = requests.Session()
client = stripe.StripeClient(session)
with client.request():
    stripe.Charge.create(...)
```

ofek:
Ah I see, we are talking about different things it seems. The `requests.Session` reuse is somewhat tangential to the `api_requestor.APIRequestor` reuse.

Addressing issue #206 requires 2 things to be complete:

1. "I'm needing to customize some behavior on API calls using the RequestClient however, RequestsClient doesn't accept a Session object and instead uses requests.request."

This is the part you've already touched on and the very small change I've made to allow that is here [https://github.com/stripe/stripe-python/pull/288/files#diff-dc8bac176fde92ca11508c0877397454](https://github.com/stripe/stripe-python/pull/288/files#diff-dc8bac176fde92ca11508c0877397454).

2. "The other issue I'm having is that I then need to get this configured Client into the method calls. However, almost every API Resource simply creates a new instance of APIRequestor instead of accepting one at construction."

So what this means is that even though you can now give `api_requestor.APIRequestor` an instance of `http_client.RequestsClient` with a reused `requests.Session`, users can't actually reuse the aforementioned `api_requestor.APIRequestor` because every resource creates a new one.

Now, it is true that after my first change users could just set `stripe.default_http_client` so every new `api_requestor.APIRequestor` uses that. Though I chose to add the argument to offer users finer-grained control, and also I'm sort of weary of globals :)

@brandur-stripe @ob-stripe If you don't like the extra arguments, I could simply leave `stripe/resource.py` unchanged and my change to `stripe/http_client.py` would still fix the issue in question.

ofek:
@brandur-stripe @ob-stripe Sorry if I wasn't explicit in that comment edit but I left `stripe/resource.py` unchanged and so now only `stripe/http_client.py` has been altered, which still fixes the issue in question.

brandur-stripe:
Hi @ofek, apologies again for the slow response :/

So I think it would be nice to keep `APIRequestor` as an implementation detail as much as possible. It makes sense to me to allow a session to be injected given that it's plausible this could be shared with the rest of an application so it just seems like good API practice.

Lets see where we can get with just that though. If you really need to override particular request behavior, you can use the technique that you outlined for setting `stripe.default_http_client` and custom session. I am also very weary of globals, but I think this is the type of situation where it might be okay.

That said, I'm open for re-exploring this in the future if the need comes up. Thanks again for the contribution!

brandur-stripe:
Released as 1.49.0.

PR #292, opened 2017-03-26
Added support for the new Payout and RecipientTransfer objects

remi-stripe:
The Transfer object used to represent all movements of funds in Stripe. It split in three resources:
- Transfer: this describes the movement of funds between Stripe accounts and is specific to Stripe Connect.
- Payout: this describes the movement of funds from a Stripe account to a bank account, debit card or any future payout method.
- RecipientTransfer: this describes the movement of funds from a Stripe account to a Recipient's card or Bank Account. This is here for legacy reasons and can only be accessed from an expanded BalanceTransaction.

This change is behind an API version so old API versions would still use the Transfer object for everything while new API version would see the split.

This applies beyond the new object as some properties/methods are removed from Transfer and other properties are renamed on other objects.


brandur-stripe:
LGTM.

Thanks Remi.

remi-stripe:
@brandur-stripe Okay I added the new `RecipientTransfer` resource. I haven't been able to figure out how I can even test this to be honest since the Python library mocks all the requests. Do you have confidence that the `RecipientTransfer` would be expanded properly and if not some ideas on how I could test this?

brandur-stripe:
@remi-stripe Minor linting problems from Travis:

```
$ if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then flake8 stripe; fi
stripe/resource.py:949:1: E302 expected 2 blank lines, found 1
stripe/resource.py:952:1: E302 expected 2 blank lines, found 1
```

> Do you have confidence that the RecipientTransfer would be expanded properly and if not some ideas on how I could test this?

Yeah, I think there's enough existing usages of this that I'm not too worried about it.

Pending linting problem, LGTM.

stripe/resource.py
```diff
@@ -936,6 +946,11 @@ def transfers(self, **params):
         return transfers
 
 
+# This resource can only be instantiated when expanded on a BalanceTransaction
+class RecipientTransfer():
```
brandur-stripe: I suspect you're still going to have to inherit from something like `APIResource` for this to do anything. Right now it's just a perfectly vanilla class.

brandur-stripe: Looking through this again, I think the main thing you need is a `construct_from` implementation, so you might be okay just inheriting from `StripeObject`. We should at least run a manual test to make sure this one works though when the API is available.

brandur-stripe:
So I think there's some possibility that we might want to export `RecipientTransfer` in `__init__` so that people could say, compare an object's type to it, but there's also a reasonable possibility that it'll never be needed either. I could see just leaving as is and exporting it if someone asks.

remi-stripe:
fixed. PTAL

stripe/resource.py
```diff
@@ -936,6 +946,11 @@ def transfers(self, **params):
         return transfers
 
 
+# This resource can only be instantiated when expanded on a BalanceTransaction
+class RecipientTransfer(APIResource):
```
brandur-stripe: Hah, sorry for the continue churn, but I just noticed that like in stripe-ruby, this can and probably should inherit `StripeObject` and not `APIResource`.

Other than that, LGTM.

remi-stripe:
Fixed!

brandur-stripe:
Cool. LGTM.

brandur-stripe:
Released as 1.53.0.

PR #294, opened 2017-03-29
Fixed bug urlparse

null-none:


ob-stripe:
Hey @null-none, it seems like you're trying to fix a Python 2 vs. 3 compatibility issue. If you try to import the raw `stripe/api_requestor.py` file from Python 3, it will fail because `urlparse` no longer exists in Python 3's standard library, and `urllib.parse` should be used instead.

That said, you should not be importing the library's raw source code directly with Python 3. We use `2to3` to automatically update the code for Python 3 (cf. [here](https://github.com/stripe/stripe-python/blob/023d742a163d4a814ac63ac4533137b2917db0eb/setup.py#L61)).

You should either install the library using `pip` or, if you want to install from source, run `python setup.py install` to make sure the source files are processed by `2to3`.

I'm going to close the issue but feel free to reply if you have any questions!

PR #299, opened 2017-04-07
Remove raw PANs from our test suite

remi-stripe:
This PR aims to use the new test tokens such as `tok_visa` instead or hardcoding raw PANs in the test suite. It also cleans up the example to not mention raw PANs.

r? @brandur-stripe

### DIFF
--- a/stripe/test/helper.py
+++ b/stripe/test/helper.py
@@ -11,20 +11,10 @@ import stripe
 
 NOW = datetime.datetime.now()
 
-DUMMY_CARD = {
-    'number': '4242424242424242',
-    'exp_month': NOW.month,
-    'exp_year': NOW.year + 4
-}
-DUMMY_DEBIT_CARD = {
-    'number': '4000056655665556',
-    'exp_month': NOW.month,
-    'exp_year': NOW.year + 4
-}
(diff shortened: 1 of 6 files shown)


stripe/test/helper.py
```diff
@@ -12,16 +12,12 @@
...
-DUMMY_DEBIT_CARD = {
-    'number': '4000056655665556',
-    'exp_month': NOW.month,
-    'exp_year': NOW.year + 4
-}
+DUMMY_CARD = 'tok_visa'
+
+DUMMY_DEBIT_CARD = 'tok_visa_debit'
+
+DUMMY_EXPIRED_CARD = 'tok_chargeDeclinedExpiredCard'
```
brandur-stripe: Could we just suffix these with `_TOKEN` so that when you're reading the referencing code it's obvious that it's a token and not some kind of card hash?

remi-stripe: Done! PTAL!

brandur-stripe:
LGTM. Lets wait on the backend fix to merge, but nice work.

remi-stripe:
@brandur-stripe  PTAL

brandur-stripe:
Looks good. Thanks!

PR #301, opened 2017-04-12
fixed README.md to say package rather than gem

ian-stripe:
Pretty self explanatory :)


ian-stripe:
@brandur r?

brandur-stripe:
Ah man. That one is my fault.

Thanks!

PR #305, opened 2017-05-02
[WIP] Add stickiness/passthrough for API version through resources

lskillen:
Hello!  This is a first contribution to `stripe-python` but we've (my company has) got Stripe integrated into production and I'm a frequent contributor to [dj-stripe](https://github.com/kavdev/dj-stripe), a library for interfacing Django with Stripe via this library. :-)

This is being submitted as a Work In Progress in order to get some initial feedback before proceeding with additional work in getting it landed.  So the disclaimer here is that there are *no* additional test cases added yet (but we'll do that prior to completion), although the full test suite does execute successfully.

The background is that in [dj-stripe](https://github.com/kavdev/dj-stripe) we're processing events using the `api_version` that is indicated in the event messages, which as of the current version means temporarily switching the global `stripe.api_version` to that (via a context manager) and then switching it back again to whatever was previously configured.  As you can [see here](https://github.com/kavdev/dj-stripe/blob/d7c6cc16d74d358a815661b2a413d3f3df169be0/djstripe/stripe_objects.py#L927-L928).

This generally works but we're worried about changing globals in the application state, especially for situations in which  the event webhooks are processed in the same multi-threaded server where other (normal) workload with Stripe happens (e.g. creating customers on user sign-up).

The thought is that in the moments when the context manager has changed the global `api_version`, it will affect any other place utilising Stripe.  This isn't likely to cause a lot of issues, but it certainly doesn't feel right, and obviously with billing related is desired to be as water-tight as possible. ;-)

So, this PR builds on a3373ac and suggests to add `api_version` as a passthrough argument for resources *and* adds additional stickiness for it when resources are constructed.  A contentious point of this PR is the addition of a slightly more uniform approach to arguments that are ultimately passed through to the `APIRequestor` via a factory classmethod that performs the work.

On one hand it greatly tidies up these arguments, and means the logic for `APIRequestor` is contained to one place, and would mean simplify future changes to this area.  On the other hand it removes some of the explicitness of the passthrough arguments, which some will see as bad (and it may break clients that use positional arguments, even though they probably shouldn't).  In saying that, there was some varying consistency on exactly what arguments were supported where, and this does fix that.

Thoughts and feedback greatly welcomed - Thank you in advance!

P.S. I'm not opposed to reversing any part of the changes, but supporting passthrough on some level would be greatly appreciated for the sake of correctness. :-)

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -6,7 +6,7 @@ from copy import deepcopy
 from stripe import api_requestor, error, oauth, util, upload_api_base
 
 
-def convert_to_stripe_object(resp, api_key, account):
+def convert_to_stripe_object(resp, api_key, account, api_version=None):
     types = {
         'account': Account,
         'alipay_account': AlipayAccount,
@@ -46,7 +46,9 @@ def convert_to_stripe_object(resp, api_key, account):
     }
 
     if isinstance(resp, list):
-        return [convert_to_stripe_object(i, api_key, account) for i in resp]
+        return [convert_to_stripe_object(i, api_key, account,
+                                         api_version=api_version)
+                for i in resp]
     elif isinstance(resp, dict) and not isinstance(resp, StripeObject):
         resp = resp.copy()
         klass_name = resp.get('object')
@@ -54,7 +56,8 @@ def convert_to_stripe_object(resp, api_key, account):
             klass = types.get(klass_name, StripeObject)
         else:
             klass = StripeObject
-        return klass.construct_from(resp, api_key, stripe_account=account)
+        return klass.construct_from(resp, api_key, stripe_account=account,
+                                    api_version=api_version)
     else:
         return resp
 
@@ -101,7 +104,8 @@ def _serialize_list(array, previous):
 
 
 class StripeObject(dict):
-    def __init__(self, id=None, api_key=None, stripe_account=None, **params):
+    def __init__(self, id=None, api_key=None, stripe_account=None,
+                 api_version=None, **params):
         super(StripeObject, self).__init__()
 
         self._unsaved_values = set()
@@ -111,6 +115,7 @@ class StripeObject(dict):
         self._previous = None
 
         object.__setattr__(self, 'api_key', api_key)
+        object.__setattr__(self, 'api_version', api_version)
         object.__setattr__(self, 'stripe_account', stripe_account)
 
         if id:
@@ -183,16 +188,19 @@ class StripeObject(dict):
             self._unsaved_values.remove(k)
 
     @classmethod
-    def construct_from(cls, values, key, stripe_account=None):
-        instance = cls(values.get('id'), api_key=key,
+    def construct_from(cls, values, key, stripe_account=None,
+                       api_version=None):
+        instance = cls(values.get('id'), api_key=key, api_version=api_version,
                        stripe_account=stripe_account)
-        instance.refresh_from(values, api_key=key,
+        instance.refresh_from(values, api_key=key, api_version=api_version,
                               stripe_account=stripe_account)
         return instance
 
-    def refresh_from(self, values, api_key=None, partial=False,
-                     stripe_account=None):
+    def refresh_from(self, values, api_key=None,
+                     partial=False, stripe_account=None,
+                     api_version=None):
         self.api_key = api_key or getattr(values, 'api_key', None)
+        self.api_version = api_version or getattr(values, 'api_version', None)
         self.stripe_account = \
             stripe_account or getattr(values, 'stripe_account', None)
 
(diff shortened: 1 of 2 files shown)


stripe/resource.py
```diff
@@ -759,23 +786,22 @@ class Customer(CreateableAPIResource, UpdateableAPIResource,
...
         params['customer'] = self.id
-        ii = InvoiceItem.create(self.api_key,
-                                idempotency_key=idempotency_key, **params)
+        ii = InvoiceItem.create(idempotency_key=idempotency_key, **params)
         return ii
 
     def invoices(self, **params):
         params['customer'] = self.id
-        invoices = Invoice.list(self.api_key, **params)
+        invoices = Invoice.list(**params)
```
brandur-stripe: Doesn't this change/break the old behavior. It seems like before we were using the API key set on the customer and now we're just using one that gets passed in with `params`.

lskillen:
@brandur-stripe Yup, absolutely no problem, and thank you for the feedback!

Even while writing it I have to admit that I wasn't particularly keen on breaking compatibility, but I wasn't sure if the emphasis was on maintaining it specifically for positional vs keyword arguments.  You're also right that explicit is better than implicit, even when it's more onerous to implement/maintain. ;-)

Leaves this with me to tweak this a bit and add a few tests around the parameter passing.

brandur-stripe:
> Leave this with me to tweak and I'll also add a few tests around the parameter passing.

Thanks!

And yeah, just with regards to maintenance: I think it's okay if only because we tend to rarely add new things to these methods. Beyond API keys, API version, account, etc. there isn't much else that I imagine ever going in there.

lskillen:
@brandur-stripe:

OK, rebased and added a pinch of explicitness back in (admittedly it isn't as tidy, but it's definitely more *obvious* now). This is still WIP (could add a few more tests) but thought I'd checkpoint with you to see if this approach is a bit more palatable. :-)  I've added some *very* minimal tests just for StripeObject, but subject to a thumbs up I can add more for the actual stickiness and some for the passthrough in other places.  I also noted (and fixed) your concern about the API key from Customer being dropped when passing to Invoice - partially due to the zealous drive to remove the arguments in the first place.

I've touched more code than I would have liked for the sake of consistency, and even then I couldn't get it as consistent as I would like with regards to argument order, but I think it's pretty decent.  The bits that I didn't touch that probably should have the API arguments are the places that use `self.request` and `self.refresh_from`, but I figured that since they're typically refreshing the current object it should be mostly OK to skip for now, since  the current object will have the API key and version it was originally fetched with.

Thanks in advance for reviewing again.

lskillen:
@brandur @brandur-stripe @ob-stripe ping. :-)

stripe/resource.py
```diff
@@ -6,7 +6,7 @@
 from stripe import api_requestor, error, oauth, util, upload_api_base
 
 
-def convert_to_stripe_object(resp, api_key, account):
+def convert_to_stripe_object(resp, api_key, account, api_version=None):
```
brandur-stripe: Hah, the positional parameter versus named parameter mismatch here isn't amazing, but +1 in that we should be moving toward named, so this is okay.

stripe/resource.py
```diff
@@ -367,8 +394,11 @@ def instance_url(self):
 
 class ListObject(StripeObject):
 
-    def list(self, **params):
-        return self.request('get', self['url'], params)
+    def list(self, api_key=None, stripe_account=None, api_version=None,
+             **params):
```
brandur-stripe: Thanks for adding in the other parameters as well!

brandur-stripe:
@lskillen Hi Lee, my apologies; it's been a busy few weeks, and I've been doing a terrible job staying on top of a few of these OSS PRs.

I think the new approach is great. Thank you for converting some of the other options over to parameters as well! This produces a chatty PR, but in the long run I think it's going to help everyone.

This looks mostly finished to me. The one thing I'd consider is moving the tests on the new API version behavior out of tests like `test_access` and into their own thing (`test_api_version_persisted` or such). That way, we're very explicit about keeping the behavior running, and it'll be a little harder to introduce an accidental regression. What do you think?

Thanks again for tackling this! Amazing work.

lskillen:
@brandur-stripe Sounds good to me, and thank you for reviewing.  I'll refactor the tests a little as suggested and add a pinch more for checking the passthrough.  Will ping you again when I get it committed and pushed!

brandur-stripe:
@lskillen Sounds good! Thanks for your patience with us here.

lskillen:
Just a note to let you know this hasn't been forgotten about, I've just been beyond busy.  I do have some of the changes completed, but just need that final push to get them ready.  Annoyingly it's about an hour worth of work, so I'll hopefully get through it this week. :-)

ob-stripe:
Going to close the PR, as we added this feature in #344 (released in version 1.65.0).

PR #306, opened 2017-05-09
code cleanup

ofek:


brandur-stripe:
@ofek This looks sweet, thanks!

Just one note: I think that technically removing those `**params` globs on the `delete_discount` methods is a breaking change because if someone happens to be passing another argument into them, it won't work anymore. What do you think about leaving `**params` there for now and putting a comment in there to take those out on the next major version rev?

ofek:
@brandur-stripe @ob-stripe Changes are complete but there is a major unrelated issue now. see https://github.com/stripe/stripe-python/issues/307

brandur-stripe:
Thanks @ofek. Mind rebasing now that #308 is in? Thanks!

ofek:
Rebased.

brandur-stripe:
Looks good to me. Thanks!

PR #309, opened 2017-05-10
upgrade PyPy3 in Travis

ofek:


ob-stripe:
I really appreciate the effort to fix the CI, but I'm not thrilled about "hacking" Travis' CI environment like this and relying on external binaries. On the other hand, Travis is usually pretty slow to update their environments :/

wdyt @brandur-stripe? Since both `pypy` and `pypy3` are flagged as `allow_failures` in our Travis configuration, we could just wait until Travis fixes the PyPy environments, but that could take a while...

brandur-stripe:
> I really appreciate the effort to fix the CI, but I'm not thrilled about "hacking" Travis' CI environment like this and relying on external binaries. On the other hand, Travis is usually pretty slow to update their environments :/

Yeah, +1. I'm used to working around Travis occasionally, but this seems especially hacky and may be a little difficult to maintain.

I must admit that I don't have a concrete understanding of all the ins and outs of Python's build system, do you know if the next Travis upgrade for pypy and pypy3 quite likely to fix the build problems? If so, yeah, I'd probably lean toward just waiting, even if it does result in failing builds in the interim.

ofek:
What happened here is new pip + setuptools requires py3.3+ but Travis' PyPy3 is based on Python 3.2, hence the failure. This PR uses latest PyPy3 based on Python 3.5. Other projects do this:

https://github.com/biopython/biopython/blob/3b03cb4c1e70a1e1ea29729f6ad3765233f6dcd7/.travis.yml#L54-L55
https://github.com/ofek/bit/blob/2b890e1c31edd10d33ba645088adf38e097fa296/.travis.yml#L13

brandur-stripe:
@ofek Thanks for the extra context. I'm still leaning slightly in favor of just waiting for a Travis upgrade.

ofek:
Up to you, but that will take a long time :)

deontologician-stripe:
Closing since we were leaning away from merging this and it's pretty been outstanding for a while. Let me know if anyone feels strongly about re-opening

Issue #307, opened 2017-05-09
Install started failing for users of older pip


ofek:
@brandur-stripe @ob-stripe Requests 2.14.1 just released and now requires pip >= 8.1.2 and a newish version of setuptools https://github.com/kennethreitz/requests/issues/4006

As of today, stripe-python will fail to install for users of older packaging infrastructure, including Travis.
https://travis-ci.org/stripe/stripe-python/jobs/230553497#L497
https://github.com/kennethreitz/requests/pull/4007#issuecomment-300364756

We should deploy an emergency release limiting requests to version 2.13.0 and documentation should be added detailing that in the next major release there will be a minimum required version of pip.

ofek:
A huge number of projects' builds broke today, not just use!

ofek:
https://github.com/stripe/stripe-python/pull/308 fixes our Travis builds.

brandur-stripe:
Thanks for the report here @ofek and the initiative to get the fix in.

Just as a quick explanatory follow up: it seems that Requests 2.14.0 was the version that broke everybody. They followed up with 2.14.1 which is designed to be adaptable to more environments, and with the pip and setuptools upgrades in #308, things seem to be relatively happy again.

Closing this out.

ofek:
Well, #308 only fixes Travis. Our users are still affected even with 2.14.1. https://github.com/kennethreitz/requests/pull/4007#issuecomment-300364756

ofek:
@brandur-stripe @ob-stripe Without releasing a new version that limits requests to 2.13.0 many of our users cannot install this. Unless we're ok with just adding a note to doc saying to upgrade pip and setuptools?

brandur-stripe:
@ofek Good point. I'd be fine with limiting the ceiling on the Requests version. We'll have to kill it eventually lest it start causing transitive dependency hell problems in the future, but hopefully by the time we unlock it more people will be upgraded on pip/setuptools.

I notice that right now we're already doing some conditional shenanigans around versioning Requests in our `setup.py`:
https://github.com/stripe/stripe-python/blob/a84a0c80b2a325a989b14ab9ecf3d59b7b74be6d/setup.py#L20,L29

Do you know if there's an additional check on pip/setuptools versions that we could also add in there with the capped Requests? That might be the best of both worlds.

brandur-stripe:
From the other thread:

[...]

PR #312, opened 2017-05-11
Reverts additional constraint on Requests

brandur-stripe:
Revert an additional constraint on Requests that was added because they
broke old versions of pip/setuptools. Since then a new version seems to
have been released that's more broadly compatible, so lets loosen this
constraint again to make dependency problems for our users less likely.

Reverts #311. See #307 for more context.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,23 +7,6 @@ try:
 except ImportError:
     from distutils.core import setup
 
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Released as 1.55.2.

PR #320, opened 2017-05-29
Fix pycurl client with Python 3

### LINKED ISSUE #317: PycurlClient doesn't work with Python 3
When using the Stripe Python API in combination with PycurlClient and Python 3, the requests made to the API cannot complete.

In the `http_client.py` file, the first line of the `_handle_request_error` function the value of e is `error(23, 'Failed writing header')`. The error results in another error because the request handling code doesn't work with this specific error value (the error object does not support indexing).

The cause of this problem seems the be the use if StringIO instead of BytesIO:
https://stackoverflow.com/a/23610279/138103

To test run any method that connects to the API in Python 3 after running the following:
```python
stripe.default_http_client = stripe.http_client.PycurlClient()
```

Python Version: 3.5.3
Library Version: 1.55.2

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fixes #317.

There are two issues preventing the pycurl client from working with Python 3:

- we use `StringIO` when we should be using `BytesIO`

- we don't access the `pycurl.error` parameters correctly

This PR fixes both issues.

### DIFF
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -239,8 +239,8 @@ class PycurlClient(HTTPClient):
         return dict((k.lower(), v) for k, v in dict(headers).iteritems())
 
     def request(self, method, url, headers, post_data=None):
-        s = util.StringIO.StringIO()
-        rheaders = util.StringIO.StringIO()
+        b = util.io.BytesIO()
+        rheaders = util.io.BytesIO()
         curl = pycurl.Curl()
 
         proxy = self._get_proxy(url)
@@ -265,7 +265,7 @@ class PycurlClient(HTTPClient):
         # pycurl doesn't like unicode URLs
         curl.setopt(pycurl.URL, util.utf8(url))
 
-        curl.setopt(pycurl.WRITEFUNCTION, s.write)
+        curl.setopt(pycurl.WRITEFUNCTION, b.write)
         curl.setopt(pycurl.HEADERFUNCTION, rheaders.write)
         curl.setopt(pycurl.NOSIGNAL, 1)
         curl.setopt(pycurl.CONNECTTIMEOUT, 30)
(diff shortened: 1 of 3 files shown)


stripe/http_client.py
```diff
@@ -282,22 +282,22 @@ def request(self, method, url, headers, post_data=None):
...
         rcode = curl.getinfo(pycurl.RESPONSE_CODE)
 
-        return rbody, rcode, self.parse_headers(rheaders.getvalue())
+        return rbody, rcode, self.parse_headers(rheaders.getvalue().decode('utf-8'))
 
     def _handle_request_error(self, e):
-        if e[0] in [pycurl.E_COULDNT_CONNECT,
-                    pycurl.E_COULDNT_RESOLVE_HOST,
-                    pycurl.E_OPERATION_TIMEOUTED]:
+        if e.args[0] in [pycurl.E_COULDNT_CONNECT,
```
ob-stripe: Using `args` seem to be the canonical way of retrieving the error code and message from the `pycurl.error` instance, e.g. https://github.com/pycurl/pycurl/blob/fa25829377f1303010001e0329828891c1c61566/tests/xferinfo_cb_test.py#L53.

brandur-stripe: Nice research. Definitely seems less magical and more readable to me ...

stripe/http_client.py
```diff
@@ -282,22 +282,22 @@ def request(self, method, url, headers, post_data=None):
...
+                         pycurl.E_COULDNT_RESOLVE_HOST,
+                         pycurl.E_OPERATION_TIMEOUTED]:
```
brandur-stripe: Just wanted to flag this as it looks like GitHub's detecting some whitespace changes. Was that on purpose?

stripe/util.py
```diff
@@ -25,12 +26,6 @@
 ]
 
 try:
-    # When cStringIO is available
-    import cStringIO as StringIO
-except ImportError:
-    import StringIO
```
brandur-stripe: Great to see one of these gone!

brandur-stripe:
@ob-stripe Thank you so much for tackling this one! (I tried to do it last week, but was having trouble getting to a working reproduction of the problem due to some cluelessness on my part around how the packaging system works.)

I think this is pretty much good to go. CI is failing because you have one flake8 violation:

```
$ if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then flake8 stripe; fi
stripe/http_client.py:288:80: E501 line too long (84 > 79 characters)
The command "if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then flake8 stripe; fi" exited with 1.
```

Can you take a look at that and we can get this in? Thanks!

stripe/http_client.py
ob-stripe: Yes, I just realigned the array elements.

ob-stripe:
Yeah, sorry about that! Should be fixed now, ptal.

brandur-stripe:
Thanks @ob-stripe!

Issue #315, opened 2017-05-22
Imports are done incorrectly in Stripe Python library


royleban:
The Python API package does imports incorrectly. This is what you do:

`from stripe.resource import ...
`

This is what you should do ...

`from resource import ...
`

Why does it matter? Right now, stripe must be at the root of the web server. There is no reason for this and many web sites have packages such as Stripe in a subdirectory. Changing the imports would enable this without any disadvantages.

Support noted that there is an expectation that Stripe is installed globally. That is impossible on Google App Engine and is also not a best practice for people who only have one web server. It makes the server code more fragile because installs must happen before it is moved. Aside from that, fixing this problem would not cause any problems for globally-installed copies of Stripe.

Interestingly, there are comments in your __init__.py which indicate that your developers even know about this bug and haven't fixed it yet.

brandur-stripe:
Hi Roy,

I must admit that I'm not a Python expert and could use some guidance here. Would you mind citing a few references that suggest that this is best practice?

I took a look around at a few popular Python packages to see what's done elsewhere. Requests uses relative paths (different than what you suggested, but involves prefixing modules with a `.`), but [Scrapy (source)](https://github.com/scrapy/scrapy/blob/d847e65aaef68009b1ba4c12e45f34a7783dcc5b/scrapy/http/__init__.py#L10-L12) and [Twisted (source)](https://github.com/twisted/twisted/blob/69b30c9a9b74f44180b8118ce9fde69a70bcfe15/src/twisted/web/distrib.py#L25-L27) are using absolute paths like is done in stripe-python, which leads me to believe that it's not that unusual.

I don't mind moving to something like relative paths if it's helpful. I took a shot at this in #321 to see what that would look like.

> Interestingly, there are comments in your init.py which indicate that your developers even know about this bug and haven't fixed it yet.

Can you reference a comment in particular? I read through it but couldn't find anything that suggested that. Whether or not this is best practice or not, I don't think we can really call it a bug.

ob-stripe:
I'm not sure if there's a strong consensus. I think that previously, PEP 8 strongly discouraged from using relative imports. In its [current version](https://www.python.org/dev/peps/pep-0008/#imports), it seems to merely recommend absolute imports over relative:

> Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on `sys.path`)

[...]

PR #321, opened 2017-05-31
Try relative Python imports

brandur-stripe:
Moves from absolute imports to relative imports (i.e. prefix with a `.`
instead of a fully qualified `stripe.`).

I didn't change the test suite because it's not necessary and we might
want to move that out of the `stripe` module and into a `tests`
directory eventually anyway.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -22,7 +22,7 @@ app_info = None
 log = None
 
 # Resource
-from stripe.resource import (  # noqa
+from .resource import (  # noqa
     Account,
     AlipayAccount,
     ApplePayDomain,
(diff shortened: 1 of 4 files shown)


brandur-stripe:
@ob-stripe Any thoughts on this? :)

ob-stripe:
\o/

And yeah, we ought to move the tests into their own directory at some point.

ofek:
I disapprove of this change; it's less pythonic with no gain. I believe the person who opened the issue was simply not installing libraries correctly in GAE (I use it myself). Though, if this is approved, be sure to use `from __future__ import absolute_import`.

+1 to dedicated test directory though!

ob-stripe:
Hey @ofek, thanks for the feedback! This tentative PR was motivated by the discussion in #315. The discussion is still ongoing, but for now @brandur-stripe and myself are leaning towards sticking with absolute imports as PEP 8 recommends.

brandur-stripe:
+1 what @ob-stripe said. I wanted to see what this looked like, but we talk it through, and it doesn't seem like the way to go. I'm going to close this out for the time being.

PR #323, opened 2017-06-02
Update OAuth implementation

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This PR updates the OAuth implementation to be much closer to the [stripe-ruby implementation](https://github.com/stripe/stripe-ruby/pull/523/files):

- no more dedicated `OAuthRequestor` class. The `APIRequestor` class can now interpret both API and OAuth errors

- dedicated exception classes for each OAuth error type

This is technically a breaking change, but we never documented this feature (and won't until ~all libraries support it) so I don't think this necessarily warrants a major version bump.

The one thing I'm not happy about is the `OAuthInvalidRequestError` class name, to differentiate it from the existing `InvalidRequestError` class. I think it might be worth it moving all the OAuth errors in a different module, e.g. `oauth_error`, so they have their own namespace and we can reuse the `InvalidRequestError` class name.

wdyt?

### DIFF
--- a/stripe/api_requestor.py
+++ b/stripe/api_requestor.py
@@ -7,7 +7,7 @@ import urlparse
 import warnings
 
 import stripe
-from stripe import error, http_client, version, util
+from stripe import error, oauth_error, http_client, version, util
 from stripe.multipart_data_generator import MultipartDataGenerator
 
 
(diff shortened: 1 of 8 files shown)


stripe/api_requestor.py
```diff
@@ -151,38 +151,88 @@ def request(self, method, url, params=None, headers=None):
...
-            err = resp['error']
+            error_data = resp['error']
         except (KeyError, TypeError):
             raise error.APIError(
                 "Invalid response object from API: %r (HTTP response code "
                 "was %d)" % (rbody, rcode),
                 rbody, rcode, resp)
 
+        err = None
+        if isinstance(error_data, basestring):
```
brandur-stripe: Maybe just add a comment here on why we're doing this. I think that the minor structural difference between an OAuth and API error is pretty subtle unless you already know about it.

brandur-stripe:
@ob-stripe Looks good to me! This feels like it strips some complexity out and standardizes on some standard paths, so big +1.

> The one thing I'm not happy about is the OAuthInvalidRequestError class name, to differentiate it from the existing InvalidRequestError class. I think it might be worth it moving all the OAuth errors in a different module, e.g. oauth_error, so they have their own namespace and we can reuse the InvalidRequestError class name.

Yeah, in my opinion adding a new module here would be fine.

> This is technically a breaking change, but we never documented this feature (and won't until ~all libraries support it) so I don't think this necessarily warrants a major version bump.

Good point. I just ran a Splunk query and it doesn't look there's any activity on `/oauth/token` coming from stripe-python right now (checked the last 24 hours and re-running for the last week now), so yeah it might be fine to refactor this a bit.

brandur-stripe:
Quick follow up: no requests from stripe-python to `/oauth/token` over the last week, so I think we're okay.

ob-stripe:
Alright, I moved the OAuth error classes into their own `oauth_error` module.

ptal @brandur-stripe

brandur-stripe:
Nice!

Just out of curiosity, any thoughts on `oauth_error` versus `error.oauth` or `error.oauth_error`? I don't mind that much either way, but I just wanted to make sure we thought about it.

ob-stripe:
@brandur-stripe Great point, I hadn't thought of this. I agree that `error.oauth` would be a bit better so I gave it a try, but I ran into some issues. Doing this requires turning `error` from a module (`error.py`) into a package (`error/__init__.py`) so we can create the `oauth.py` submodule inside the `error` namespace.

This works on Python 2, but apparently causes circular import errors on Python 3 for some reason. This is probably fixable but I'm wary that changing how the imports are done could cause some unforeseen side-effects, so I'd rather play it safe and reverted to using the `oauth_error` module.

brandur-stripe:
> This works on Python 2, but apparently causes circular import errors on Python 3 for some reason. This is probably fixable but I'm wary that changing how the imports are done could cause some unforeseen side-effects, so I'd rather play it safe and reverted to using the oauth_error module.

Thanks for looking into it; that sounds good to me. I don't think there's a huge ergonomic different either way, so it seems fine to leave as `oauth_error`.

brandur-stripe:
As discussed, released as 1.59.0.

PR #325, opened 2017-06-02
Re-use `pycurl.Curl` objects between requests

### LINKED ISSUE #324: PycurlClient does not support TCP keep-alive
The RequestsClient uses a session with which one is able to get TCP keep-alive by creating a single client per process:

`stripe.default_http_client = stripe.http_client.RequestsClient()`

After the initial stripe API handshake, the API calls can be completed much quicker. For long running processes (say a web-application) this yields a great speed up. In our case it cuts the costs of API calls in half.

Doing the same thing with the PycurlClient doesn't work however. The following setup will still discard the connection after every single request:

`stripe.default_http_client = stripe.http_client.PycurlClient()`

To facilitate this the Curl instance would have to be kept around if I'm not mistaken. So basically the PycurlClient might want to use curl as a kind of session, just like the RequestsClient does:

```python
class PycurlClient(object):
    def __init__(..., curl=None):
        self.curl = curl or pycurl.Curl()
    def request(...):
        curl = self.curl
```

Python Version: 3.5.3
Library Version: 1.57.1

brandur-stripe:
This is poorly documented, but it seems that Pycurl will maintain
established connections between requests [1]. Here we move the
initialization of the `Curl` object to the constructor and keep almsot
everything else the same, with the exception of the addition of a
`self._curl.reset()` call before any request to make sure that our
options are zeroed (`reset` doesn't reset open connections).

Fixes #324.

r? @ob-stripe 

[1] http://pycurl.io/docs/latest/curlobject.html#pycurl.Curl.reset

### DIFF
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -223,6 +223,10 @@ class PycurlClient(HTTPClient):
     def __init__(self, verify_ssl_certs=True, proxy=None):
         super(PycurlClient, self).__init__(
             verify_ssl_certs=verify_ssl_certs, proxy=proxy)
+
+        # Initialize this within the object so that we can reuse connections.
+        self._curl = pycurl.Curl()
+
         # need to urlparse the proxy, since PyCurl
         # consumes the proxy url in small pieces
         if self._proxy:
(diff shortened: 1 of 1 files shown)


ob-stripe:
Nice find! I was thinking we'd need to manually reset the options before every request, but this is much more convenient.

lgtm!

brandur-stripe:
Thanks @ob-stripe!

brandur-stripe:
Released as 1.58.0.

PR #326, opened 2017-06-08
Add support for ephemeral keys

grey-stripe:
This adds the `EphemeralKey` resource to the bindings.  Currently, the only supported operations are creation and deletion.

Creation is a bit of a special case, in that it requires `Stripe-Version` to be set explicitly (because this will be requested by the backend for use in the frontend, which may be expecting an arbitrary API version).  This means that it is an error if the header is set by the globally configured API version, and it must be provided explicitly in the `create` call.  As such, `create` will throw a `ValueError` if `api_version` is not specified.

Everything else should be fairly standard.


stripe/__init__.py
```diff
@@ -39,6 +39,7 @@
     Customer,
     Dispute,
     Event,
+    EphemeralKey,
```
brandur-stripe: Sorry, this is the pedantic part of me, but mind flipping the two "E"s here so that it's fully alphabetized? Thanks!

stripe/resource.py
```diff
@@ -898,6 +899,31 @@ class Event(ListableAPIResource):
     pass
 
 
+class EphemeralKey(DeletableAPIResource):
```
brandur-stripe: And same with the location of `EphemeralKey` versus `Event` in this file.

stripe/test/resources/test_ephemeral_keys.py
```diff
@@ -0,0 +1,57 @@
...
+            api_version='2017-06-05',
+            account=None
+        )
+
+    def test_create_without_global_version(self):
+        stripe.api_version = None
+        self.do_version_tests()
+
+    def test_create_with_global_version(self):
+        stripe.api_version = '2017-05-25'
```
brandur-stripe: Just to prevent state leakage in case of future library changes, mind doing something here like:

``` python
old_api_version = stripe.api_version
stripe.api_version = ...

finally:
    stripe.api_version = old_api_version
```

stripe/test/resources/test_ephemeral_keys.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def test_create(self):
+        stripe.EphemeralKey.create(customer='cus_123',
+                                   api_version='2017-05-25')
+
+        self.requestor_class_mock.assert_called_with(
+            None,
+            api_version='2017-05-25',
+            account=None
+        )
```
brandur-stripe: Rather than exposing `request_class_mock` here, what do you think about testing for the presence of the `Stripe-Version` header in the mock below? I think that should be enough? Here's an example of someone doing  a check on idempotency key:

``` python
        self.requestor_mock.request.assert_called_with(
            'post',
            '/v1/charges/ch_update_id/dispute/close',
            {},
            {'Idempotency-Key': 'foo'},
        )
```

I think generally it's preferable to keep the interface as small as possible so that these tests are mostly consistent in case we want to refactor all of them at some point in the future.

stripe/test/resources/test_ephemeral_keys.py
```diff
@@ -0,0 +1,57 @@
...
+
+        self.requestor_mock.request.assert_called_with(
+            'post',
+            '/v1/ephemeral_keys',
+            {'customer': 'cus_123'},
+            None
+        )
+
+    def do_version_tests(self):
```
brandur-stripe: Again, kind of pedantic, but do you think we could rename this to `_do_version_tests` and move it down to the bottom of the class? This will keep all the test cases together and better demonstrate that this method is supposed to be an internal helper.

stripe/test/resources/test_ephemeral_keys.py
grey-stripe: So, this didn't work because `Stripe-Version` isn't included in the call to `request` since it's configured on the `APIRequestor` itself.  In fact, this test would be failing if it worked the other way, since it's doing a deep comparison of the objects (rather than just checking the keys specified).

I figured any other solution would need to change more code, so I erred on the side of the smallest change.

brandur-stripe: Hm, I see. Alright, I guess it's not too bad to have both exposed.

PR #330, opened 2017-06-24
Add optional `code` to `InvalidRequestError`.

mastermatt:
A POST to `/v1/charges/ch_xxxx/capture` after the auth has released/expired will respond with a 400 error and body of:
```
{
  "error": {
    "type": "invalid_request_error",
    "message": "Charge ch_xxxx cannot be captured because the charge has expired (a charge must be captured within 7 days). You will need to create a new charge to retry the payment.",
    "code": "charge_expired_for_capture"
  }
}
```

This change exposes the `code` value as an attr on the error instance just as `CardError` does.

### DIFF
--- a/stripe/api_requestor.py
+++ b/stripe/api_requestor.py
@@ -192,7 +192,7 @@ class APIRequestor(object):
         elif rcode in [400, 404]:
             return error.InvalidRequestError(
                 error_data.get('message'), error_data.get('param'),
-                rbody, rcode, resp, rheaders)
+                error_data.get('code'), rbody, rcode, resp, rheaders)
         elif rcode == 401:
             return error.AuthenticationError(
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Thanks!

brandur-stripe:
Released as 1.61.0.

mastermatt:
Is there a way to test this response from the API? 
The list of test numbers here https://stripe.com/docs/testing#cards-responses don't seem to offer this type of error as an option.

remi-stripe:
@mastermatt there isn't no. You need to create an authorization, wait 7 days for it to expire and then try to capture it. No better way at the moment.

mastermatt:
@remi-stripe ok, thanks.

PR #331, opened 2017-06-27
Add parameters when calling `pay` on an invoice

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @ark-stripe 

Users can now provide a `source` parameter when paying invoices, but the `pay` method currently does not let users provide any parameters. This PR fixes that.

Thanks to keyword args, this is not a breaking change.


brandur-stripe:
LGTM.

brandur-stripe:
Released as 1.62.0.

Issue #329, opened 2017-06-23
python 3.6 support


lestatcheb:
stripe client doesn't work with python 3.6 (tried last stripe==1.59.0)

```
  File "/mnt/storage/code/venv/lib/python3.6/site-packages/stripe/resource.py", line 466, in create
    response, api_key = requestor.request('post', url, params, headers)
  File "/mnt/storage/code/venv/lib/python3.6/site-packages/stripe/api_requestor.py", line 150, in request
    method.lower(), url, params, headers)
  File "/mnt/storage/code/venv/lib/python3.6/site-packages/stripe/api_requestor.py", line 339, in request_raw
    util.log_debug('API response body', body=rbody)
  File "/mnt/storage/code/venv/lib/python3.6/site-packages/stripe/util.py", line 86, in log_debug
    msg = logfmt(dict(message=message, **params))
  File "/mnt/storage/code/venv/lib/python3.6/site-packages/stripe/util.py", line 129, in logfmt
    return ' '.join([fmt(key, val) for key, val in sorted(props.items())])
  File "/mnt/storage/code/venv/lib/python3.6/site-packages/stripe/util.py", line 129, in <listcomp>
    return ' '.join([fmt(key, val) for key, val in sorted(props.items())])
  File "/mnt/storage/code/venv/lib/python3.6/site-packages/stripe/util.py", line 122, in fmt
    val = str(val)
BytesWarning: str() on a bytes instance
```

ob-stripe:
@lestatcheb That's weird, the library is tested against 3.6: https://travis-ci.org/stripe/stripe-python/jobs/244747027.

Can you share the code that led to this exception? Make sure to redact your API keys.

lestatcheb:
Hi, sure.

```
> import stripe
> stripe.api_key = 'my_api_key'
> stripe.api_version = '2017-02-14'
> token_create_data = {
>     'number': '4242424242424242',
>     'name': 'TEST TESTOV',
>     'exp_month': '12',
>     'exp_year': '2019',
>     'cvc': '123',
> }
> res = stripe.Token.create(card=token_create_data)


message='Request to Stripe api' method=post path=https://api.stripe.com/v1/tokens
message='Stripe API response' path=https://api.stripe.com/v1/tokens response_code=200
---------------------------------------------------------------------------
BytesWarning                              Traceback (most recent call last)
<ipython-input-1-f8c858aee1e6> in <module>()
     11     'cvc': '123',
     12 }
[...]

PR #332, opened 2017-06-27
Use `-W all` instead of `-W always`

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This was reported by @jleclanche in #329. `always` is not a a valid value for the `-W` flag. The correct value is `all`. From the man page:

> **all** to print a warning
> each time it occurs (this may generate many messages if a  warn-
> ing  is  triggered  repeatedly for the same source line, such as
> inside a loop)


brandur-stripe:
LGTM. Thanks!

PR #333, opened 2017-06-28
Fix example for Python 3 compatibility

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @revmischa

Fix the webhooks example to make it work with Python 3 as well as 2. This was pointed out by @revmischa a couple of weeks back [here](https://github.com/stripe/stripe-python/commit/59078610ff7dfc67c7adb63c8977f6309ae88885#commitcomment-22550934).


brandur-stripe:
Good call. LGTM.

PR #338, opened 2017-08-07
Fix serialization of items when saving subscriptions

### LINKED ISSUE #336: Subscription items are not converted to dict when using setattr + save
When calling `Subscription.create` or `Subscription.modify`, the library converts the `items` parameter from a list of dicts to a numerical-indexed dict of dicts to work around the infamous list of dicts URL-encoding issue.

(Relevant code: https://github.com/stripe/stripe-python/blob/9d149e747758b39e2d3ad939a2cd11a44e3c0267/stripe/resource.py#L868)

However, this doesn't work when retrieving the subscription, setting `items` via `__setattr__` and calling `save`:

```python
sub = stripe.Subscription.retrieve('sub_...')
sub.items = [
  {'id': 'si_...', 'deleted': True},
  {'plan': 'gold'},
]
sub.save()
```

In this case, the `items` parameter is not converted, which causes the API to decode it as `[{'id': 'si_...', 'deleted': True, 'plan': 'gold'}]` due to the lack of a common key between the two dicts. This of course produces incorrect results.

Manually providing a numerical-indexed dict instead of a list does not work either, because in this case the library tries to unset the `object`, `url`,  `data`, `has_more` and `total_count` attributes from the original list object that was set as the value for the `items` attribute when the subscription was retrieved, causing the API to return an "Invalid array" error.

cc @brandur-stripe @kjc-stripe

brandur:
Code had been added to the class methods to properly handle serializing
subscription items to an integer-indexed hash, but the same wasn't
happening for the instance-level `save` method. This patch adds that and
a test to make sure it's happening as expected.

Fixes #336.

r? @ob-stripe
cc @kjc-stripe If you wouldn't mind also taking a look at this ...
cc @stripe/api-libraries

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -874,6 +874,13 @@ class Subscription(CreateableAPIResource, DeletableAPIResource,
             params["items"] = convert_array_to_dict(params["items"])
         return super(Subscription, cls).create(**params)
 
+    def serialize(self, previous):
+        updated_params = super(UpdateableAPIResource, self).serialize(previous)
+        if "items" in updated_params:
+            updated_params["items"] = convert_array_to_dict(
+                updated_params["items"])
+        return updated_params
+
 
 class SubscriptionItem(CreateableAPIResource, DeletableAPIResource,
                        UpdateableAPIResource, ListableAPIResource):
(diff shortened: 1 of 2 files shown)


kjc-stripe:
> serialization
> integer-indexed hash
🙈

kjc-stripe:
👍 

Can we make sure to add a large warning to the changelog about this? It's technically a breaking change, as calls to save will now have different behavior.

brandur-stripe:
Thanks Kyle!

> Can we make sure to add a large warning to the changelog about this? It's technically a breaking change, as calls to save will now have different behavior.

I *think* we're mostly going to be okay on this one though. `items` parameters that worked before should stay working, and `items` parameters that weren't being encoded properly would almost certainly have been erroring from the API anyway. But cool, I'll add a note.

brandur-stripe:
Released as 1.62.1.

PR #339, opened 2017-08-22
Add type-checking for `id` in APIResource.retrieve

seanpianka:
My original, and incorrect, invocation was:

`subscription = stripe.Subscription.retrieve({g.user.stripe_sub_id})`

`stripe.Subscription.retrieve` is defined by `APIResource.retrieve`, and it
expects its first parameter `id` to be of type `str`. When an non-`str` type,
and more importantly an un-`urllib.quote_plus`-able type, is provided, the
following stacktrace results:

```
File "/var/task/delete_users.py", line 19, in delete_users
subscription = stripe.Subscription.retrieve({g.user.stripe_sub_id})
File "/var/task/stripe/resource.py", line 338, in retrieve
instance.refresh()
File "/var/task/stripe/resource.py", line 342, in refresh
self.refresh_from(self.request('get', self.instance_url()))
File "/var/task/stripe/resource.py", line 367, in instance_url
extn = urllib.parse.quote_plus(id)
File "/var/lang/lib/python3.6/urllib/parse.py", line 791, in quote_plus
string = quote(string, safe + space, encoding, errors)
File "/var/lang/lib/python3.6/urllib/parse.py", line 775, in quote
return quote_from_bytes(string, safe)
File "/var/lang/lib/python3.6/urllib/parse.py", line 800, in quote_from_bytes
raise TypeError("quote_from_bytes() expected bytes")
TypeError: quote_from_bytes() expected bytes
```

This is due to `urllib.quote_plus` making use of `urllib.quote`, which performs
type-checking to ensure that the argument provided for the parameter `string`
is of type `str`. `urllib.quote_from_bytes` is then invoked, but that expects
an encoded `str`. This causes an exception to be raised related to incorrect
invocation of `urllib.quote_from_bytes` rather than a semantically-incorrect
invocation of `APIResource.retrieve`.

The following error handling would raise a more useful exception to the developer
when the argument for an `id` to `APIResource.retrieve` is not of type `str`.

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -357,10 +357,13 @@ class APIResource(StripeObject):
 
     def instance_url(self):
         id = self.get('id')
-        if not id:
+
+        if not isinstance(id, basestring):
             raise error.InvalidRequestError(
                 'Could not determine which URL to request: %s instance '
-                'has invalid ID: %r' % (type(self).__name__, id), 'id')
+                'has invalid ID: %r, %s. ID should be of type `str` (or'
+                ' `unicode`)' % (type(self).__name__, id, type(id)), 'id')
+
         id = util.utf8(id)
         base = self.class_url()
         extn = urllib.quote_plus(id)
--- a/stripe/test/resources/test_api_resource.py
+++ b/stripe/test/resources/test_api_resource.py
@@ -77,6 +77,11 @@ class APIResourceTests(StripeApiTestCase):
         self.assertEqual({"f": "b"},
                          stripe.resource.convert_array_to_dict({"f": "b"}))
 
+    def test_raise_on_incorrect_id_type(self):
+        for obj in [None, 1, 3.14, dict(), list(), set(), tuple(), object()]:
+            self.assertRaises(stripe.error.InvalidRequestError,
+                              MyResource.retrieve, obj)
+
 
 class SingletonAPIResourceTests(StripeApiTestCase):
 


mastermatt:
I'm not on the Stripe team, but I'm in favor of this. I've found myself passing ints before and being confused by the error. I would recommend adding a test to test/resources/test_api_resource.py and squashing your commits.

brandur-stripe:
Hi Sean, yeah I like it — I'm always in favor of better error messages.

A couple (hopefully small) asks:

1. Would you mind adding a test like @mastermatt (thanks BTW!) suggested? Just way easier to make sure that this behavior will keep working if we have one.
2. Can we add something to the error message that mentions something to the effect that IDs are expected to be strings? I think that will make it even easier for the user to figure out what they're supposed to do.

And also yes, a squash would be great.

Thanks for the patch!

stripe/resource.py
```diff
@@ -357,10 +357,13 @@ def class_url(cls):
 
     def instance_url(self):
         id = self.get('id')
-        if not id:
+
+        if not isinstance(id, (str, unicode)):
```
ob-stripe: `isinstance(id, basestring)` would be slightly more Pythonic :)

seanpianka: See [#7ae76e8](https://github.com/stripe/stripe-python/pull/339/commits/7ae76e8e4fe549b2e0f0c8a7bd89ad7f83678362) as for why I chose `(str, unicode)`.

In summary, it was a Python 2 compatibility issue concerning the following behavior:

```
Python 2.7.13 (default, Jul 21 2017, 03:24:34) 
[GCC 7.1.1 20170630] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> str == unicode
False
```

So `if not isinstance(id, str)` was failing with Python 2.

ob-stripe: [`basestring`](https://docs.python.org/2/library/functions.html#basestring) is the abstract parent class for both `str` and `unicode` in Python 2, and is translated to `str` by `2to3`.

seanpianka: Okay, I will use `basestring` as I didn't completely realize stripe-python was using `2to3`. Thank you for the suggestion!

seanpianka:
I have added a testcase, updated the error message to mention that an ID's expected type is of `basestring`, and squashed my commits.

brandur-stripe:
Great. Thanks Sean!

brandur-stripe:
Released as 1.64.0.

Issue #273, opened 2016-12-21
stripe.api_requestor._api_encode ambiguously encodes list of dicts


nickbaum:
I'm filing this bug to follow up on a support conversation with Thomas and Armand, who were able to reproduce this issue.

I dug into how the request arguments are converted, and it seems stripe.api_requestor._api_encode encodes the items in an ambiguous way:

```
>>> items = [{'quantity': 1, 'parent': 'black', 'type': 'sku'}, {'description': 'Gift credit', 'type': 'discount', 'amount': -3900, 'currency': 'usd'}] 
>>> list(stripe.api_requestor._api_encode({"items": items}))
[('items[][quantity]', 1), ('items[][type]', 'sku'), ('items[][parent]', 'black'), ('items[][type]', 'discount'), ('items[][description]', 'Gift credit'), ('items[][currency]', 'usd'), ('items[][amount]', -3900)]
```

This can result in the Python library submitting incorrect data to the API.

From Armand:

> Armand here, jumping in for Thomas, I hope you don't mind! I've spent some time investigating the issue and I believe you're correct: lists of hashes are encoded in an ambiguous way. This, coupled with the fact that Python hashes (dicts) are unordered, can result in inconsistent behavior.
> 
> One possible workaround would be to use OrderedMap and make sure to insert the `type` attribute first in each item's dict. On the API's end, when a key appears twice in the same list, it will be interpreted as a new element in the list.
> 
> Using OrderedMaps is a bit unwieldy though, so a simpler workaround is to replace the list by a hash with incrementing numerical keys, e.g.:
> 
> ```
> >>> items = {0: {'quantity': 1, 'parent': 'black', 'type': 'sku'}, 1: {'description': 'Gift credit', 'type': 'discount', 'amount': -3900, 'currency': 'usd'}}
> >>> list(stripe.api_requestor._api_encode({"items": items}))
> [('items[1][amount]', -3900), ('items[1][description]', 'Gift credit'), ('items[1][type]', 'discount'), ('items[1][currency]', 'usd'), ('items[0][quantity]', 1), ('items[0][parent]', 'black'), ('items[0][type]', 'sku')]
> ```
> 
> This removes the ambiguity and will be decoded correctly on the API's end.
> 
> In any case, this is probably something that should be handled in the bindings themselves. I will forward the details to the bindings maintainers and discuss the best way to fix this issue.

brandur:
Thank you for logging this! We're looking into the best way to get this fixed.

[...]

PR #342, opened 2017-09-01
Fix encoding for order items

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fixes encoding of order items.

This issue is related to #273 and #336. The `items` attribute in [order creation requests](https://stripe.com/docs/api#create_order) and [order return requests](https://stripe.com/docs/api#return_order) is a list of dicts. Since Python dicts are not ordered, this can produce inconsistent results.

E.g. with the following request:

```python
order = stripe.Order.create(items=[
  {'type': 'sku', 'parent': 'sku_...'},
  {'type': 'discount', 'amount': -500, 'currency': 'usd'},
], ...)
```

the `items` attribute may be encoded as:

```
items[][type]=sku&items[][parent]=sku_...&items[][type]=discount&items[][amount]=-500&items[][currency]=usd
```

which would be correctly decoded on the API's end (because the common `type` key is used to detect the multiple list elements) and produce the expected result. But because the dicts are not ordered, the `items` attribute could (among other possibilities) also be encoded as:

```
items[][type]=sku&items[][parent]=sku_...&items[][amount]=-500&items[][currency]=usd&items[][type]=discount
```

which would cause the list to be incorrectly decoded and the request to fail.

The API accepts an integer-indexed hash of hashes in place of a list of hashes for this parameter, so we can use the same fix as we did for subscription items, i.e. transparently convert the parameter into an integer-indexed hash of hashes if it's a list of hashes.

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -1086,12 +1086,20 @@ class SKU(CreateableAPIResource, UpdateableAPIResource,
 
 class Order(CreateableAPIResource, UpdateableAPIResource,
             ListableAPIResource):
+    @classmethod
+    def create(cls, **params):
+        if "items" in params:
+            params["items"] = convert_array_to_dict(params["items"])
+        return super(Order, cls).create(**params)
+
     def pay(self, idempotency_key=None, **params):
         headers = populate_headers(idempotency_key)
         return self.request(
             'post', self.instance_url() + '/pay', params, headers)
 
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Thanks OB! Yep, this was definitely needed.

brandur-stripe:
Released as 1.63.1.

PR #344, opened 2017-09-06
Add support for stripe_version argument in requests

### LINKED ISSUE #343: Stripe api_version not customizable for stripe.Customer.list
When calling `stripe.Customer.list(api_key=key, api_version='2015-10-16')` with a custom api_version, we get the following error:

```
InvalidRequestError: Request req_I25zAXmrYWc3Mj: Received unknown parameter: api_version
```

This is using python 2.7 via the CLI.

It appears that the `api_version` param is not correctly passed into the `APIRequestor` constructor at https://github.com/stripe/stripe-python/blob/master/stripe/resource.py#L451 . I think it should look like https://github.com/stripe/stripe-python/blob/master/stripe/resource.py#L924 . There might be other places where this happens as well.

We are able to work around this by just setting the global stripe api version using `stripe.api_version='2015-10-16'`.

Thank you!

ob-stripe:
r? @brandur-stripe 

Fixes #343. Basically, it turns out we never supported specifying a custom API version on a per-request basis in the Python library (except for ephemeral keys, which _require_ an explicit API version).

This PR adds supports for a `stripe_version` named parameter in all request methods. The parameter is handled like `stripe_account`, i.e. it is persistently maintained on instances.

E.g. when doing the following:
```python
c = stripe.Customer.create(stripe_version='2015-02-18')
c.description = "foo"
c.save()
```

the update request sent by `save()` will also include the `Stripe-Version: 2015-02-18` header. This behavior is consistent with the Ruby and PHP libraries.

In order to be consistent with the other libraries, the parameter is named `stripe_version`, not `api_version` (cf. [Ruby](https://github.com/stripe/stripe-ruby/blob/855ce0c7d45e7a096ae404c1790c8c9223ffefd9/lib/stripe/api_operations/request.rb#L38) and [PHP](https://github.com/stripe/stripe-php/blob/c8c8d384636b6b78c4c17e2b00102e914f0b7cc5/lib/Util/RequestOptions.php#L67-L69)).

Potential breaking changes in this PR:

- I renamed the existing `api_version` parameter in `EphemeralKey` methods to `stripe_version`. This obviously isn't great. Maybe we could alias `api_version` to `stripe_version` and issue a deprecation warning so that users update their code to use the latter?

- The `convert_to_stripe_object` method now has 4 arguments instead of 3. While the method is [publicly exposed](https://github.com/stripe/stripe-python/blob/ef50aa5e96b05a47f4b277ac6eab76e336860b99/stripe/__init__.py#L103), I think it's fair to say that this is an internal method that shouldn't be called directly by users so modifying the signature isn't a breaking change, but that's definitely debatable.

I don't think this PR should be merged until the above two points have been addressed. Let me know what you think!

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -6,7 +6,7 @@ from copy import deepcopy
 from stripe import api_requestor, error, oauth, util, upload_api_base
 
 
-def convert_to_stripe_object(resp, api_key, account):
+def convert_to_stripe_object(resp, api_key, version, account):
     types = {
         'account': Account,
         'alipay_account': AlipayAccount,
@@ -48,7 +48,8 @@ def convert_to_stripe_object(resp, api_key, account):
     }
 
     if isinstance(resp, list):
-        return [convert_to_stripe_object(i, api_key, account) for i in resp]
+        return [convert_to_stripe_object(i, api_key, version, account)
+                for i in resp]
     elif isinstance(resp, dict) and not isinstance(resp, StripeObject):
         resp = resp.copy()
         klass_name = resp.get('object')
@@ -56,7 +57,8 @@ def convert_to_stripe_object(resp, api_key, account):
             klass = types.get(klass_name, StripeObject)
         else:
             klass = StripeObject
-        return klass.construct_from(resp, api_key, stripe_account=account)
+        return klass.construct_from(resp, api_key, stripe_version=version,
+                                    stripe_account=account)
     else:
         return resp
 
@@ -103,7 +105,8 @@ def _serialize_list(array, previous):
 
 
 class StripeObject(dict):
-    def __init__(self, id=None, api_key=None, stripe_account=None, **params):
+    def __init__(self, id=None, api_key=None, stripe_version=None,
+                 stripe_account=None, **params):
         super(StripeObject, self).__init__()
 
         self._unsaved_values = set()
@@ -113,6 +116,7 @@ class StripeObject(dict):
         self._previous = None
 
         object.__setattr__(self, 'api_key', api_key)
+        object.__setattr__(self, 'stripe_version', stripe_version)
         object.__setattr__(self, 'stripe_account', stripe_account)
 
         if id:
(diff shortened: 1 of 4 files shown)


brandur-stripe:
Hey @ob-stripe, thanks for doing the work and the great writeup!

This looks good to me.

> I renamed the existing api_version parameter in EphemeralKey methods to stripe_version. This obviously isn't great. Maybe we could alias api_version to stripe_version and issue a deprecation warning so that users update their code to use the latter?

Hm, actually would you mind if we did the aliasing instead? I think renaming the parameter is obviously the right move, but we're on a great 1.* run with this library and it would be a shame to break it for something like ephemeral keys that likely few people are using yet anyway :)

> The convert_to_stripe_object method now has 4 arguments instead of 3. While the method is publicly exposed, I think it's fair to say that this is an internal method that shouldn't be called directly by users so modifying the signature isn't a breaking change, but that's definitely debatable.

Yeah, agreed that we should consider this part of our internal API.

Given that we're about to change it anyway though — what do you think about tweaking the signature so that just `response` is a required argument? `api_key`, `stripe_account` and `stripe_version` becomes optional arguments in a similar way that they are for other methods.

ob-stripe:
> Hm, actually would you mind if we did the aliasing instead? I think renaming the parameter is obviously the right move, but we're on a great 1.* run with this library and it would be a shame to break it for something like ephemeral keys that likely few people are using yet anyway :)

Oh, definitely! I just wanted to hear your feedback before settling on the aliasing approach. Let's not introduce avoidable breaking changes :)

I pushed a commit to use `api_version` if `stripe_version` is `None`, and output a `DeprecationWarning` so existing integrations can start updating their code to use `stripe_version` instead.

> Given that we're about to change it anyway though — what do you think about tweaking the signature so that just response is a required argument? api_key, stripe_account and stripe_version becomes optional arguments in a similar way that they are for other methods.

Good idea! Done, all 3 parameters are now optional.

ptal @brandur-stripe

brandur-stripe:
> I pushed a commit to use api_version if stripe_version is None, and output a DeprecationWarning so existing integrations can start updating their code to use stripe_version instead.

Nice! With a test and everything. LGTM.

PR #345, opened 2017-09-21
Handle case where payload is a bytearray or bytes

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @spastorelli-stripe

This PR fixes the `Webhook.construct_event` method to handle the case where `payload` is a `bytearray` or `bytes`.

In particular, with Python 3, retrieving the request's body in Django via `request.body` will return a `bytes` instance (where as with Python 2, a `str` instance is returned). This fix makes it so users don't have to worry about this and can always pass `request.body` to the method. (It also means that our [sample code](https://stripe.com/docs/webhooks#verify-official-libraries) will work regardless of which Python version is used.)

### DIFF
--- a/stripe/test/test_webhook.py
+++ b/stripe/test/test_webhook.py
@@ -1,3 +1,4 @@
+import sys
 import time
 
 import stripe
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Thanks for getting those tests!

LGTM.

brandur-stripe:
Released as 1.65.1.

PR #348, opened 2017-09-29
ListObject: proxy __len__ to data list

blueyed:
Otherwise it will use the one from the underlying dict it seems, and
empty ListObjects (without data) are trueish then.

### DIFF
--- a/stripe/resource.py
+++ b/stripe/resource.py
@@ -427,6 +427,9 @@ class ListObject(StripeObject):
     def __iter__(self):
         return getattr(self, 'data', []).__iter__()
 
+    def __len__(self):
+        return getattr(self, 'data', []).__len__()
+
 
 class SingletonAPIResource(APIResource):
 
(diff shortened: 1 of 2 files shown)


ob-stripe:
Hey @blueyed, thanks for the PR!

There was some debate a while back on how "array-like" we want to our list objects to be. In this case, since `ListObjects` already have iterator semantics, this PR seems safe to me.

r? @brandur-stripe for a second set of eyes

blueyed:
@ob-stripe 
Thanks.

As for us, we fell into the `if not customer.sources.list():` trap.

brandur-stripe:
Yeah, looks fine. Thanks! (And thanks for writing tests.)

I guess the one possible gotcha is that it might be possible to conflate a length measurement as the actual number of objects in the collection rather than the number of objects on the current page. As OB mentioned though, we seem roughly committed to iterator-like behavior over the current page anyway, so it's okay.

brandur-stripe:
Released as 1.66.0.

blueyed:
Thank you!

PR #356, opened 2017-10-23
Allow uploading in-memory files

### LINKED ISSUE #227: Better support for uploading in-memory images
A fairly common use case for file uploads is to upload an image to Stripe that was just uploaded to the server, or to retrieve an image from a URL, then upload it to Stripe. As it stands, the file uploader seems to expect to be given a file stored on disk, which isn't really necessary. More specifically, the file uploader wants something that has a `.read` method, and a `.name` method.

Right now, if I take a `StringIO` (which supports `.read`), then monkey-patch it to have `.name="foo"`, this seems to upload fine, and I more or less get the thing I want. This is a bit gross, though. Ideally, we wouldn't require a file name for the file at all, since this doesn't appear to be getting used in any meaningful way. (The upload response or anything else I can find doesn't seem to make reference to the filename.)

Though it doesn't seem like uploads.stripe.com actually does anything with the file name, it does get unhappy if the parameter is omitted or left blank. As far as I can tell, if I patch stripe-python to just pass a hard-coded dummy value as the file name, this works fine. This is clearly a hack, though. In an ideal world, uploadsrv wouldn't look for a `filename` at all, though I don't really know if that would be a simple change or not.

Thoughts on whether this is something we'd want to support, and if so, what the cleanest way of doing this is?

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @praboud-stripe

Fixes #227.

I spent some time investigating this and I agree with @praboud-stripe's analysis in #227. Rack returns multipart blocks with no filenames directly as `String`s rather than as `Hash`es -- they explicitly test for this behavior (cf. [here](https://github.com/rack/rack/blob/d4793a0707efa9011ded6ebc88154ac9d7f98c8a/test/spec_multipart.rb#L36)).

Fixing the backend to handle either a string or a hash might be possible, but sending a dummy filename in the client libraries is definitely the path of least resistance. We actually already do this in stripe-node (cf. [here](https://github.com/stripe/stripe-node/blob/a7382226965ee399ce6b5b9fd162c5af74c8131d/lib/MultipartDataGenerator.js#L30)).

### DIFF
--- a/stripe/api_resources/file_upload.py
+++ b/stripe/api_resources/file_upload.py
@@ -1,4 +1,5 @@
-from stripe import api_requestor, upload_api_base, util
+import stripe
+from stripe import api_requestor, util
 from stripe.api_resources.abstract import ListableAPIResource
 
 
@@ -7,7 +8,7 @@ class FileUpload(ListableAPIResource):
 
     @classmethod
     def api_base(cls):
-        return upload_api_base
+        return stripe.upload_api_base
 
     @classmethod
     def class_name(cls):
--- a/stripe/multipart_data_generator.py
+++ b/stripe/multipart_data_generator.py
@@ -18,10 +18,12 @@ class MultipartDataGenerator(object):
             self._write(self.param_header())
             self._write(self.line_break)
             if hasattr(value, 'read'):
+                filename = value.name if hasattr(value, 'name') else "blob"
+
                 self._write("Content-Disposition: form-data; name=\"")
                 self._write(key)
                 self._write("\"; filename=\"")
-                self._write(value.name)
+                self._write(filename)
                 self._write("\"")
                 self._write(self.line_break)
                 self._write("Content-Type: application/octet-stream")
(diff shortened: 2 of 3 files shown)


stripe/api_resources/file_upload.py
```diff
@@ -7,7 +8,7 @@ class FileUpload(ListableAPIResource):
 
     @classmethod
     def api_base(cls):
-        return upload_api_base
+        return stripe.upload_api_base
```
ob-stripe: Opportunistic fix to ensure that the current value of `upload_api_base` is used (before, the value was only read once when the library is loaded).

brandur-stripe: Nice find.

brandur-stripe:
Oh man, the conversation we had about this in #227 was like 10x longer than the fix turned out to be! Thanks for being the one to take action here Olivier :) And thanks for adding the new test too — really helpful in helping to understand that this does the right thing.

LGTM.

brandur-stripe:
Released as 1.70.0.

praboud-stripe:
Thanks @ob-stripe !

PR #357, opened 2017-10-23
Catch `Exception` instead of using bare `except`

brandur:
This seems to be new, but flake8 is complaining that we're using a bare
except and it's breaking the build:

    flake8 stripe
    stripe/webhook.py:48:9: E722 do not use bare except'

Here we change over to catch the parent `Exception` class instead.

r? @ob-stripe


ob-stripe:
Wonder how this passed before, but lgtm!

brandur-stripe:
> Wonder how this passed before, but lgtm!

Yeah seriously! All I can think of is that the default `flake8` behavior changed.

Thanks!

PR #361, opened 2017-10-30
Add support for exchange_rates requests

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries @alixander-stripe

Adds support for the new `/v1/exchange_rates` endpoints.


brandur-stripe:
I was about to ask, but looks like this is updated already! You're too fast OB!

LGTM.

brandur-stripe:
Released as 1.72.0.

PR #350, opened 2017-10-16
Nested resource class methods

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Basically a port of https://github.com/stripe/stripe-ruby/pull/597 for the Python library.

A couple differences:

- implemented as a decorator, because Python

- the update methods are called `modify_<nested_resource>` instead of `update_<nested_resource>` to match the pre-existing naming convention in the library.

Everything else should be the same.

PR #362, opened 2017-10-30
Minor fixes

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

A few random fixes:

- removed the existing `modify_external_account` and `modify_source` class methods on `Account` and `Customer` resp. since those methods are now created by the `@nested_resource_class_methods` decorator (#350)

- fixed the object name for `InvoiceItem` and added `BalanceTransaction` in `util.convert_to_stripe_object`. I think I broke both of these in #351 :/

The existing `@nested_resource_class_methods` tests ensure the removed methods still actually exist. Added a very basic test to avoid regression for the other fixes.

### DIFF
--- a/stripe/api_resources/invoice_item.py
+++ b/stripe/api_resources/invoice_item.py
@@ -6,4 +6,4 @@ from stripe.api_resources.abstract import ListableAPIResource
 
 class InvoiceItem(CreateableAPIResource, UpdateableAPIResource,
                   ListableAPIResource, DeletableAPIResource):
-    OBJECT_NAME = 'invoice_item'
+    OBJECT_NAME = 'invoiceitem'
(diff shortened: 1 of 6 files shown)


stripe/api_resources/invoice_item.py
```diff
@@ -6,4 +6,4 @@
 
 class InvoiceItem(CreateableAPIResource, UpdateableAPIResource,
                   ListableAPIResource, DeletableAPIResource):
-    OBJECT_NAME = 'invoice_item'
+    OBJECT_NAME = 'invoiceitem'
```
brandur-stripe: We should probably cut a new version for this or something to prevent many future instances of confusion around this bad naming ...

brandur-stripe:
Thanks for following up with these OB! Am I good to release these as a patch version? (i.e. the interface of `modify_external_account` and `modify_source` is identical to previous usage)

ob-stripe:
Yep, the interface is identical.

brandur-stripe:
Released as 1.71.2.

PR #366, opened 2017-11-02
Move tests out of stripe directory

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

With `2to3` out of the way, we can finally move tests in their own top-level dedicated directory! This means that test code will no longer be installed as part of the `stripe` module.

Lots of files impacted, but the PR is pretty straightforward:

- `stripe/test` --> `tests`

- updated `setup.py` accordingly

- updated every test file to replace `from stripe.test` with `from tests`

- removed some unneeded / outdated stuff such as the contents of `tests/__init__.py` and the calls to `unittest2.main()` in some of the older test files (convenient, but not very clean -- tests should always be executed through `tox` or `python setup.py test` to ensure that the environment is correctly set up)


setup.py
```diff
@@ -7,10 +7,7 @@
 except ImportError:
     from distutils.core import setup
 
-try:
-    from distutils.command.build_py import build_py_2to3 as build_py
-except ImportError:
-    from distutils.command.build_py import build_py
+from distutils.command.build_py import build_py
```
ob-stripe: This really should have been part of #364 :shame:

brandur-stripe:
These improvements are coming so quick now! Can't wait to see what's next :)

Tests still look like they're running pretty happily in CI. Nice. LGTM.

PR #367, opened 2017-11-03
Use stripe-mock for API resources tests

ob-stripe:
r? @brandur-stripe 

Opening an early PR to get feedback on the approach.

I tried to stick close to the Ruby implementation of stripe-mock tests. One of the difficulties is that Python has no webmock equivalent (or at least I couldn't find anything that really suited our needs).

I implemented a (very) poor man's webmock by using the mock library to patch `APIRequestor.request`. If a stubbed request was registered for the method and URL, then the stubbed response is returned, otherwise the original `APIRequestor.request` is called and the request goes through to stripe-mock. We can use the patch object to assert that `request` was called with the expected parameters. I've also added an `assert_requested` convenience method that lets us assert on some of the method arguments only.

The implementation is still a little rough around the edges but it does work. I've updated two tests to show how it's used:
- `test_coupon.py` is a "regular" test where the requests go through to stripe-mock
- `test_source_transaction.py` stubs the request as the resource is not yet supported by stripe-mock


tests/helper.py
```diff
@@ -4,12 +4,105 @@
...
+    def _patched_request(self, requestor, method, url, *args, **kwargs):
+        if (method, url) in self._stubbed_requests:
+            response_body = self._stubbed_requests.pop((method, url))
+            return response_body, stripe.api_key
+        return self._real_request(requestor, method, url, *args, **kwargs)
+
+    def stub_request(self, method, url, response_body={}):
+        self._stubbed_requests[(method, url)] = response_body
+
+    def assert_requested(self, method, url, params=ANY, headers=ANY):
```
brandur-stripe: OOC, do you know what's up with Python's mixing of snakecase and camelcase? It's weird to see `assertEquals` mixed with `assert_requested` everywhere. I see that PEP8 recommends snakecasing, so I guess the unittest API is mostly an accident of history that's been left in place for compatibility?

brandur-stripe:
It's happening!!! Looking great to me so far.

> I tried to stick close to the Ruby implementation of stripe-mock tests. One of the difficulties is that Python has no webmock equivalent (or at least I couldn't find anything that really suited our needs).
>
> I implemented a (very) poor man's webmock by using the mock library to patch APIRequestor.request. If a stubbed request was registered for the method and URL, then the stubbed response is returned, otherwise the original APIRequestor.request is called and the request goes through to stripe-mock. We can use the patch object to assert that request was called with the expected parameters. I've also added an assert_requested convenience method that lets us assert on some of the method arguments only.

Yeah, I'm not sure I could give you a strong opinion on what the right course of action for Python is, but your implementations seems to be pretty solid.

As one more data point, did you look at Go's stripe-mock implementation at all? Like in Python it seems, mocking HTTP in Go is just not really a thing so it performs real HTTP requests against the local HTTP daemon. I guess the major disadvantage to this is that you lose your ability to put something like `assertRequested` in play, which admittedly is a fairly sizable downside.

tests/helper.py
```diff
@@ -4,12 +4,105 @@
...
+             "it running? Please see README for setup instructions." %
+             MOCK_PORT)
+
+version = info.get('Stripe-Mock-Version')
+if version != 'master' \
+        and StrictVersion(version) < StrictVersion(MOCK_MINIMUM_VERSION):
+    sys.exit("Your version of stripe-mock (%s) is too old. The minimum "
+             "version to run this test suite is %s. Please "
+             "see its repository for upgrade instructions." %
+             (version, MOCK_MINIMUM_VERSION))
```
brandur-stripe: Thanks for copying these various checks over! I think they're fairly ergonomically advantageous for people trying to make use of stripe-mock, so it's important to port them everywhere that we port stripe-mock.

tests/helper.py
ob-stripe: Yeah, AFAIK it's because many parts of the standard library predate PEP8 and it was decided that maintaining backwards compatibility was more important than PEP8 compliance.

I generally try to ignore this and always use snake case for my own methods, but I can rename the method to `assertRequested` if you'd rather!

brandur-stripe: Nah, I think what you're doing is right — just too bad this crazy historical inconsistency exists :)

remi-stripe:
@brandur-stripe Even though it's a WIP we think it's ready to be reviewed. We'd love for you to catch anything missing before we do all the other tests and merge next week.

A lot of tests were removed because they were broken in the first place (not testing the right thing, working by chance, etc.) though we also added some tests that were not there before.

tests/api_resources/test_alipay_account.py
```diff
@@ -0,0 +1,60 @@
...
+            'post',
+            '/v1/customers/cus_123/sources/%s' % TEST_RESOURCE_ID
+        )
+
+    def test_is_not_retrievable(self):
+        with self.assertRaises(NotImplementedError):
+            stripe.AlipayAccount.retrieve(TEST_RESOURCE_ID)
+
+    # Below, we don't use stripe-mock as it always returns a Bank Account
+    # object when you hit /v1/customers/%s/sources/%s
```
brandur-stripe: Good call in putting this in a separate section.

brandur-stripe:
@remi-stripe @ob-stripe I didn't go over every line for this one, but reading through the patch my pretty strong feeling is that you'e on the right track here. This is almost exactly what I did for stripe-ruby as a I was doing the conversion: I tried to move every resource test suite to a relatively standardized set of normal tests that look as close as possible to every other file. Where there were special considerations, I added a few additional ones that were less standard.

Sorry that I will probably miss this touching down, but +1.

tests/api_resources/test_subscription.py
```diff
@@ -1,162 +1,137 @@
...
-
-        subscription.delete_discount()
-
-        self.requestor_mock.request.assert_called_with(
-            'delete',
-            '/v1/subscriptions/test_sub/discount'
         )
+        self.assertIsInstance(resource, stripe.Subscription)
+
+    # TODO: Fix this test
```
remi-stripe: @ob-stripe Flagging this one because I don't understand it. The error is 
```
======================================================================
FAIL: test_is_saveable_with_items (tests.api_resources.test_subscription.SubscriptionTest)
----------------------------------------------------------------------
  File "/Users/remi/Workspace/bindings/stripe-python/tests/api_resources/test_subscription.py", line 129, in test_is_saveable_with_items
    "id": "si"
  File "/Users/remi/Workspace/bindings/stripe-python/tests/helper.py", line 92, in assert_requested
    return self.request_mock.assert_requested(*args, **kwargs)
  File "/Users/remi/Workspace/bindings/stripe-python/tests/request_mock.py", line 55, in assert_requested
    raise exception
AssertionError: Expected call: request(<ANY>, 'post', '/v1/subscriptions/sub_123', {'items': {'0': {'plan': 'foo', 'id': 'si'}}}, <ANY>)
Actual call: request(<stripe.api_requestor.APIRequestor object at 0x10dca58d0>, 'post', '/v1/subscriptions/sub_BQp6MLuprKbRgM', {u'items': {'0': {'id': 'si', 'plan': 'foo'}}, u'metadata': {}}, None)
```

There seems to be:
* an encoding issue?
* metadata being there when I'm not sure it should?

tests/api_resources/test_subscription.py
```diff
@@ -1,162 +1,137 @@
...
+    #                    "plan": "foo",
+    #                    "id": "si"
+    #                },
+    #            },
+    #        },
+    #    )
+    #    self.assertIsInstance(resource, stripe.Subscription)
+
+
+    # TODO: Test serialize
```
remi-stripe: @ob-stripe I was not sure how to test serialize so leaving this one to you

tests/api_resources/test_subscription_item.py
```diff
@@ -1,90 +1,73 @@
...
         )
-
-    def test_modify_subscription(self):
-        stripe.SubscriptionItem.modify('test_sub_item',
-                                       plan=DUMMY_PLAN['id'])
-
-        self.requestor_mock.request.assert_called_with(
+        self.assertIsInstance(resource, stripe.SubscriptionItem)
+
+    # TODO: Fix this test
```
remi-stripe: @ob-stripe same as subscription, weird metadata/encoding error on this one.

tests/api_resources/test_account.py
```diff
@@ -151,101 +71,116 @@ def test_verify_additional_owner(self):
...
             'post',
-            '/v1/accounts/acct_123/external_accounts',
-            {'source': 'btok_123'},
-            None
+            '/v1/accounts/%s/reject' % account.id
         )
+        self.assertIsInstance(resource, stripe.Account)
+        self.assertTrue(resource is account)
+
+    # TODO: Test deauthorize
```
remi-stripe: @ob-stripe Flagging that we need to test this. Did not fully grasp how to change the base on the stub and assert_requested

tests/api_resources/test_account.py
```diff
@@ -1,148 +1,68 @@
...
+        resource = account.save()
+        self.assert_requested(
             'post',
-            '/v1/accounts/acct_reject/reject',
-            {},
-            None
+            '/v1/accounts/%s' % resource.id
         )
+        self.assertIsInstance(resource, stripe.Account)
+        self.assertTrue(resource is account)
```
remi-stripe: @ob-stripe we should make sure that all calls to save() (on all resources) test this explicitly.

tests/test_integration.py
```diff
@@ -39,59 +39,6 @@ def tearDown(self):
...
-    def test_dns_failure(self):
-        api_base = stripe.api_base
-        try:
-            stripe.api_base = 'https://my-invalid-domain.ireallywontresolve/v1'
-            self.assertRaises(stripe.error.APIConnectionError,
-                              stripe.Customer.create)
-        finally:
-            stripe.api_base = api_base
-
-    def test_run(self):
```
remi-stripe: @ob-stripe I removed this one because it did not seem to do anything that was not done in test_charge.py. Let me know if that was wrong.

tests/test_integration.py
```diff
@@ -156,8 +103,67 @@ def test_declined_card_props(self):
             self.assertTrue(e.request_id.startswith('req_'))
 
 
-class InvalidRequestErrorTest(StripeTestCase):
+# TODO: find a way to move all the ones below those to stripe-mock or to stub
```
remi-stripe: @ob-stripe Flagging this and all the ones below to avoid missing them. stripe-mock seems to be broken (unless I missed something which is likely).

Basically it returns a string for the error that is not a JSON payload and does not set proper headers such as req_XXXX. I filed https://github.com/stripe/stripe-mock/issues/31 though I spent around an hour trying to fix it to no avail and cut my losses.

tests/api_resources/test_subscription.py
ob-stripe: > an encoding issue?

Do you mean the `items` key string being converted to Unicode? That shouldn't be an issue as it doesn't affect string comparison:
```python
>>> 'foo' == u'foo'
True
```

> metadata being there when I'm not sure it should?

It's expected that `metadata` is present here. It won't be sent in the actual request (the value is an empty dict, so it will just be ignored when the parameters are encoded).

Ideally, `metatadata` should _not_ be in the parameters (even though it has no impact on functionality, it's annoying for our tests). #372 seems to take care of that though! I still need to take a closer look but it looks promising.

tests/api_resources/test_subscription.py
ob-stripe: Ack.

tests/api_resources/test_account.py
ob-stripe: Both `stub_request` and `assert_requested` will work regardless of the base URL.

This is because the base URL is set when constructing an `APIRequestor` instance, but our mocks will catch all calls to the `request` method on any `APIRequestor` instance.

I've pushed an update with a test for the `deauthorize` method.

tests/test_integration.py
ob-stripe: Nice catch! Once your fix in https://github.com/stripe/stripe-mock/pull/32 lands, we'll just need to update `.travis.yml` to pull the updated stripe-mock, and update the `MOCK_MINIMUM_VERSION` in `helper.py` accordingly.

tests/test_integration.py
ob-stripe: No, that seems fine. A lot of the tests in this file were no longer useful.

tests/api_resources/test_account.py
ob-stripe: Yep. More generally, we should have a similar assert for all instance methods that `return self`.

tests/api_resources/test_subscription.py
ob-stripe: I've added two tests for `serialize` and what do you know, one of them revealed a bug! Yay tests.

The bug is identical to one I fixed in stripe-ruby a few weeks back, so hopefully it shouldn't be too hard to fix stripe-python as well. I'll tackle this tomorrow.

ob-stripe: I've pushed a fix in #373 and added tests there, and removed the tests from this branch for now. Once #373 is merged we can rebase and bring in the tests as well (they don't actually use stripe-mock or anything since they're just testing the `serialize` method).

tests/api_resources/test_file_upload.py
```diff
@@ -3,40 +3,65 @@
...
+
+class FileUploadTest(StripeMockTestCase):
+    FIXTURE = {
+        'id': TEST_RESOURCE_ID,
+        'object': 'file_upload'
+    }
+
+    def test_is_listable(self):
+        self.stub_request(
+            'get',
```
remi-stripe: is it safe to stub something like this? Feels like we're not testing the feature properly anymore

ob-stripe: Listing and retrieving file_upload objects is not different from retrieving any other API resources, save for the different hostname. I can add an assertion for the hostname if you think that's safer.

Creating a file_upload object is a bit different because of the `multipart` encoding, but most of the logic is tested in `test_multipart_data_generator.py`.

brandur-stripe:
How's this one coming @ob-stripe?

ob-stripe:
Okay, I think this is ready to be pulled in. There's still a lot of room for improvements in all the non-API resources tests, but this PR is already pretty beefy :)

r? @brandur-stripe

tox.ini
```diff
@@ -28,7 +28,7 @@ deps =
     mock
     coverage
 commands =
-    flake8 stripe
+    flake8 stripe tests
```
brandur-stripe: Oops, nice find :)

brandur-stripe:
LGTM @ob-stripe. Nice! Between this pull and breaking up the tests, it's amazing how many lines of code have change in this repository over the last few months :)

You have a couple outstanding conflicts (that might be my fault for merging other PRs before this one). After you have those resolved, I think this is good to come in.

ob-stripe:
Alright, I squashed commits and rebased the PR. Also updated the README file to add instructions for stripe-mock.

brandur-stripe:
Thanks again for all the hard work on this — it's a pretty epic accomplishment.

PR #368, opened 2017-11-07
Disputes are not createable

ob-stripe:
r? @brandur-stripe 

This was noticed by @remi-stripe while working on tests. The `Dispute` resource was declared as createable, however that is not the case. The test was passing because the request was not being sent to the API.

### DIFF
--- a/stripe/api_resources/dispute.py
+++ b/stripe/api_resources/dispute.py
@@ -1,13 +1,11 @@
 from __future__ import absolute_import, division, print_function
 
 from stripe import util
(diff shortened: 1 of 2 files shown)


remi-stripe:
~@ob-stripe why not ignore since we're going to merge stripe-mock in a few days?~
ignore me this is about the resource itself

brandur-stripe:
Nice catch.

Hm, technically this is breaking, but I think I'm going to ship this as a minor release anyway given that this API was never really usable ...

brandur-stripe:
Released as 1.74.0.

PR #369, opened 2017-11-08
Make custom action methods more consistent

ob-stripe:
r? @brandur-stripe 

Custom methods (e.g. pay an invoice, reject an account, etc.) were handled inconsistently throughout the library:
- some updated the instance but did not return it (`Payout.cancel`, `Transfer.cancel`)
- some returned the new object but did not update the instance (`Invoice.pay`, `Order.pay`)
- some did not handle idempotency keys (`Source.detach`, `Payout.cancel`, `Transfer.cancel`)
- some did not accept parameters, or accepted a single specific parameter (those were not a problem as the API does not accept parameters for those, but we might as well make everything more consistent as well as future-proof the library)

~I have not written any new tests because I'm fairly confident about these changes and because @remi-stripe is working on updating the tests to use stripe-mock in #367.~

I did add some tests after all.

### DIFF
--- a/stripe/api_resources/account.py
+++ b/stripe/api_resources/account.py
@@ -41,16 +41,10 @@ class Account(CreateableAPIResource, ListableAPIResource,
     def instance_url(self):
         return self._build_instance_url(self.get('id'))
 
-    def reject(self, reason=None, idempotency_key=None):
+    def reject(self, idempotency_key=None, **params):
         url = self.instance_url() + '/reject'
         headers = util.populate_headers(idempotency_key)
-        if reason:
-            params = {"reason": reason}
-        else:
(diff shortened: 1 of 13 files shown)


brandur-stripe:
Thanks for finding all of these and getting this patch written! It's really nice to see some better consistency here, and it'll pay dividends as there are now fewer (and hopefully zero) poorly-written methods that can be accidentally copied and pasted as new APIs ar being implemented.

It looks like in some cases we went from returning a new object to mutating the existing one, but I suspect that for most people's uses it won't make a difference — I'll bump as minor.

brandur-stripe:
Released as 1.75.0.

PR #372, opened 2017-11-13
StripeObject: only add changed values to _unsaved_values

blueyed:


ob-stripe:
Hey @blueyed, thanks a lot for the contribution! This looks very promising.

> Doesn't it empty legal_entity.address?

So, no, although I can see why you'd think so! We mock `APIRequestor`, which can receive empty dicts. Those will be ignored in the outgoing request though:

```python
>>> list(stripe.api_requestor._api_encode({'metadata': {}}))
[]
```

I agree it would be better and clearer if `APIRequestor` would not receive empty dicts in the first place! I'll take a closer look at your patch today.

stripe/stripe_object.py
```diff
@@ -237,7 +238,9 @@ def serialize(self, previous):
             elif isinstance(v, stripe.api_resources.abstract.APIResource):
                 continue
             elif hasattr(v, 'serialize'):
-                params[k] = v.serialize(previous.get(k, None))
+                child = v.serialize(previous.get(k, None))
+                if child:
```
ob-stripe: I'm not sure if it's possible in practice, but in theory `serialize` could return an empty string `""` to unset the entire attribute, and because empty strings are falsey in Python this would get ignored. It might be better to explicitly compare the returned value to an empty dict.

ob-stripe:
I left a comment, but generally this looks pretty great to me!

r? @deontologician-stripe Do you mind taking a look at this?

stripe/stripe_object.py
blueyed: Makes sense.
Should have a test for it then, too.

deontologician-stripe:
I am +1 on this. I wish we could do the checking for `_unsaved_values` in [`__setstate__`](https://docs.python.org/3/library/pickle.html#handling-stateful-objects) instead of just in time when it's needed, but that is unrelated to this PR.

stripe/stripe_object.py
blueyed: > if it's possible in practice

The two internal `serialize` methods will return a dict always, so it is unlikely to come from there.

Added a test with a specialized StripeObject, which is likely what you had in mind?!

ob-stripe: Yes, exactly. Thanks again!

ob-stripe:
lgtm

ob-stripe:
@dpetrovics-stripe (dev-platform run, feel free to re-assign) Mind merging this and releasing a version? Thanks!

PR #373, opened 2017-11-15
Fix items serialization when items is a dict

ob-stripe:
r? @dpetrovics-stripe (DP run, feel free to re-assign)

Fixes an issue where if you set `items` to an integer-indexed dict on a subscription that was retrieved from the API, the library would try to unset the properties of the list object that was returned in the `items` attribute.

This is basically exactly the same problem as the 2nd issue mentioned here: https://github.com/stripe/stripe-ruby/pull/596, and the fix is similar (simply ignore the previous value of `items`).

### DIFF
--- a/stripe/api_resources/subscription.py
+++ b/stripe/api_resources/subscription.py
@@ -32,8 +32,13 @@ class Subscription(CreateableAPIResource, DeletableAPIResource,
         return super(Subscription, cls).create(**params)
 
     def serialize(self, previous):
-        updated_params = super(UpdateableAPIResource, self).serialize(previous)
-        if "items" in updated_params:
-            updated_params["items"] = util.convert_array_to_dict(
-                updated_params["items"])
-        return updated_params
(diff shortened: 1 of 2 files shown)


grey-stripe:
I'm fairly nervous about this change, I don't think we should be digging around in the parent class's implementation details, but I'll need to investigate a bit to see if there's a better way to approach this.

brandur-stripe:
Yeah, it is a little unfortunate that we need to go up and touch the parent's internals.

Just trying to think of some more options: I guess one possibility is to expose a "protected" (in the nomenclature of other OO languages that support more OO stuff) API on the parent that's designed to be called from a child for exactly this purpose (e.g. `unset_previous` or something).

ob-stripe:
Closing this super old PR. The issue was probably fixed by #473, though the current state of the library makes it difficult to write a test for this (since the encoding happens in the `api_requestor` class, which we mock in tests).

PR #375, opened 2017-11-20
Use custom JSON encoder to handle datetime objects

### LINKED ISSUE #374: Possible to enter state where Account.__repr__() results in Type error in attempting to serialize Datetime object
In attempting to make mutations to an Account object, I can set the following:

```
account = stripe.Account.retrieve("acc_123") 
account['tos_acceptance'] = {
     'date': datetime.datetime(year = 2017, month = 11, day=16),
     'ip': '192.168.1.1' #For example's sake
    }
```

Now, if I run `account.__repr__()` (or just `account` in some sort of REPL, I get the following: 

```
TypeError: Object of type 'datetime' is not JSON serializable
```

Given that in the `Account.save()` method, Python `datetime` objects automatically get translated into Stripe API comprehensible timestamps, it would probably make sense to do this within the `__repr__` as well for inspection in iPython, ipynb, IDEs or elsewhere I'm not thinking of.

ob-stripe:
r? @dpetrovics-stripe (dev-platform run)
cc @stripe/api-libraries @dalanmiller

Fixes #374.

### DIFF
--- a/stripe/stripe_object.py
+++ b/stripe/stripe_object.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import datetime
 import warnings
 from copy import deepcopy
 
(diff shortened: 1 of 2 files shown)


grey-stripe:
LGTM

PR #380, opened 2017-12-08
Remove unneeded OAuthError class in error.py

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Removes the `OAuthError` class from `error.py`. The actual `OAuthError` class is defined in `oauth_error.py`.


brandur-stripe:
Nice catch. LGTM.

PR #383, opened 2017-12-24
Fix warnings in tests

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fix some test warnings:
- a few tests used the deprecated `assertEquals` function instead of `assertEqual`
- explicitly silenced the deprecation warnings in the legacy subscription tests


brandur-stripe:
+1. LGTM.

PR #385, opened 2018-01-05
Drop support for Python 2.6 and 3.3

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Drops support for Python 2.6 and 3.3, and removes some no longer needed stuff (simplejson, etc.).

The merge target is `integration-v2` so we can bundle other breaking changes in the next major version.


brandur-stripe:
Great to see that crazy JSON patching go away. LGTM.

PR #386, opened 2018-01-05
stripe-python 2.0.0

### LINKED ISSUE #376: Better exception for Source.detach() on detached sources
https://github.com/stripe/stripe-python/blob/8629450495adf3ea6d2c1ae343a99c8a8480e90e/stripe/api_resources/source.py#L29

Could this get its own exception type? I want to give customers the option to remove a source from their payment methods, and right now I have to do something like this:

```py
try:
    source.detach()
except NotImplementedError:
    resync(source)
```

Catching notimpl is weird, and this isn't really the appropriate exception type I think anyway.

ob-stripe:
- [x] Drop support for Python 2.6 and 3.3 (#385)
- [x] Use py.test for tests (#384)
- [x] Remove deprecated code (#399)
- [x] Remove `util.json` and use `json` module directly everywhere (#402)
- [x] Update `setup.py` and test flow (#403)
- [x] Use pipenv (#410)
- [x] Change exception when detaching unattached sources from `NotImplementedError` to `stripe.error.InvalidRequestError` (#415)

Fixes #376.


Makefile
```diff
@@ -0,0 +1,11 @@
+test:
+	pipenv run detox
+
+ci:
+	pipenv run py.test --cov=stripe -n 8
```
blueyed: The preferred name is just "pytest" (also above).

Pipfile
```diff
@@ -0,0 +1,27 @@
...
+
+[dev-packages]
+
+pytest = ">=3.4"
+pytest-mock = ">=1.7"
+pytest-xdist = ">=1.22"
+pytest-cov = ">=2.5"
+tox = "*"
+detox = "*"
+"flake8" = "*"
```
blueyed: Does "flake8" need to be quoted?

tox.ini
```diff
@@ -4,39 +4,22 @@
...
-[coverage:run]
-source = stripe
-omit =
-    stripe/six.py
+[testenv:lint]
+description = run static analysis and style check using flake8
+basepython = python2.7
+deps = flake8
+commands = python -m flake8 --show-source stripe setup.py {posargs}
+           python -m flake8 --show-source tests {posargs}
```
blueyed: Could be a single command (moving/including `tests` to/with the first).

Makefile
ob-stripe: 👍 Thanks! Will fix.

Pipfile
ob-stripe: Good question! I'm not entirely sure why pipenv wraps some package names in quotes, but I figure it knows what it's doing :)

tox.ini
ob-stripe: 👍

Pipfile
jleclanche: It doesn't need to be quoted (source: https://github.com/toml-lang/toml#keys), but pipenv's toml library (prettytoml) is a bit overzealous and quotes any package name with a digit in it.

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Alright, I think this is ready to be released. All PRs included in this branch have already been approved. Let me know if you have any concerns.

brandur-stripe:
Very exciting :)

LGTM.

jleclanche:
:tada:

ob-stripe:
Alrighty, let's go :)

PR #387, opened 2018-01-05
Remove deprecated helpers from all methods

remi-stripe:
When pushing v2.0.0 for stripe-python, we want to remove old helpers or methods that have been deprecated for a long time and shouldn't be used anymore.

cc @stripe/api-libraries


stripe/api_resources/customer.py
```diff
@@ -20,63 +20,3 @@
 class Customer(CreateableAPIResource, UpdateableAPIResource,
                ListableAPIResource, DeletableAPIResource):
     OBJECT_NAME = 'customer'
-
```
remi-stripe: @ob-stripe do we want to remove the `stripe.Customer.*_source()` methods?

ob-stripe: No, those were recently added and are useful (they let users manage payment sources without retrieving the customer object first). We should try to document them somewhere though.

ob-stripe:
You have a few linting errors, but lgtm otherwise!

stripe/api_resources/customer.py
remi-stripe: I know they were added but I do wonder if there should be a different way to handle them. Deferring to you here obviously as I know you added them and know the library better than I do.

remi-stripe:
@ob-stripe The `encode*` methods on `api_requestor.py` are deprecated per our code but they are still used in the library. For example [here](https://github.com/stripe/stripe-python/blob/60aeefaf22f96789968926c39657d3d7853e5613/bin/stripe#L41) so I'm not sure why we remove them and what to replace it with. In general there are a ton of things to remove in that file that I'm a bit confused about. Removing those mostly works in test but my guess is that it's because we don't have enough coverage in all the code.

remi-stripe:
Okay I removed the previous commit. I realize this should likely live in a separate PR because it's not about helpers on classes and more about cleaning up deprecated methods instead.

remi-stripe:
@ob-stripe PTAL

ob-stripe:
Closing this one in favor of #399. Sorry @remi-stripe!

PR #389, opened 2018-01-12
Register unsaved attributes on assignment regardless of new value

### LINKED ISSUE #388: Changing subscription's plan ignores quantity
## Description

When a customer has an existing subscription with quantity 2 and plan `plan1`, updating the subscription's plan always uses quantity 1.

## Example

```python
subscription = stripe.Subscription.retrieve({SUBSCRIPTION_ID})
assert subscription.plan == 'plan1'
assert subscription.quantity == 2
subscription.plan = 'plan2'
subscription.quantity = 2
subscription.save()

subscription = stripe.Subscription.retrieve({SUBSCRIPTION_ID})
assert subscription.quantity == 2
AssertionError
```

Stripe logging shows only `plan_id` is sent to the api when calling `subscription.save()`.

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

When setting an attribute on a Stripe object, register the attribute as unsaved regardless of whether the new value is different from the existing value.

This aligns the behavior with that of our Ruby and PHP libraries.

Fixes #388.

### DIFF
--- a/stripe/stripe_object.py
+++ b/stripe/stripe_object.py
@@ -98,12 +98,11 @@ class StripeObject(dict):
                 "You may set %s.%s = None to delete the property" % (
                     k, str(self), k))
 
-        if not hasattr(self, k) or v != getattr(self, k):
-            # Allows for unpickling in Python 3.x
-            if not hasattr(self, '_unsaved_values'):
-                self._unsaved_values = set()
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Great fix! (And thanks for checking for prior art in other libraries.)

PR #390, opened 2018-01-17
Update README.md to include link to stripe-mock

tedkornish:
Looks like this link was missing and was breaking some formatting.


ob-stripe:
👍  Thanks for the catch and the fix!

PR #394, opened 2018-02-08
Make last_response available on save()

### LINKED ISSUE #393: StripeObject.last_response is None on save() calls
Repro steps from #371:

```py
customer = stripe_customer.api_retrieve()
customer.account_balance -= 250
customer.save(idempotency_key=ik)
credit_request_id = customer.last_response.request_id  # Breaks! last_response is None
```

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @jleclanche

Fixes #393.

When calling `save()`, `convert_to_stripe_object` will call `refresh_from` with a `StripeObject` instance that already has the `_last_response` attribute set. However `refresh_from` only set the the `_last_response` attribute from an explicit `last_response` argument and did not check whether the instance already had a `_last_response` attribute like it does for other memorized attributes (e.g. `stripe_account`). This PR fixes that, and adds a bunch of tests to ensure that `last_response` is available on instances returned by all common API operations.

While fixing this, I also noticed a regression introduced by #371: the `OAuth.token` and `OAuth.deauthorize` methods used to directly return a dict with the JSON values returned by `connect.stripe.com`. After #371, those methods instead returned a `StripeResponse` instance. I've fixed it so that they return a dict again.

### DIFF
--- a/stripe/oauth.py
+++ b/stripe/oauth.py
@@ -40,7 +40,7 @@ class OAuth(object):
         requestor = api_requestor.APIRequestor(api_base=connect_api_base)
         response, api_key = requestor.request(
             'post', '/oauth/token', params, None)
-        return response
+        return response.data
 
     @staticmethod
     def deauthorize(**params):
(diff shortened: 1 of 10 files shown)


jleclanche:
Yep, looks good.

brandur-stripe:
Woo, great eye Olivier, and thanks for adding those additional tests/asserts!

LGTM.

PR #396, opened 2018-02-16
Add support for /v1/topups endpoints

jkakar-stripe:
This add standard retrieve, create and update client support for the new `/v1/topups` endpoint.

r? @apakulov-stripe @ccontinanza-stripe @chellman-stripe @kenneth-stripe @miguel-stripe
r? @stripe/api-libraries

### DIFF
--- a/stripe/api_resources/topup.py
+++ b/stripe/api_resources/topup.py
@@ -1,0 +1,7 @@
+from stripe.api_resources.abstract import CreateableAPIResource
+from stripe.api_resources.abstract import UpdateableAPIResource
+from stripe.api_resources.abstract import ListableAPIResource
+
+
+class Topup(CreateableAPIResource, ListableAPIResource, UpdateableAPIResource):
+    OBJECT_NAME = 'topup'
--- a/tests/api_resources/test_topup.py
+++ b/tests/api_resources/test_topup.py
@@ -1,0 +1,58 @@
+import stripe
+from tests.helper import StripeTestCase
+
+
+TEST_RESOURCE_ID = 'tu_123'
+
+
+class TopupTest(StripeTestCase):
+    def test_is_listable(self):
+        resources = stripe.Topup.list()
+        self.assert_requested(
+            'get',
+            '/v1/topups'
+        )
+        self.assertIsInstance(resources.data, list)
+        self.assertIsInstance(resources.data[0], stripe.Topup)
+
+    def test_is_retrievable(self):
+        resource = stripe.Topup.retrieve(TEST_RESOURCE_ID)
+        self.assert_requested(
+            'get',
+            '/v1/topups/%s' % TEST_RESOURCE_ID
+        )
+        self.assertIsInstance(resource, stripe.Topup)
+
+    def test_is_creatable(self):
+        resource = stripe.Topup.create(
+            amount=100,
+            currency='usd',
+            source='src_123',
+            description='description',
+            statement_descriptor='statement descriptor'
+        )
+        self.assert_requested(
+            'post',
+            '/v1/topups'
+        )
+        self.assertIsInstance(resource, stripe.Topup)
+
+    def test_is_saveable(self):
+        resource = stripe.Topup.retrieve(TEST_RESOURCE_ID)
+        resource.metadata['key'] = 'value'
+        resource.save()
+        self.assert_requested(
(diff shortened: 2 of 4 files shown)


jkakar-stripe:
This branch is stacked on #395. I'll do the rebase when that lands, but want to get these changes out for reviews.

stripe/api_resources/topup.py
```diff
@@ -0,0 +1,9 @@
+from __future__ import absolute_import, division, print_function
```
miguel-stripe: Are these imports used anywhere?

tests/api_resources/test_topup.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import absolute_import, division, print_function
```
miguel-stripe: Are these imports needed?

tests/api_resources/test_topup.py
```diff
@@ -0,0 +1,60 @@
...
+
+import stripe
+from tests.helper import StripeTestCase
+
+
+TEST_RESOURCE_ID = 'tu_123'
+
+
```
miguel-stripe: Nit: Unnecessary extra lines?

tests/api_resources/test_topup.py
```diff
@@ -0,0 +1,60 @@
...
+    def test_is_modifiable(self):
+        resource = stripe.Topup.modify(
+            TEST_RESOURCE_ID,
+            metadata={'key': 'value'}
+        )
+        self.assert_requested(
+            'post',
+            '/v1/topups/%s' % TEST_RESOURCE_ID
+        )
+        self.assertIsInstance(resource, stripe.Topup)
```
miguel-stripe: Worth checking the `resource` has its `metadata` set to `{'key': 'value'}`?

stripe/api_resources/topup.py
```diff
@@ -0,0 +1,9 @@
...
+
+from stripe.api_resources.abstract import CreateableAPIResource
+from stripe.api_resources.abstract import UpdateableAPIResource
+from stripe.api_resources.abstract import ListableAPIResource
+
+
```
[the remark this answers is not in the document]
jkakar-stripe: This is the style used in other files, so I copied. I'll leave it as is to maintain consistency.

stripe/api_resources/topup.py
jkakar-stripe: Apparently not. I removed them.

tests/api_resources/test_topup.py
jkakar-stripe: Apparently not. I removed them.

tests/api_resources/test_topup.py
jkakar-stripe: I tried this, but it looks like stripe-mock doesn't actually mutate received data correctly. I believe it just validates what it received and returns stock data.

miguel-stripe: 👍

stripe/api_resources/topup.py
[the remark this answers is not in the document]
brandur-stripe: It seems a little weird coming from other languages, but believe it or not, this is conformant with the most universally accepted style for Python, [PEP 8](https://www.python.org/dev/peps/pep-0008). From the spec:

> Surround top-level function and class definitions with two blank lines.

There's a check that will fail the test suite if the style is off, so you're safe from pulling something in that's wrong, but I'd suggest as a general guideline just doing what Jamu did and copying the style seen elsewhere.

tests/api_resources/test_topup.py
brandur-stripe: (Similarly to the comment above, I believe PEP 8 will require this doubled up blank line too.)

tests/api_resources/test_topup.py
brandur-stripe: Yes that's correct Jamu. We want to fix this at some point, but for now it's not working checking any of the data that you get back.

brandur-stripe:
Maybe just ack that we're good to go here as well, and we'll pull it in. Thanks!

apakulov-stripe:
lgtm

brandur-stripe:
LGTM.

brandur-stripe:
Released as 1.78.0.

PR #397, opened 2018-02-21
Start building universal wheels

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Since we no longer use `2to3` and the library can now directly be used by Python 2 or 3, we also no longer need to build distinct wheels.

This PR enables the flag to start building universal wheels by default, as well as add a few opportunistic fixes to the `setup.py` and license files.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License
 
-Copyright (c) 2010-2011 Stripe (http://stripe.com)
+Copyright (c) 2010-2018 Stripe (http://stripe.com)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,5 @@
+[bdist_wheel]
+universal = 1
+
+[metadata]
+license_file = LICENSE
--- a/setup.py
+++ b/setup.py
@@ -28,9 +28,9 @@ else:
 with open('LONG_DESCRIPTION.rst') as f:
     long_description = f.read()
 
-# Don't import stripe module here, since deps may not be installed
-sys.path.insert(0, os.path.join(os.path.dirname(__file__), 'stripe'))
-from version import VERSION
+version_contents = {}
+with open(os.path.join('stripe', 'version.py')) as f:
+    exec(f.read(), version_contents)
 
 # Get simplejson if we don't already have json
 if sys.version_info < (3, 0):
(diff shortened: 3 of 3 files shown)


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License
 
-Copyright (c) 2010-2011 Stripe (http://stripe.com)
+Copyright (c) 2010-2018 Stripe (http://stripe.com)
```
ob-stripe: This one hadn't been updated in a while ;)

setup.cfg
```diff
@@ -0,0 +1,5 @@
+[bdist_wheel]
+universal = 1
```
ob-stripe: This enables the build of universal wheels by default. Cf. https://packaging.python.org/tutorials/distributing-packages/#universal-wheels

setup.cfg
```diff
@@ -0,0 +1,5 @@
...
+
+[metadata]
+license_file = LICENSE
```
ob-stripe: This is to ensure the license file is included in the wheel files. Cf. https://wheel.readthedocs.io/en/stable/#including-the-license-in-the-generated-wheel-file.

setup.py
```diff
@@ -28,9 +28,9 @@
 with open('LONG_DESCRIPTION.rst') as f:
     long_description = f.read()
 
-# Don't import stripe module here, since deps may not be installed
-sys.path.insert(0, os.path.join(os.path.dirname(__file__), 'stripe'))
-from version import VERSION
+about = {}
+with open(os.path.join('stripe', 'version.py')) as f:
+    exec(f.read(), about)
```
ob-stripe: This is slightly cleaner than importing, and removes a `flake8` warning about imports needing to be grouped at the top of the file. (Note that we `setup.py` is not automatically validated by flake8 currently.)

ob-stripe:
(This is tagged as WIP because we'll also need to update our internal tooling for cutting releases to account for this change.)

ob-stripe:
@brandur-stripe PTAL

LICENSE
brandur-stripe: lol. Maybe another seven years until the next time too ...

setup.py
brandur-stripe: Kind of a nit again, but could change `about` to something a little more descriptive like ` version_contents`?

Love the same though. The old code is tremendously opaque by comparison.

brandur-stripe:
Minor comment above, but LGTM.

Having very recently ran into a problem with a package missing from one particular Python version and failing my whole build, this is a pretty exciting change. Thanks OB!

setup.py
ob-stripe: 👍

PR #399, opened 2018-02-25
Remove deprecated code

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Cleanup for version 2.0:
- remove `bin/stripe` client
- remove a lot of deprecated methods in `APIRequestor`
- remove deprecated `all()` method for list requests (the proper method is `list()`)
- remove legacy subscription management methods in `Customer`
- remove support for `api_version` argument in `EphemeralKey.create()` (the proper argument is `stripe_version`)
- remove `delete()` method in `Source` (the proper method is `detach()`)
- remove the comment about removing arg in `delete_discount()` methods in `Customer` and `Subscription`. Instead, support sending params to the API. (Those methods currently don't take accept any parameters, but generally the library doesn't try to enforce which methods support parameters and which don't. Invalid arguments will be declined server-side.)
- remove warning on `StripeObject.to_dict()`. The method is trivial, might as well keep supporting it.
- clean up global `stripe` namespace. Invididual exception classes are no longer imported directly in the global namespaces, as well as other internal classes and methods.


bin/stripe
```diff
@@ -1,344 +0,0 @@
-#!/usr/bin/env python
```
brandur-stripe: Geeze, I didn't even know this was a thing.

brandur-stripe:
Amazing that you've been keeping track of all these little inconsistencies all along. Nice!

LGTM.

PR #402, opened 2018-02-28
Remove util.json and use json module directly everywhere

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Since we dropped simplejson, there's no point in centralizing all JSON usage through the `util` module anymore. Other modules can simply import the stdlib `json` module as needed.


brandur-stripe:
Oh man, the code quality here has improved like 2,000% YOY. Nice!

LGTM.

PR #403, opened 2018-03-05
Update setup.py and test flow

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

(This is for the upcoming v2.0.)

Update setup.py and test flow:
- refactor `setup.py` to use best practices described in https://github.com/pypa/sampleproject/blob/master/setup.py
- bump minimum `requests` version to 2 (current major version)
- add `requests[security]` to dependencies for Python 2. This should help with TLS 1.2 issues, e.g. when using the default OS X Python interpreter.
- removed `pytest-runner` from `setup_requires` to avoid installing the package for every `setup.py` invocation. Instead use a custom `PyTest` command as described [here](https://docs.pytest.org/en/latest/goodpractices.html#manual-integration).
- use `pytest-xdist` to distribute tests over multiple cores. This is not very useful when running tests on a reasonably powerful local machine, but gives a noticeable speed increase on Travis.
- use `find_packages` to discover packages automatically

- Greatly simply `tox.ini` by putting coverage and flake8 configurations in dedicateds files, and simply running `python setup.py test`

- Don't install `pycurl` package when running tests as it wasn't actually used anywhere

- Remove `allow_failures` from Travis configuration as we probably want to not merge PRs that break PyPy or 3.7 compatibility

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,4 @@
+[run]
+source = stripe
+omit =
+    stripe/six.py
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,3 @@
+[flake8]
+exclude =
+    six.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,21 +12,13 @@ python:
   - "pypy3"
 
 cache:
-  apt: true
   directories:
     - stripe-mock
   pip: false
 
 env:
   global:
-    - PYCURL_SSL_LIBRARY=gnutls
     - STRIPE_MOCK_VERSION=0.8.0
-
-addons:
-  apt:
-    packages:
-      - libcurl4-gnutls-dev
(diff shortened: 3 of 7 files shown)


brandur-stripe:
Nice cleanups!

Just to confirm, all test invocations are still the same right? i.e., no changes are required in `README.md` or anything?

LGTM.

ob-stripe:
> Just to confirm, all test invocations are still the same right? i.e., no changes are required in README.md or anything?

Oh, good point, my update to `tox.ini` broke the possibility of passing custom parameters. This is now fixed: if you don't pass any parameters, then it will run with `-n auto` (which spreads tests over as many workers as there are CPU cores), otherwise it will use your parameters. The examples from the README will still work.

PR #406, opened 2018-03-12
Hide .pyc files and __pycache__ folders in VSCode

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Just a nice little improvement for developers using VSCode. This hides `.pyc` files (with a matching `.py` file) and `__pycache__` folders from the sidebar.


brandur-stripe:
Nice!

LGTM.

PR #408, opened 2018-03-12
Add close() method to HTTPClient classes

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

We're seeing intermittent test failures because unclosed requests sessions can cause `ResourceWarning`s (cf. https://github.com/requests/requests/issues/3912) and some of our tests capture and count warnings.

This PR adds a `close()` method to all `HTTPClient` subclasses. The method is a no-op for all clients except `RequestClient`, which cleanly closes the session (if it exists). The PR also updates the `StripeTestCase` class to manually instantiate a client so that we can call `close()` in the `tearDown()` method. This should hopefully get rid of the unwanted `ResourceWarning`s.

### DIFF
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -93,6 +93,10 @@ class HTTPClient(object):
         raise NotImplementedError(
             'HTTPClient subclasses must implement `request`')
 
+    def close(self):
+        raise NotImplementedError(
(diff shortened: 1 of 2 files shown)


brandur-stripe:
This is awesome. (And nice detective work too.)

LGTM.

PR #409, opened 2018-03-13
Configure VSCode linting

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Configure linting with flake8 when using VSCode.

(Technically, comments are not allowed in JSON, but VSCode supports this.)

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,5 @@
+{
+    "recommendations": [
+        "ms-python.python"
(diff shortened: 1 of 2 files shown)


brandur-stripe:
LGTM.

> (Technically, comments are not allowed in JSON, but VSCode supports this.)

Hah, I love how GitHub colors them in the diff with a color I'd only describe as ALARMING RED.

PR #410, opened 2018-03-13
Use pipenv for managing developement virtualenv

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This PR ended up being larger than I intended, so here's some helpful information:

- [pipenv](https://github.com/pypa/pipenv) is a fancy new tool for managing development virtualenvs. It's fairly recent, but is officially recommended by [python.org](https://packaging.python.org/key_projects/). (It's also developed by Kenneth Reitz of requests fame, and is used by requests itself.)
- `Pipfile` and `Pipfile.lock` work similarly to `Gemfile` in Ruby projects. I've committed the lock file because it's recommended (cf. https://github.com/pypa/pipenv/issues/598) and is what's done in requests.
- Added a `Makefile` for common operations
- Updated `.travis.yml` to use the new `make` targets
- Updated `tox.ini` to set up a separate environment for flake8 linting

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,11 @@
+test:
+	pipenv run detox
+
+ci:
+	pipenv run py.test --cov=stripe -n 8
+
+coveralls:
+	pipenv run coveralls
+
+lint:
+	pipenv run tox -e lint
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,27 @@
+[[source]]
+
+url = "https://pypi.python.org/simple"
+verify_ssl = true
+name = "pypi"
+
+
+[packages]
+
+"stripe" = {path = ".", editable = true}
+
+
+[dev-packages]
+
+pytest = ">=3.4"
+pytest-mock = ">=1.7"
+pytest-xdist = ">=1.22"
+pytest-cov = ">=2.5"
+tox = "*"
+detox = "*"
+"flake8" = "*"
+coveralls = "*"
+
+
+[requires]
+
+python_version = "3.6"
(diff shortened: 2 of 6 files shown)


Pipfile
```diff
@@ -0,0 +1,27 @@
+[[source]]
+
+url = "https://pypi.python.org/simple"
+verify_ssl = true
+name = "pypi"
+
+
+[packages]
+
+"e1839a8" = {path = ".", editable = true}
```
ob-stripe: This line basically means that pipenv will run `python setup.py install`, which will pull the install dependencies from `setup.py` according to the Python environment.

brandur-stripe: Any idea what "e1839a8" is?

Makefile
```diff
@@ -0,0 +1,15 @@
+init:
+	pip install --upgrade pipenv
+	pipenv install --dev
```
brandur-stripe: Do you mind if we remove the `init` target in favor of just having this duplicated in `.travis.yml` and the `README.md`? I feel like many Python people will already have have a `pipenv` available, and there are other recommended ways of installing it (e.g., I used Homebrew as recommended on their page).

brandur-stripe:
Very cool! I just tried it to make sure it wasn't too much trouble, and your instructions in the `README.md` went off without a hitch :)

I left a few comments above, but otherwise LGTM.

Pipfile
ob-stripe: It's a short hash of the path (in this case `.`) automatically generated by pipenv to avoid collisions if you have multiple entries. I think it can actually be anything unique -- I'll try renaming it to `stripe` for clarity.

Makefile
ob-stripe: 👍

ob-stripe:
Okay, I made the requested changed. Going to pull this in.

PR #411, opened 2018-03-13
Remove stub for source_transaction test

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

stripe-mock supports this endpoint now, so the stub is unnecessary.


brandur-stripe:
Nice :)

LGTM.

PR #412, opened 2018-03-14
flake8: add "unicode" to builtins

blueyed:
This avoids "F821 undefined name 'unicode'" when using flake8 with
Python 3.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -3,3 +3,6 @@ universal = 1
 
 [metadata]
 license_file = LICENSE
+
+[flake8]
+builtins = unicode


ob-stripe:
Hey @blueyed. There's already a `[flake8]` configuration section in `tox.ini`, so any additional configuration should go there and not in `setup.cfg`.

That said, I'm not sure we should pull this in. The false positives are annoying but there's also a risk of masking real issues. Since the library now vendors six, maybe we could replace `unicode()` calls with [`six.text_type()`](https://pythonhosted.org/six/#six.text_type) -- it might also help us get rid of some `if six.PY2` / `if six.PY3` blocks. I'll experiment and give an update here.

blueyed:
@ob-stripe 
Sounds good.
Otherwise it could also use `# noqa: F821` explicitly in the code.

ob-stripe:
Closing this in favor of #413.

PR #413, opened 2018-03-14
Remove unicode symbols, use six helpers instead

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @blueyed 

This PR replaces all uses of the `unicode` symbol with helper methods from six. This fixes flake8 warnings when running flake8 with Python 3 ( as `unicode` is not a valid symbol in Py3).

Replaces #412.


brandur-stripe:
Very nice!

LGTM.

PR #414, opened 2018-03-14
Add support for running unit tests from VSCode

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

As the title says.

Because we don't enforce a specific interpreter in the VS Code settings, the unit tests will be run with the default interpreter, so it's the responsibility of the developer to ensure that the `unittest2` and `mock` modules are available with the default interpreter.

In the 2.0 branch, this will not be an issue because VSCode will automatically use the Python interpreter from the pipenv virtual environment, which will have all the required dependencies.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -7,5 +7,10 @@
     // Linting
     "python.linting.enabled": true,
     "python.linting.pylintEnabled": false,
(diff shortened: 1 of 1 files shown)


brandur-stripe:
LGTM.

PR #415, opened 2018-03-15
Change exception type when detaching unattached sources to InvalidRequestError

### LINKED ISSUE #376: Better exception for Source.detach() on detached sources

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @jleclanche

Changes the exception raised when trying to detach unattached sources from `NotImplementedError` to `stripe.error.InvalidRequestError`. This is consistent with other libraries.

Fixes #376.

### DIFF
--- a/stripe/api_resources/source.py
+++ b/stripe/api_resources/source.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-from stripe import util
+from stripe import error, util
 from stripe.api_resources import Customer
 from stripe.api_resources.abstract import CreateableAPIResource
 from stripe.api_resources.abstract import UpdateableAPIResource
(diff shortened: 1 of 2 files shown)


jleclanche:
:+1:

brandur-stripe:
Much more sensible.

LGTM.

PR #422, opened 2018-04-21
Add property accessor for error messages

seants:
In python3, it's no longer possible to do `error.message`.

The official python alternative of `str(error)` does not work with the stripe library because `__str__` is overridden and would print 
```
Request ch_ae7361827ab7: Your card was Declined.
``` 
which is not user-friendly. The three alternatives I've identified are
```python
error.args[0]
```
```python
error._message
```
```python
super(stripe.error.StripeError, error).__str__()
```
none of which are super elegant. This PR would allow
```python
error.plain_message
```
instead. Happy to rename it to something else too.

### DIFF
--- a/stripe/error.py
+++ b/stripe/error.py
@@ -32,6 +32,10 @@ class StripeError(Exception):
         else:
             return self._message
 
+    @property
+    def plain_message(self):
+        return self._message
+
 
 class APIError(StripeError):
     pass
--- a/stripe/error.py
+++ b/stripe/error.py
@@ -33,7 +33,11 @@ class StripeError(Exception):
             return self._message
 
     @property
-    def plain_message(self):
+    def user_message(self):
+        # Returns the underlying Exception (base class) message,
+        # which was available in python2 via error.message. As
+        # opposed to str(error), this omits the Request ch_...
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Ah, I see! Thanks for the clarification.

@ob-stripe Do you have an opinion on this one? I could see either doing something similar to the suggested above, and I could also see inverting the problem by removing the "Request xxx: " prefix from the standard string representation and adding a `message_with_request` method that makes it available.

seants:
That would've been nice initially but is a breaking change for all clients that depend on that behavior. I could totally imagine someone doing `str(error)[20:]` to chop that first part off. This PR is kind of a workaround for that so that it can be released anytime.

seants:
In terms of naming, I was gonna go for `user_message` but I didn't do a deep-dive into whether that message is always user-friendly. As mentioned, happy to rename.

seants:
Haha no worries; yes I'll take care of those this weekend!

seants:
Updated!

brandur-stripe:
Thanks!

Released as 1.82.0.

seants:
Thanks!

PR #423, opened 2018-04-24
Adds a couple more error assertions for Python 3

brandur:
This is basically a no-op, but just adds a few more test case branches
for Python 3 to demonstrate what `str(...)` on an error should be
expected to return in that version (instead of just on Python 2).

r? @ob-stripe


brandur-stripe:
Thanks!

PR #426, opened 2018-05-09
Upgrade stripe-mock to 0.16.0

fay-stripe:
Routing to 120 path(s) and 202 endpoint(s) with 110 validator(s)
Listening on port 12111 to v0.16.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ cache:
 env:
   global:
     - PYCURL_SSL_LIBRARY=gnutls
-    - STRIPE_MOCK_VERSION=0.16.0
+    - STRIPE_MOCK_VERSION=0.16.1
 
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Hm, unfortunately it looks like there's another problem with the upgrade here which was introduced by another recent tweak that I made in stripe-mock.

I'll take over this pull request if you want to take a look at the other libraries for now.

fay-stripe:
Sure - thanks! 

Also, I notice that when I commit with the message "Upgrade `stripe-mock` to v0.16.0" the commit message changes to "Upgrade stripe-mock to 0.16.0" and adds two more lines "Routing to 120...." and "Listening on...". Do you know why that's happening?

brandur-stripe:
> Also, I notice that when I commit with the message "Upgrade stripe-mock to v0.16.0" the commit message changes to "Upgrade stripe-mock to 0.16.0" and adds two more lines "Routing to 120...." and "Listening on...". Do you know why that's happening?

I'm afraid to say that this is something to do with your local Git configuration. I could help you take a look, but it'd probably be worth taking a look at whatever customizations you have and whether they're working as expected.

PR #430, opened 2018-05-14
Handle the case where both error messages are None and request ids are None.

jfly:
https://github.com/stripe/stripe-python/pull/196 fixed this TypeError for
when request_id is set and message is None, but it didn't handle the
case where both were None.

### DIFF
--- a/stripe/error.py
+++ b/stripe/error.py
@@ -26,11 +26,11 @@ class StripeError(Exception):
         self.request_id = self.headers.get('request-id', None)
 
     def __str__(self):
+        msg = self._message or "<empty message>"
         if self.request_id is not None:
-            msg = self._message or "<empty message>"
             return u"Request {0}: {1}".format(self.request_id, msg)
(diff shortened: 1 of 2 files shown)


brandur-stripe:
LGTM (and thanks for the test).

brandur-stripe:
And just out of curiosity: did you run into a came where a request ID actually was `None` in production, or is this more for testing synthetic errors?

brandur-stripe:
Released as 1.82.1.

jfly:
Wow, thanks for the quick merge and release!

This was a very synthetic issue: we were constructing an StripeError via `StripeError()`, and then running into this TypeError.

brandur-stripe:
No worries!

> This was a very synthetic issue: we were constructing an StripeError via StripeError(), and then running into this TypeError.

Okay, thanks! I just wanted to make sure that nothing unexpected was going on with the API.

PR #431, opened 2018-05-17
Fix some parameters being sent in tests

brandur:
Fixes one parameter being sent so that the suite is compliant with a
version of stripe-mock that's checking for extra parameters.

See https://github.com/stripe/stripe-ruby/pull/650 as another sample
with more context.


brandur-stripe:
(Quite minor, so going to self-LGTM.)

PR #435, opened 2018-06-18
setup.py: add testing extra

blueyed:
This allows for easily `pip install -e ".[testing]"` to install test
dependencies in a new virtualenv.

For what it's worth: I also install pytest then still, but it is not the
official test runner, so shouldn't be added there.


brandur-stripe:
Cool, this seems useful. Thanks!

PR #436, opened 2018-06-27
Add support for the PaymentIntent resource

remi-stripe:
This is gated so for now the tests are stubbed.


ob-stripe:
Looks like you need to add your new resource class to `stripe/util.py`.

ob-stripe:
Released in 1.84.0.

PR #437, opened 2018-06-28
Adds 'partner_id' property to set_app_info hash

zachwick:
This adds support for an optional partner_id attribute in the SetAppInfo hash.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -79,10 +79,11 @@ from stripe.util import (  # noqa
 # communicating with Stripe.
 #
 # Takes a name and optional version and plugin URL.
-def set_app_info(name, version=None, url=None):
+def set_app_info(name, partner_id=None, url=None, version=None):
     global app_info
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released in 1.83.0.

ob-stripe:
I'm just realizing that if someone uses `set_app_info` with positional args instead of keyword args, then this is a breaking change :/

@brandur-stripe wdyt? Should we reorder the params?

zachwick:
@ob-stripe @brandur-stripe There was some discussion that the arguments should be in alphabetical order, but I'm open to instead adding `partner_id` as the final argument instead.

brandur-stripe:
> @brandur-stripe wdyt? Should we reorder the params?

Ah, good point. I'd probably just leave it as this point ... realistically the use of this feature in Python is probably still pretty minimal, and the use that we recommend in the README uses named parameters.

PR #439, opened 2018-07-04
Fix serialization of empty list objects

### LINKED ISSUE #438: stripe.Account.external_accounts deserializes into StripeObject instead of ListObject
Bug:
When getting a JSON representation of a `stripe.Account` object, the `external_accounts` `ListObject` is serialized as an empty JSON object. This is a problem when the JSON is later deserialized back into a `stripe.Account` object because the `external_accounts` JSON object is missing the `object` key-value that is required to deserialize `external_accounts` back into a `ListObject`. Instead, `external_accounts` is deserialized into a generic `StripeObject` that lacks critical capabilities, such as the ability to create a new external account for the deserialized  `stripe.Account`. Attempting to call `external_accounts.create`, for instance, will raise an `AttributeError: create` exception.

This bug was encountered when upgrading our Stripe lib version from `1.27.1` to `1.84.0`. It affects how we cache `stripe.Account` objects.

Expected:
When getting a JSON representation of a `stripe.Account` object, the `external_accounts` object should be serialized in such a way that it can be deserialized back into a `ListObject`.

FWIW, the serialized `external_accounts` object use to be serialized like the following back in lib version `1.27.1`:
```
    ...
    "external_accounts": {
        "data": [
        ],
        "has_more": false,
        "object": "list",
        "total_count": 0,
        "url": "/v1/accounts/<account id>/external_accounts"
    },
    ...
```
This was much better, and such an object seems to be properly deserialized back into a `ListObject` and behaviors such as `external_accounts.create(...)` behave as expected.

Repro:
```
import json
import stripe
stripe.api_key = <an api key>

account_id = <an account id>

account = stripe.Account.retrieve(account_id)
account_serialized = str(account)
account_deserialized = stripe.Account.construct_from(json.loads(account_serialized), stripe.api_key)

account.external_accounts  # <ListObject list at 0x1140ae618> JSON: {}
account_deserialized.external_accounts  # <StripeObject at 0x113eb6178> JSON: {}

account.external_accounts.create  # <bound method ListObject.create of <ListObject list at 0x1140ae618> JSON: {}>
account_deserialized.external_accounts.create  # AttributeError: create
```

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @sazlin

Fixes #438.

### DIFF
--- a/stripe/stripe_object.py
+++ b/stripe/stripe_object.py
@@ -40,6 +40,13 @@ class StripeObject(dict):
                 return api_requestor._encode_datetime(obj)
             return super(StripeObject.ReprJSONEncoder, self).default(obj)
 
+        def encode(self, obj):
+            # ListObject has a custom __len__ method, but we want to serialize
+            # its attributes even if __len__ returns 0.
+            if isinstance(obj, stripe.ListObject):
+                obj = dict(obj)
+            return super(StripeObject.ReprJSONEncoder, self).encode(obj)
+
     def __init__(self, id=None, api_key=None, stripe_version=None,
                  stripe_account=None, last_response=None, **params):
         super(StripeObject, self).__init__()
--- a/tests/api_resources/test_list_object.py
+++ b/tests/api_resources/test_list_object.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import, division, print_function
 
+import json
+
 import stripe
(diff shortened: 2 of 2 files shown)


brandur-stripe:
Thanks for getting to the bottom of this one OB, and thanks for adding a test!

LGTM.

PR #441, opened 2018-07-06
Fix serialization of nested `ListObject`s

### LINKED ISSUE #440: stripe.Account.external_accounts still deserializes into StripeObject instead of ListObject
Bug:
The recent fix for https://github.com/stripe/stripe-python/issues/438 does not fix the repro provided in that bug. See the same issue here:
```python
...
account = stripe.Account.retrieve(account_id)
account_serialized = str(account)
print account_serialized  # Note that the `external accounts` key has a value of `{}`. Expecting at least an 'object' key in there with value 'list'.
account_deserialized = stripe.Account.construct_from(json.loads(account_serialized), stripe.api_key)
account_deserialized.external_accounts  # <StripeObject at 0x113deb050> JSON: {}  <-- wrong type
account_deserialized.external_accounts.create  # raises AttributeError
```

Repros with Stripe lib version 1.84.1
Python version 2.7.14

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @sazlin

Another attempt at fixing the serialization of `ListObject`s, this time for good hopefully 🤞 

Before calling the JSON serializer, all `StripeObject`s are recursively converted into regular `dict`s, which should prevent any of our custom stuff from interfering with the serialization.

Fixes #440.

### DIFF
--- a/stripe/stripe_object.py
+++ b/stripe/stripe_object.py
@@ -40,13 +40,6 @@ class StripeObject(dict):
                 return api_requestor._encode_datetime(obj)
             return super(StripeObject.ReprJSONEncoder, self).default(obj)
 
-        def encode(self, obj):
-            # ListObject has a custom __len__ method, but we want to serialize
-            # its attributes even if __len__ returns 0.
-            if isinstance(obj, stripe.ListObject):
-                obj = dict(obj)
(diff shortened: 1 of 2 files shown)


brandur-stripe:
This seems like an elegant solution that's hopefully pretty safe — LGTM.

PR #445, opened 2018-07-10
Fix setup.py

### LINKED ISSUE #444: pypi stripe 2.0 package can not be imported
Installing stripe 2.0 using pip in python 3.5 gives an package that can not be imported:

```
pip install stripe
python -c 'import stripe'
ImportError: No module named 'stripe'
```

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fixes `setup.py` so that it can be run from any directory (which is what our release tooling does).

Fixes #444.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,6 @@
+import os
 import sys
 from codecs import open
-from os import path
 from setuptools import setup, find_packages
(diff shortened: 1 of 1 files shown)


brandur-stripe:
LGTM.

Issue #454, opened 2018-07-25
ScheduledQueryRun missing


jleclanche:
Could we get a class for ScheduledQueryRun please?

https://stripe.com/docs/api#scheduled_query_run_object

:)

ob-stripe:
Released as 2.3.0.

PR #456, opened 2018-07-26
Add support for ScheduledQueryRun

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries 

This fixes https://github.com/stripe/stripe-python/issues/454

### DIFF
--- a/stripe/api_resources/sigma/scheduled_query_run.py
+++ b/stripe/api_resources/sigma/scheduled_query_run.py
@@ -1,0 +1,11 @@
+from __future__ import absolute_import, division, print_function
+
+from stripe.api_resources.abstract import ListableAPIResource
+
+
+class ScheduledQueryRun(ListableAPIResource):
+    OBJECT_NAME = 'scheduled_query_run'
+
+    @classmethod
+    def class_url(cls):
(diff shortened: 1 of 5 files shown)


stripe/api_resources/sigma/scheduled_query_run.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import absolute_import, division, print_function
+
+from stripe.api_resources.abstract import UpdateableAPIResource
+from stripe.api_resources.abstract import ListableAPIResource
+
+
+class ScheduledQueryRun(ListableAPIResource):
+    OBJECT_NAME = 'scheduled_query_run'
```
remi-stripe: The object name should really be `sigma.scheduled_query_run` based on recent API decisions we made. It's not though, similarly to apple pay domains.

The issue is that if we change this in the future, we'll need to have some custom logic for the deserializer which will be tricky. I don't think we have renamed an object before.

This comment is mostly here as food for thoughts though!

ob-stripe:
Released as 2.3.0.

jleclanche:
Thanks!

PR #459, opened 2018-08-02
Update stripe-mock version to 0.25.0

foresthu-stripe:
I would like to upgrade the stripe-mock version to add the new `cancel` topup endpoint

r? @stripe/api-libraries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ cache:
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


alexander-stripe:
LGTM

brandur-stripe:
LGTM as well.

(Just for future reference, it's usually fine to just do the upgrade in the same PR that you're adding your new feature.)

PR #462, opened 2018-08-09
Let's deprecate easy_install from the docs.

martinvol:


brandur-stripe:
Can you please provide more context/rationale?

martinvol:
`pip` is now the de facto Python package manager tool and it's evolving
fast with the Pipfile and Pipenv. easy_install is fading out and having it
in the docs may confuse newcomers.

brandur-stripe:
Okay, thanks. It does seem to be all but totally superseded by now.

PR #468, opened 2018-08-30
Upgrade stripe-mock to 0.30.0

brandur:
Let's make sure that everything is still working on the latest version
of stripe-mock.

stripe-mock 0.30.0 brought in additional validation of query parameters,
so it caught a few places where the Python test suite wasn't making
valid requests. I've fixed up a few tests around invoices and
subscription items to add parameters that should have been there.

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,8 @@ cache:
 
 env:
   global:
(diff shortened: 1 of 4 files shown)


brandur-stripe:
Thanks OB!

PR #469, opened 2018-08-31
Add support for the Reporting resources

remi-stripe:
Adding support for the new Reporting resources. Those are in OpenAPI but not yet public but since they are starting to be used and are stable we should merge them.

It's WIP for now while I get some feedback on the implementation and the shape of the resources.

cc @stripe/api-libraries
cc @brahn-stripe

### DIFF
--- a/stripe/api_resources/reporting/report_type.py
+++ b/stripe/api_resources/reporting/report_type.py
@@ -1,0 +1,7 @@
+from __future__ import absolute_import, division, print_function
+
+from stripe.api_resources.abstract import ListableAPIResource
+
+
+class ReportType(ListableAPIResource):
+    OBJECT_NAME = 'reporting.report_type'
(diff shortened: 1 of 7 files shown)


stripe/api_resources/reporting/report_type.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import absolute_import, division, print_function
+
+from stripe.api_resources.abstract import CreateableAPIResource
+from stripe.api_resources.abstract import ListableAPIResource
+
+
+class ReportType(CreateableAPIResource, ListableAPIResource):
```
ob-stripe: This resource is not creatable.

remi-stripe:
Fixed PTAL
r? @ob-stripe

ob-stripe:
Released as 2.7.0.

Issue #464, opened 2018-08-16
Retry mechanism for failing requests


jonathan-s:
Whether stripe times out, there are other network issues or general randomness of the interwebz it's good to be able to retry. 

So I would suggest that you implement a retry mechanism as described here: https://gist.github.com/benjiao/28dc36bd87121b3273e0b3e079a8e8d8

Make it an optional parameter for when initializing the client with a sane default. That way the user of the library can customize it if they deem it necessary.

brandur-stripe:
Yeah, it's a good idea to have something like this. I'm tagging this issue a "future" for now because I don't think we'll have anyone actively working on it right away, but it's open for implementation.

PR #470, opened 2018-09-07
Python retry mechanism

mickjermsurawong-stripe:
Summary:
- Implement retry mechanism at abstract HTTPClient class, and use delegated request call to each client implementation.   
- Only consider errors from `requests` http-client, but not others due to overwhelming majority of our clients. This design allows future extension the support to other http-clients. (As opposed to urllib3 `Retry` that is coupled at the execution invocation.
- Retry conditions follow other language implementations (Ruby[0]/Go[1]) 
  - Retry on 409 resource conflict (both)
  - Retry on timeout/connection (only Ruby) 
- Exponential backoff delay follows other language implementations (Ruby[2]/Go[3])
- Added idempotency key only for POST method, and when it is not given in the header

Concerns:
- Not sure where's best to put delay constants in config. Currently made as static class var
- I'm not sure how to access the actual `requests` module from within test, and that results in mocking kinds of exception classes that we want to handle (existing tests also rely on the similar mechanism)

Motivation:
- Issue https://github.com/stripe/stripe-python/issues/464 and from conversation in PR https://github.com/stripe/stripe-python/pull/466 

[0] Ruby retry conditions
https://github.com/stripe/stripe-ruby/blob/ec66c3f0f44274f885de8d13de5dce2657932121/lib/stripe/stripe_client.rb#L61
[1] Go retry conditions
https://github.com/stripe/stripe-go/blob/442690c3856640e4913163f506f0070561f3368d/stripe.go#L495
[2] Ruby sleep
https://github.com/stripe/stripe-ruby/blob/ec66c3f0f44274f885de8d13de5dce2657932121/lib/stripe/stripe_client.rb#L80
[3] Go sleep
https://github.com/stripe/stripe-go/blob/442690c3856640e4913163f506f0070561f3368d/stripe.go#L511

### DIFF
--- a/stripe/error.py
+++ b/stripe/error.py
@@ -53,7 +53,11 @@ class APIError(StripeError):
 
 
 class APIConnectionError(StripeError):
-    pass
+    def __init__(self, message, http_body=None, http_status=None,
+                 json_body=None, headers=None, code=None, should_retry=False):
+        super(APIConnectionError, self).__init__(message, http_body, http_status,
+                                                 json_body, headers, code)
+        self.should_retry = should_retry
 
 
 class StripeErrorWithParamCode(StripeError):
--- a/stripe/error.py
+++ b/stripe/error.py
@@ -55,7 +55,8 @@ class APIError(StripeError):
 class APIConnectionError(StripeError):
     def __init__(self, message, http_body=None, http_status=None,
                  json_body=None, headers=None, code=None, should_retry=False):
-        super(APIConnectionError, self).__init__(message, http_body, http_status,
+        super(APIConnectionError, self).__init__(message, http_body,
+                                                 http_status,
                                                  json_body, headers, code)
         self.should_retry = should_retry
 
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -206,7 +206,8 @@ class RequestsClient(HTTPClient):
 
     def _handle_request_error(self, e):
 
-        # Catch SSL error first as it belongs to ConnectionError
+        # Catch SSL error first as it belongs to ConnectionError,
+        # but we don't want to retry
         if isinstance(e, requests.exceptions.SSLError):
             msg = ("Could not verify Stripe's SSL certificate.  Please make "
                    "sure that your network is not intercepting certificates.  "
@@ -222,7 +223,7 @@ class RequestsClient(HTTPClient):
                    "support@stripe.com.")
             err = "%s: %s" % (type(e).__name__, str(e))
             should_retry = True
-        # Catch all request specific with descriptive class/messages
+        # Catch remaining request exceptions
         elif isinstance(e, requests.exceptions.RequestException):
             msg = ("Unexpected error communicating with Stripe.  "
                    "If this problem persists, let us know at "
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -107,7 +107,15 @@ class HTTPClient(object):
                 response = None
 
             if self._should_retry(response, connection_error, num_retries):
-                self._sleep(num_retries)
+                if connection_error:
+                    util.log_info("Encountered a retryable error %s" %
+                                  connection_error.user_message)
+
+                sleep_time = self._sleep_time_seconds(num_retries)
+                util.log_info(("Initiating retry %i for request %s %s after "
+                               "sleeping %.2f seconds." %
+                               (num_retries, method, url, sleep_time)))
+                time.sleep(sleep_time)
             else:
                 if response is not None:
                     return response
@@ -123,6 +131,10 @@ class HTTPClient(object):
             _, status_code, _ = response
             should_retry = status_code == 409
         else:
+            # We generally want to retry on timeout and connection
+            # exceptions, but defer this decision to underlying subclass
+            # implementations. They should evaluate the driver-specific
+            # errors worthy of retries, and set flag on the error returned.
             should_retry = api_connection_error.should_retry
         return should_retry and num_retries < self._max_network_retries()
 
@@ -130,10 +142,7 @@ class HTTPClient(object):
         # Configured retries, isolated here for tests
         return max_network_retries
 
-    def _sleep(self, num_retries):
-        time.sleep(self._sleep_time(num_retries))
-
-    def _sleep_time(self, num_retries):
+    def _sleep_time_seconds(self, num_retries):
         # Apply exponential backoff with initial_network_retry_delay on the
         # number of num_retries so far as inputs.
         # Do not allow the number to exceed max_network_retry_delay.
@@ -145,7 +154,6 @@ class HTTPClient(object):
 
         # But never sleep less than the base sleep seconds.
         sleep_seconds = max(HTTPClient.INITIAL_DELAY, sleep_seconds)
-
         return sleep_seconds
 
     def _add_jitter_time(self, sleep_seconds):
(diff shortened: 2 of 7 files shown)


mickjermsurawong-stripe:
cc @jonathan-s

mickjermsurawong-stripe:
cc @stripe/api-libraries

stripe/error.py
```diff
@@ -53,7 +53,12 @@ class APIError(StripeError):
 
 
 class APIConnectionError(StripeError):
-    pass
+    def __init__(self, message, http_body=None, http_status=None,
+                 json_body=None, headers=None, code=None, should_retry=False):
+        super(APIConnectionError, self).__init__(message, http_body,
+                                                 http_status,
+                                                 json_body, headers, code)
+        self.should_retry = should_retry
```
mickjermsurawong-stripe: How children implementation can signal the retry to the generic parent HTTPClient, since all client-specific exceptions are now wrapped into this

stripe/http_client.py
```diff
@@ -77,6 +79,10 @@ def new_default_http_client(*args, **kwargs):
 
 
 class HTTPClient(object):
+    MAX_RETRIES = 3
+    MAX_DELAY = 2
+    INITIAL_DELAY = 0.5
```
mickjermsurawong-stripe: Not so sure what's the best way to manage these constants here

ob-stripe: I think we should be consistent with stripe-ruby: the maximum number of retries should be configurable (and default to 0), so it should be a top-level variable in the `stripe` module.

The other two parameters are read-only in stripe-ruby, so I think it's fine to keep them as class constants here.

stripe/http_client.py
```diff
@@ -90,9 +96,62 @@ def __init__(self, verify_ssl_certs=True, proxy=None):
         self._proxy = proxy.copy() if proxy else None
 
     def request(self, method, url, headers, post_data=None):
```
ob-stripe: Do you think we could keep `request` as it is now, name the new wrapper method `request_with_retries`, and change `api_requestor.request_raw` to call `_client.request_with_retries` instead of `_client.request`?

stripe/http_client.py
```diff
@@ -90,9 +96,62 @@ def __init__(self, verify_ssl_certs=True, proxy=None):
...
+        sleep_seconds = self._add_jitter_time(sleep_seconds)
+
+        # But never sleep less than the base sleep seconds.
+        sleep_seconds = max(HTTPClient.INITIAL_DELAY, sleep_seconds)
+
+        return sleep_seconds
+
+    def _add_jitter_time(self, sleep_seconds):
+        # Randomize the value in [(sleep_seconds/ 2) to (sleep_seconds)]
+        # Also separated method here to isolate randomness for tests
```
ob-stripe: 👍 Great idea making this a separate method.

stripe/http_client.py
mickjermsurawong-stripe: Yup! more explicit name sounds better to me

ob-stripe:
Alright, this looks great to me, but the feature is complex enough that I'd like to get another set of eyes on it.

r? @brandur-stripe

stripe/http_client.py
```diff
@@ -89,10 +94,66 @@ def __init__(self, verify_ssl_certs=True, proxy=None):
...
+                connection_error = e
+                response = None
+
+            if self._should_retry(response, connection_error, num_retries):
+                self._sleep(num_retries)
+            else:
+                if response is not None:
+                    return response
+                else:
+                    raise connection_error
```
brandur-stripe: Can we add a little logging here so that it's possible to get some visibility into what's happening for users that care about it?

Use stripe-go for inspiration, but I'd suggest at the least:

1. Emit an `error` line for the error that was received (although possibly only if we're going to retry if this would result in a duplicate line from elsewhere).
2. An `info` line something like this that announces its intention to sleep (from Go):

``` go
			s.Logger.Printf("Initiating retry %v for request %v %v%v after sleeping %v\n",
				retry, req.Method, req.URL.Host, req.URL.Path, sleepDuration)
```

If you check `util`, we have methods like `log_info`, `log_error` already built in.

stripe/http_client.py
```diff
@@ -89,10 +94,66 @@ def __init__(self, verify_ssl_certs=True, proxy=None):
...
     def request(self, method, url, headers, post_data=None):
         raise NotImplementedError(
             'HTTPClient subclasses must implement `request`')
 
+    def _should_retry(self, response, api_connection_error, num_retries):
+        if response is not None:
+            _, status_code, _ = response
+            should_retry = status_code == 409
+        else:
+            should_retry = api_connection_error.should_retry
```
brandur-stripe: Can we add a comment here that's something to the effect of:

> Specific error classes that we might want to retry on it (timeout, connection reset, etc.) will depend on the driver that's in use, so we defer to those drives to tell us which errors are retry worthy by setting the `should_retry` flag on the error they return.

tests/test_http_client.py
```diff
@@ -52,6 +52,107 @@ def test_new_default_http_client_urllib2(self):
...
+    def test_should_retry_on_codes(self):
+        one_xx = list(range(100, 104))
+        two_xx = list(range(200, 209))
+        three_xx = list(range(300, 308))
+        four_xx = list(range(400, 431))
+        five_xx = list(range(500, 512))
+
+        client = stripe.http_client.new_default_http_client()
+        codes = one_xx + two_xx + three_xx + four_xx + five_xx
+        codes.remove(409)
```
brandur-stripe: Haha, we probably don't need to be _this_ exhaustive on the tests, but definitely applaud your attention to detail ;)

(No action required — it's fine given it's done already.)

stripe/http_client.py
mickjermsurawong-stripe: @brandur-stripe, `log_error` doesn't exist[0]-- are there particular reasons we don't implement so? Otherwise, I can add a method to it. 
If so should it look something like this; I'm concerned about the `_console_log_level()` line
```
def log_error(message, **params):
    msg = logfmt(dict(message=message, **params))
    if _console_log_level() in ['debug', 'info', 'error']: # added here?
        print(msg, file=sys.stderr)
    logger.error(msg) #[1] built-in python logger?
# existing method
def log_info(message, **params):
    msg = logfmt(dict(message=message, **params))
    if _console_log_level() in ['debug', 'info']:
        print(msg, file=sys.stderr)
    logger.info(msg)
```
[0] https://github.com/stripe/stripe-python/blob/master/stripe/util.py#L58
[1] https://docs.python.org/2/library/logging.html#logging.Logger.error

brandur-stripe: Oh, weird! Yeah, we should probably add that, but I wouldn't worry about it for now (we'll do it in another PR) — just use `log_info`.

mickjermsurawong-stripe:
Added comment/logging
ptal @brandur-stripe

brandur-stripe:
LGTM! Thanks.

mickjermsurawong-stripe:
self-approve after lgtm ^

mickjermsurawong-stripe:
Released in 2.8 fb8117d75cd20043701bcd0fedbb2f300132aa6a

kyleries:
Ended up needing this feature and came across this thread to check out the implementation. Wanted to drop a belated "thank you" for this PR!

PR #474, opened 2018-09-13
Don't URL-encode square brackets

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Don't URL-encode square brackets. Most of our other libraries already do this.


brandur-stripe:
Good call. LGTM!

PR #476, opened 2018-09-17
Use stripe-mock for FileUpload tests

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Updates the `FileUpload` tests to use stripe-mock instead of stubbing requests.

### DIFF
--- a/stripe/multipart_data_generator.py
+++ b/stripe/multipart_data_generator.py
@@ -21,7 +21,12 @@ class MultipartDataGenerator(object):
             self._write(self.param_header())
             self._write(self.line_break)
             if hasattr(value, 'read'):
-                filename = value.name if hasattr(value, 'name') else "blob"
+                filename = "blob"
+                if hasattr(value, 'name'):
+                    # Convert the filename to string, just in case it's not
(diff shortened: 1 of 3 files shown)


stripe/multipart_data_generator.py
```diff
@@ -21,7 +21,9 @@ def add_params(self, params):
             self._write(self.param_header())
             self._write(self.line_break)
             if hasattr(value, 'read'):
-                filename = value.name if hasattr(value, 'name') else "blob"
+                filename = "blob"
+                if hasattr(value, 'name'):
+                    filename = str(value.name)
```
ob-stripe: This fixes an edge case when using a `tempfile.TemporaryFile` instance: the `name` attribute exists but is an `int`, which caused an exception when trying to insert the filename into the buffer.

brandur-stripe: Mind adding a short comment that says what you said here and what conditions we expect this to occur under?

brandur-stripe:
One minor nit, but LGTM.

ob-stripe:
Added a comment as requested. Self-approving.

PR #477, opened 2018-09-17
Handle `file` objects like `file_upload`

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Exactly like https://github.com/stripe/stripe-ruby/pull/689, but for Python.


brandur-stripe:
LGTM!

ob-stripe:
Pushed another commit to only use `files.stripe.com` for file creation requests. ptal @brandur-stripe

tests/api_resources/test_file.py
```diff
@@ -2,13 +2,23 @@
...
 
 
 class TestFile(object):
+    @pytest.fixture
+    def setup_upload_api_base(self):
+        stripe.upload_api_base = stripe.api_base
+        stripe.api_base = None
+        yield
+        stripe.api_base = stripe.upload_api_base
+        stripe.upload_api_base = 'https://files.stripe.com'
```
brandur-stripe: I'm not too familiar with the framework here, but are you sure this doesn't need to be in a `finally` to guarantee that these values are set back in case of an exception?

brandur-stripe:
Left a minor comment above, but otherwise LGTM!

tests/api_resources/test_file.py
ob-stripe: Just tested it, and it's not necessary. The framework already catches exceptions in tests and reports them as errors, and still executed the post-`yield` code.

PR #480, opened 2018-10-01
feature: support deletion without a stripe object

r1b:
This is roughly the same as `modify` on `UpdatableAPIResource`.

### DIFF
--- a/stripe/api_resources/abstract/deletable_api_resource.py
+++ b/stripe/api_resources/abstract/deletable_api_resource.py
@@ -1,9 +1,26 @@
 from __future__ import absolute_import, division, print_function
 
+from stripe import api_requestor, util
 from stripe.api_resources.abstract.api_resource import APIResource
+from stripe.six.moves.urllib.parse import quote_plus
 
 
 class DeletableAPIResource(APIResource):
+
+    @classmethod
+    def _remove(cls, url, api_key=None, stripe_version=None,
+                stripe_account=None, **params):
+        requestor = api_requestor.APIRequestor(api_key,
+                                               api_version=stripe_version,
+                                               account=stripe_account)
+        response, api_key = requestor.request('delete', url, params)
+        return util.convert_to_stripe_object(response, api_key, stripe_version,
+                                             stripe_account)
+
+    @classmethod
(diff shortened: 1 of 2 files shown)


r1b:
My colleague points out that you can just do `stripe.Subscription.delete(stripe.Subscription(id='foo'))`.

ob-stripe:
Hi @r1b, thanks for the contribution!

I think this would be a great addition to the library. My only uncertainty is about the naming -- `save` vs `update` is already not great in terms of making it clear that one is an instance method and one is a class method, and I fear `delete` vs `remove` is in the same vein.

@brandur-stripe @remi-stripe What do you think?

brandur-stripe:
> @brandur-stripe @remi-stripe What do you think?

I feel exactly the same as you — we should have this feature, but not crazy about the `delete` versus `remove` distinction.

> My colleague points out that you can just do 
> `stripe.Subscription.delete(stripe.Subscription(id='foo'))`.

Hah, this almost works well enough just to act as a substitute. I think the only trouble is that it'll need a `self` which has to be a `Subscription` object. That might be okay though.

ob-stripe:
This should work today:

```python
stripe.Subscription(id='sub_123').delete()
```

but it's pretty hacky to have to manually create an instance just to call `delete()` on it.

I'm still not sure about the naming... Since `update()` is the class method, `delete()` should also be a class method if we wanted to stick to the CRUD terminology. But that would be a breaking change since I don't think it's possible to have a class method and an instance method with the same name in Python.

remi-stripe:
We could just name it `delete_without_fetch` or something similar. Not perfectly elegant but it'd be fine until a future major version

remi-stripe:
@r1b We ended up taking another approach in https://github.com/stripe/stripe-python/pull/543 which we're going to release soon so I'm going to close this PR! Thanks again for the work here and pushing us to make this better in the library

ob-stripe:
@r1b We just released version 2.24.0 of the library, which allows you to call `delete` from a class context:
```python
stripe.Subscription.delete('sub_123')
```

Sorry it took us so long to ship this! Feel free to reply if you have comments or run into any issues.

PR #489, opened 2018-10-31
Add support for Radar List and ListItem resources

remi-stripe:
cc @stripe/api-libraries


ob-stripe:
PyPy3 builds are red, but it's yet another pipenv issue :/ Feeling pretty confident that our own code is fine, so going to pull this in.

PR #491, opened 2018-11-05
Bump minimum requests version to 2.20.0

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

GitHub reports a security alert on the repo because the version of requests currently mentioned in `Pipfile.lock` is 2.19.1, which is subject to this vulnerability: https://nvd.nist.gov/vuln/detail/CVE-2018-18074.

`Pipfile.lock` is only used in development and has no impact on library users, but we might as well bump the minimum version of requests to 2.20.0 to ensure users are not affected by this vulnerability.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,8 +37,7 @@ before_install:
     STRIPE_MOCK_PID=$!
 
 install:
-  - pip install --upgrade pipenv
-  - pipenv install --dev
+  - make init
 
 script:
   - if [[ $TRAVIS_PYTHON_VERSION == '2.7' ]]; then make lint; fi
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,7 @@
+init:
+	pip install pipenv --upgrade
+	pipenv install --dev --skip-lock
+
 test:
 	pipenv run detox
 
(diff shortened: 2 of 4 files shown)


ob-stripe:
Hrm, looks like one of the upgraded packages broke compatibility with 2.7 and 3.4.

brandur-stripe:
Oof, what a pain. Did you happen to take a look to see if maybe the Request people are going to backport the fix?

The only good news here is that from my reading of the CVE, I think we're relatively safe — we only target our own hostname and I don't think we ever do an https-to-http redirect on the backend anywhere.

ob-stripe:
The compatibility issue wasn't caused by requests itself, but by another updated package (pipenv will only update all packages at once). It was caused by pytest, which requires the `pathlib2` module on older Python versions (https://github.com/pytest-dev/pytest/blob/ccdb248397fb54325675c5a359de6804d27f9b4d/setup.py#L16).

The module wasn't installed because our `Pipfile.lock` is created with Python 3.7. I've updated the CI command to add the `--skip-lock` flag to avoid this and ensure that the correct dependencies are installed for each Python version. This is also how the requests package handles this (https://github.com/requests/requests/blob/943a5c8e89db1758ae24adbbedacb3b05c32df4a/Makefile#L4).

ptal @brandur-stripe

brandur-stripe:
Nice investigation! LGTM.

PR #495, opened 2018-11-21
Make StripeResponse a new-style class

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Make `StripeResponse` a [new-style class](https://docs.python.org/2/glossary.html#term-new-style-class). This ensures that the `@property` decorators work as expected with Python 2.7. (Python 3 doesn't have old-style classes at all so isn't affected.)

### DIFF
--- a/stripe/stripe_response.py
+++ b/stripe/stripe_response.py
@@ -3,8 +3,7 @@ from __future__ import absolute_import, division, print_function
 import json
 
 
-class StripeResponse:
-
+class StripeResponse(object):
     def __init__(self, body, code, headers):
         self.body = body
         self.code = code


remi-stripe:
@ob-stripe Can you assign someone else? I don't really understand that change and the side effects and since tests are broken on top of it (though likely unrelated) and the entire pipfile.lock file changed, I'd rather not blindly approve

ob-stripe:
For context:
- failing tests are unrelated -- the tests fail on master too. I think it's due to some weird interaction between the latest version of pip or pipenv and Travis' virtualenv
- updating `Pipfile.lock` was just a failed attempt at making tests pass
- there is basically no reason to use old-style classes, ever
- I tried looking for a way to enforce this via the linter, but it doesn't seem to be possible without writing a custom extension. I assume this is because old-style classes are gone in Python 3.

r? @brandur-stripe

brandur-stripe:
LGTM.

A little [extra information on new and old style classes here](https://docs.python.org/2/reference/datamodel.html#newstyle). My interpretation as well that it's just strictly better to use new-style at this point.

Too bad about that build. I didn't look into it, but LMK if you want a second pair of eyes on it or anything.

ob-stripe:
Closing the loop re. Travis builds: I've confirmed this is an issue with the latest pipenv. They've already fixed the issue so our builds should start working again as soon as the new version is released.

PR #500, opened 2018-11-27
Add support for the Review resource

remi-stripe:
cc @stripe/api-libraries


remi-stripe:
The build error is specific to a bug in pipenv that @ob-stripe is working on with them

PR #503, opened 2018-11-30
 Add support for providing custom CA cert bundle

### LINKED ISSUE #501: Support custom certificate bundle
Allow users to pass their own certificate bundle or extra certificate, for example when using a proxy.

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Add support for providing a custom CA bundle, by setting `stripe.ca_bundle_path`.

Fixes #501.

It's kind of difficult to write automated tests for this, but I've manually tested that this works with both `RequestsClient` and `PycurlClient`.

While I was at it, I also updated the embedded bundle to the latest version from https://curl.haxx.se/docs/caextract.html.


brandur-stripe:
Nice! LGTM. (And thanks for following the same naming convention as Stripe Ruby for `ca_bundle_path`.)

It looks like the PyPi build is having trouble, and I don't understand it exactly.:

``` sh
  File "/home/travis/virtualenv/pypy3.5-5.8.0/site-packages/pipenv/_compat.py", line 66, in force_encoding
    from ctypes import pythonapi, py_object, c_char_p
ImportError: cannot import name 'pythonapi'
make: *** [init] Error 1
```

I restarted it, but it doesn't appear to be a transient problem because it failed again. Mind taking a look?

ptal @ob-stripe

ob-stripe:
@brandur-stripe It's a problem with pipenv itself. I submitted a fix (https://github.com/pypa/pipenv/pull/3322) but they haven't released a new version yet.

ob-stripe:
Released as 2.15.0.

brandur-stripe:
Ah, I see! And wow — nice work on that pipenv patch.

PR #518, opened 2019-01-10
Add Stripe client telemetry to request headers

jameshageman-stripe:
Follows https://github.com/stripe/stripe-ruby/pull/696 and https://github.com/stripe/stripe-php/pull/549 in adding telemetry metadata to request headers. 

The telemetry is disabled by default, and can be enabled like so:
```python
stripe.enable_telemetry = True
```

### DIFF
--- a/stripe/api_requestor.py
+++ b/stripe/api_requestor.py
@@ -11,6 +11,7 @@ import stripe
 from stripe import error, oauth_error, http_client, version, util, six
 from stripe.multipart_data_generator import MultipartDataGenerator
 from stripe.six.moves.urllib.parse import urlencode, urlsplit, urlunsplit
+from stripe.request_metrics import RequestMetrics
 from stripe.stripe_response import StripeResponse
 
 
@@ -28,6 +29,10 @@ def _encode_nested_dict(key, data, fmt='%s[%s]'):
     for subkey, subvalue in six.iteritems(data):
         d[fmt % (key, subkey)] = subvalue
     return d
+
+
+def _now_ms():
+    return int(round(time.time() * 1000))
 
 
 def _api_encode(data):
@@ -79,6 +84,8 @@ class APIRequestor(object):
         self._client = client or stripe.default_http_client or \
             http_client.new_default_http_client(
                 verify_ssl_certs=verify, proxy=proxy)
+
+        self._last_request_metrics = None
 
     @classmethod
     def format_app_info(cls, info):
@@ -226,6 +233,11 @@ class APIRequestor(object):
         if self.api_version is not None:
             headers['Stripe-Version'] = self.api_version
 
+        if stripe.enable_telemetry and self._last_request_metrics:
+            headers['X-Stripe-Client-Telemetry'] = json.dumps({
+                'last_request_metrics': self._last_request_metrics.payload()
+            })
+
         return headers
 
     def request_raw(self, method, url, params=None, supplied_headers=None):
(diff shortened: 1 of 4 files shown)


remi-stripe:
cc @stripe/api-libraries for awareness

jameshageman-stripe:
Pinging @akropp-stripe @ob-stripe @brandur-stripe for review as you were all part of https://github.com/stripe/stripe-ruby/pull/696.

The 2 failing tests are also failing on master and are unrelated to my changes.

stripe/api_requestor.py
```diff
@@ -12,6 +12,7 @@
 from stripe.multipart_data_generator import MultipartDataGenerator
 from stripe.six.moves.urllib.parse import urlencode, urlsplit, urlunsplit
 from stripe.stripe_response import StripeResponse
+from stripe.request_metrics import RequestMetrics
```
brandur-stripe: Can you try to arrange this as alphabetically as possible (since it's already roughly in that order)? Just swap it with the line above.

stripe/request_metrics.py
```diff
@@ -0,0 +1,10 @@
+class RequestMetrics(object):
```
brandur-stripe: Minor note, but the one in Ruby is called `StripeRequestMetrics`, and it'd be nice to standardize on a name. `RequestMetrics` is probably better given it's already in the Stripe namespace.

brandur-stripe:
@ob-stripe I looked into the test failures that are happening on `master` a bit. They're caused because our `TestPycurlClientHttpProxy` and `TestPycurlClientHttpsProxy` classes inherit from `TestPycurlClient` and are calling that parent's `check_call` implementation (see the last line below):

``` python
class TestPycurlClientHttpProxy(TestPycurlClient):
     ...

    @pytest.fixture
    def check_call(self, request_mocks, curl_mock):
        def check_call(mock, meth, url, post_data, headers):
            ...

            super(TestPycurlClientHttpProxy, self).check_call(request_mocks)(
                mock, meth, url, post_data, headers)
```

Because all our `check_call`s are marked as `@pytest.fixture`, Python doesn't expect any explicit invocations, and therefore this error is triggered (I assume that something must have changed in some library because this seemed to be fine before).

I could fix this by just moving `TestPycurlClient`'s `check_call` body into a new method (`check_call_inner`?) and just having `check_call` as well as `TestPycurlClientHttpProxy` and `TestPycurlClientHttpsProxy` call that, but I'm not sure that's a "clean" solution. Can you think of a better way?

stripe/api_requestor.py
jameshageman-stripe: Done!

stripe/request_metrics.py
jameshageman-stripe: I'm fine either way, I just figured `from stripe.stripe_request_metrics import StripeRequestMetrics` was overly verbose.

brandur-stripe: Yeah, the `Stripe` prefix seems pretty redundant. Let's stick with what you have.

ob-stripe:
@jameshageman-stripe Can you rebase on the latest master? I fixed the unrelated test error.

@brandur-stripe Oops, sorry, I'm only seeing your comment now! FWIW I initially went with your suggestion, but changed my mind to make the new method slightly more consistent with what we already do in `TestRequestsClient.check_call`.

brandur-stripe:
> @brandur-stripe Oops, sorry, I'm only seeing your comment now! FWIW I initially went with your suggestion, but changed my mind to make the new method slightly more consistent with what we already do in `TestRequestsClient.check_call`.

NP. Your fix ended up being better than what I was imagining.

jameshageman-stripe:
@ob-stripe Rebased: all tests are passing except `pypy3`, which is failing with `ImportError: cannot import name 'pythonapi'`.

stripe/request_metrics.py
ob-stripe: Could you add `from __future__ import absolute_import, division, print_function` at the beginning of the file? Even though that line won't do anything here, it's present in all source files to prevent Python 2/3 compatibility issues.

ob-stripe:
Requested one minor change, but otherwise LGTM!

> @ob-stripe Rebased: all tests are passing except pypy3, which is failing with ImportError: cannot import name 'pythonapi'. This appears to be happening on master as well, so I'd say it is unrelated to this PR's changes.

Yes, this is an issue with `pipenv` that will be fixed in the next version. Sorry, should have let you know that was expected!

jameshageman-stripe:
@ob-stripe Done!

ob-stripe:
@jameshageman-stripe Thanks! Mind squashing the commits before I merge and cut a release?

ob-stripe:
Released as 2.18.0.

PR #519, opened 2019-01-14
Fix failing tests

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Two tests are failing because the way they create fixtures is no longer supported with the latest version of pytest. This PR fixes the failing tests.

(To be honest, the fixture setup in `test_http_client` is way too complex. I'm hoping to greatly simplify it at some point, but probably won't get around to it until we drop support for non-requests clients.)


ob-stripe:
Thanks Brandur!

PR #521, opened 2019-01-15
Use Black for formatting code

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

[Black](https://github.com/ambv/black) is a code formatter that is quickly gaining traction in the Python community. (Basically `go fmt` for Python.)

This PR introduces support for automatically formatting stripe-python's code using black. It adds two new make targets:
- `fmt` is used to automatically format code
- `fmtcheck` is used to check that the code is formatted as black expects. This is used in CI, to ensure that code is compliant before merging.

Tagged as WIP to get your opinion, but the PR is mostly ready -- the only thing's missing is updating `README.md` with instructions.


.flake8
```diff
@@ -1,3 +1,4 @@
 [flake8]
 exclude =
     six.py
+ignore = E501, W503
```
ob-stripe: `E501` is the max line length error. Black makes a best effort to respect the maximum line length, but sometimes goes over if other rules require it. Black's own [documentation](https://github.com/ambv/black#line-length) recommends ignoring `E501`.

`W503` is a non-PEP8 compliant warning that is ignored by default, but must be explicitly ignored when an explicit `ignore` configuration line is provided.

pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[tool.black]
+line-length=79
```
ob-stripe: Black defaults to 88 characters per line. I've changed it back to 79, which is the PEP8 recommended value. As explained above, Black might still go over the maximum length in some cases.

.flake8
brandur-stripe: Could we add a comment over this with the content of your description here (these alphanumerical designations are pretty obtuse).

brandur-stripe:
Wow, cool. I didn't go through the whole diff, but scanned about half of it, and all the formatting changes that Black makes seem pretty sane to me.

I love formatters. +1 from my perspective.

.flake8
ob-stripe: 👍 Done.

ob-stripe:
Pulling this in!

PR #522, opened 2019-01-15
Detect custom OpenAPI spec and fixtures and start stripe-mock from test suite

ob-stripe:
cc @stripe/api-libraries 

Like https://github.com/stripe/stripe-ruby/pull/715, for Python.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ test:
 	pipenv run tox -p auto
 
 ci:
-	pipenv run pytest --cov=stripe -n 8
+	pipenv run pytest --cov=stripe
 
 coveralls:
 	pipenv run coveralls
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import atexit
 import os
 import sys
 from distutils.version import StrictVersion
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -6,7 +6,7 @@ test:
 	pipenv run tox -p auto
 
 ci:
-	pipenv run pytest --cov=stripe -n 8
+	pipenv run pytest --cov=stripe
```
ob-stripe: I had to disable parallel test execution on Travis, otherwise it'd run into errors when starting a stripe-mock process and running tests in parallels. It's not a big deal since tests are still very fast (~10 seconds with coverage).

brandur-stripe: Doh! I assume you looked into it, but as long as you could get just the stripe mock startup process to run *non*-parallel, stripe-mock can take the parallel traffic just fine.

(And just thinking about it a little more ... this will probably be an issue for stripe-go's implementation as well.)

tests/conftest.py
```diff
@@ -11,10 +12,19 @@
 from stripe.six.moves.urllib.error import HTTPError
 
 from tests.request_mock import RequestMock
+from tests.stripe_mock import StripeMock
 
 
 MOCK_MINIMUM_VERSION = '0.40.0'
-MOCK_PORT = os.environ.get('STRIPE_MOCK_PORT', 12111)
+if StripeMock.start():
```
brandur-stripe: Should've noticed this in stripe-ruby too probably, but it might be worthwhile to have a succinct informational comment here to the tune of:

> Starts stripe-mock if an OpenAPI spec override is found in `openapi/`, and otherwise falls back to `STRIPE_MOCK_PORT` or 12111.

brandur-stripe:
LGTM.

Makefile
ob-stripe: To be clear, I don't think the issue had to do with stripe-mock itself but rather with resource exhaustion on Travis. The errors were inconsistent, but I saw a few "Cannot allocate memory" messages. I didn't try it, but maybe reducing the number of workers would have worked.

PR #526, opened 2019-01-25
Try to reuse the HttpClient between requests by default

jameshageman-stripe:
Previously, if the user has not set `stripe.default_http_client`, a new client would be created for every request. This commit changes the `APIRequestor` to populate `stripe.default_http_client` if it had not been set by the user, so the client can be reused between requests.

r? @brandur-stripe @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,8 @@ production use, but `debug` is also available for more verbosity.
 
 There are a few options for enabling it:
 
-1. Set the environment variable `STRIPE_LOG` to the value `debug` or `info`
+1. Set the environment variable `STRIPE_LOG` to the value `debug`, `info` or
+   `warning`
 
    ```
    $ export STRIPE_LOG=debug
--- a/README.md
+++ b/README.md
@@ -113,8 +113,7 @@ production use, but `debug` is also available for more verbosity.
 
 There are a few options for enabling it:
 
-1. Set the environment variable `STRIPE_LOG` to the value `debug`, `info` or
-   `warning`
+1. Set the environment variable `STRIPE_LOG` to the value `debug` or `info`
 
    ```
    $ export STRIPE_LOG=debug
--- a/tests/test_api_requestor.py
+++ b/tests/test_api_requestor.py
@@ -232,14 +232,17 @@ class TestAPIRequestor(object):
             "api_key": stripe.api_key,
             "api_version": stripe.api_version,
             "enable_telemetry": stripe.enable_telemetry,
+            "default_http_client": stripe.default_http_client,
         }
         stripe.api_key = "sk_test_123"
         stripe.api_version = "2017-12-14"
         stripe.enable_telemetry = False
+        stripe.default_http_client = None
         yield
         stripe.api_key = orig_attrs["api_key"]
         stripe.api_version = orig_attrs["api_version"]
         stripe.enable_telemetry = orig_attrs["enable_telemetry"]
+        stripe.default_http_client = orig_attrs["default_http_client"]
 
     @pytest.fixture
     def http_client(self, mocker):
@@ -484,6 +487,25 @@ class TestAPIRequestor(object):
                 extra={"Stripe-Account": account}, request_method="get"
             ),
         )
+
+    def test_sets_default_http_client(self, http_client):
+        assert not stripe.default_http_client
+
+        stripe.api_requestor.APIRequestor(client=http_client)
+
+        # default_http_client is not populated if a client is provided
+        assert not stripe.default_http_client
+
+        stripe.api_requestor.APIRequestor()
+
+        # default_http_client is set when no client is specified
+        assert stripe.default_http_client
+
+        new_default_client = stripe.default_http_client
+        stripe.api_requestor.APIRequestor()
+
+        # the newly created client is reused
+        assert stripe.default_http_client == new_default_client
 
     def test_uses_app_info(self, requestor, mock_response, check_call):
         try:
--- a/tests/test_api_requestor.py
+++ b/tests/test_api_requestor.py
@@ -231,18 +231,18 @@ class TestAPIRequestor(object):
         orig_attrs = {
             "api_key": stripe.api_key,
             "api_version": stripe.api_version,
-            "enable_telemetry": stripe.enable_telemetry,
             "default_http_client": stripe.default_http_client,
+            "enable_telemetry": stripe.enable_telemetry,
         }
         stripe.api_key = "sk_test_123"
         stripe.api_version = "2017-12-14"
-        stripe.enable_telemetry = False
         stripe.default_http_client = None
+        stripe.enable_telemetry = False
         yield
         stripe.api_key = orig_attrs["api_key"]
         stripe.api_version = orig_attrs["api_version"]
-        stripe.enable_telemetry = orig_attrs["enable_telemetry"]
         stripe.default_http_client = orig_attrs["default_http_client"]
+        stripe.enable_telemetry = orig_attrs["enable_telemetry"]
 
     @pytest.fixture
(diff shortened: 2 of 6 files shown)


tests/test_api_requestor.py
```diff
@@ -232,14 +232,17 @@ def setup_stripe(self):
...
         }
         stripe.api_key = "sk_test_123"
         stripe.api_version = "2017-12-14"
         stripe.enable_telemetry = False
+        stripe.default_http_client = None
         yield
         stripe.api_key = orig_attrs["api_key"]
         stripe.api_version = orig_attrs["api_version"]
         stripe.enable_telemetry = orig_attrs["enable_telemetry"]
+        stripe.default_http_client = orig_attrs["default_http_client"]
```
brandur-stripe: In this instance and the two above it, could you make it so that `default_http_client` is inserted into this list alphabetically?

jameshageman-stripe: Sure thing!

brandur-stripe:
@jameshageman-stripe Just a note that there was an internal JIRA ticket opened contemporaneously with this PR where Alex expressed some usability concern regarding setting `proxy` after an HTTP client's been initialized. I mentioned you on it, but just double-posting here to make sure you see it before we action too much further here.

brandur-stripe:
And one more note that we've temporarily removed Pypy3 from the build matrix until we can get a new release of Pipenv to fix that Pypy3 problem: https://github.com/stripe/stripe-python/pull/528 You should be able to get a fully passing build on a rebase.

jameshageman-stripe:
@brandur-stripe @ob-stripe I have written a few integration tests that spin up a real `HTTPServer` on a random port. 

One of the tests exposed an interesting find: the global `stripe.proxy` doesn't seem to get used. The only way to configure the proxy is to pass it to the client's constructor like this:

```python
stripe.default_http_client = stripe.http_client.new_default_http_client(proxy="http://localhost:1234")
```

My integration tests show that `stripe.proxy` does not work, as this test fails:

https://github.com/stripe/stripe-python/blob/3b86e58e2ebd55bd7a29b0c488a77e5b3045ef1d/tests/test_integration.py#L77-L96

but this test passes:

https://github.com/stripe/stripe-python/blob/3b86e58e2ebd55bd7a29b0c488a77e5b3045ef1d/tests/test_integration.py#L98-L119

I think there are two paths forward. Either we fix the client to respect `stripe.proxy`, or we completely remove `stripe.proxy` from the code and documentation since it never worked anyway.

jameshageman-stripe:
My mistake, I messed up the test. `stripe.proxy` is used, _only_ when a no custom client is created. Here is the usage: https://github.com/stripe/stripe-python/pull/526/files#diff-06e8e6a0633cc19f1fef8fbc334012cfR87. 

Back to the discussion around what to do about changing the proxy between requests, I think a warning will suffice. I'll try to work that into this PR.

jameshageman-stripe:
@brandur-stripe I've added the warning with appropriate tests.

Aside: I'd like to keep the `tests/test_integration.py` tests in here, although I think it could use a better name. This library has a whole bunch of abstractions, making it difficult to know which layer to mock out when writing tests. The integration tests file actually spins up a real HTTPServer and points the client at it, to allow us to assert that requests actually hit the expected endpoint. It should also come in handy when I go to fix the client telemetry stuff. It's a little different than then rest of our unit tests, so let me know what you think of them.

tests/test_integration.py
```diff
@@ -0,0 +1,137 @@
...
+        stripe.default_http_client = orig_attrs["default_http_client"]
+        stripe.log = orig_attrs["log"]
+        stripe.proxy = orig_attrs["proxy"]
+
+    def setup_mock_server(self, handler):
+        # Configure mock server.
+        self.mock_server_port = get_free_port()
+        self.mock_server = HTTPServer(
+            ("localhost", self.mock_server_port), handler
+        )
```
brandur-stripe: Yeesh, I was surprised to find that Python's docs for this are astoundingly awful, but passing `0` here should have the server select a port, which allows you to remove `get_free_port` and a possible race condition.

Here's an example with `ThreadedTCPServer`:

``` python
if __name__ == "__main__":
    # Port 0 means to select an arbitrary unused port
    HOST, PORT = "localhost", 0

    server = ThreadedTCPServer((HOST, PORT), ThreadedTCPRequestHandler)
    ip, port = server.server_address

    # Start a thread with the server -- that thread will then start one
    # more thread for each request
    server_thread = threading.Thread(target=server.serve_forever)
    # Exit the server thread when the main thread terminates
    server_thread.daemon = True
    server_thread.start()
    print "Server loop running in thread:", server_thread.name

    client(ip, port, "Hello World 1")
    client(ip, port, "Hello World 2")
    client(ip, port, "Hello World 3")

    server.shutdown()
    server.server_close()
```

README.md
```diff
@@ -113,7 +113,8 @@ production use, but `debug` is also available for more verbosity.
 
 There are a few options for enabling it:
 
-1. Set the environment variable `STRIPE_LOG` to the value `debug` or `info`
+1. Set the environment variable `STRIPE_LOG` to the value `debug`, `info` or
+   `warning`
```
brandur-stripe: I couldn't tell that easily from the changes whether this already happens, but warnings should be emitted by default (if `STRIPE_LOG` isn't set), because otherwise no one will ever see them. Does the code do that right now?

tests/test_integration.py
```diff
@@ -0,0 +1,137 @@
...
+
+        stripe.proxy = "http://bad-url"
+        logger_mock.assert_not_called()
+        stripe.Balance.retrieve()
+        logger_mock.assert_called_with(
+            "stripe.proxy was updated after sending a request - this is a no-op. To use a different proxy, set stripe.default_http_client to a new client configured with the proxy."
+        )
+        assert MockServerRequestHandler.num_requests == 2
+
+    def test_hits_client_proxy(self):
```
brandur-stripe: Can we name this one something like `test_hits_proxy_through_default_http_client`?

README.md
brandur-stripe: Could we have it respect either `warn` or `warning`? Especially in logging libraries these two are often interchanged that it'd be sort of a nice feature to support both. If we're supporting only one, I'd say that we might want to do `warn` instead.

jameshageman-stripe: @brandur-stripe My code currently requires a log level to be set in order to show warnings. I just did this to follow the existing pattern from `info` and `debug` - I'm fine with changing it.

We could simplify the interface and not allow `warn` (or `warning`) at all, and just always print warnings. Is there a case (`STRIPE_LOG`/`stripe.log` value) where you wouldn't expect warnings to be printed?

jameshageman-stripe: That sounds a whole lot simpler - I'll do that!

jameshageman-stripe:
@brandur-stripe @ob-stripe re-review?

brandur-stripe:
LGTM. Thanks James!

@ob-stripe Excellent call on using the built-in warnings module — that one was a TIL for me.

brandur-stripe:
Released as 2.20.0.

ob-stripe:
Er, sorry for the late feedback, but I wonder if this change is thread safe, since now all threads will share the same client.

Should we do something similar to stripe-ruby and use thread-local storage to store a client per thred instead?

jameshageman-stripe:
@ob-stripe Good catch. I can look into it.

jameshageman-stripe:
That sounds good to me - I came across the same issue while working on #530. We can't avoid sharing clients between threads, so we should ensure that the clients are thread safe.

brandur-stripe:
Took a shot at fixing the thread safety problem for `RequestsClient` in #531.

PR #527, opened 2019-01-25
Fix quotes in `version.py`

brandur:
The project recently switched over to Black for code formatting in #521.
Black expects double-quoted strings throughout the project. We
double-quoted this file when that patch was merged, but our release
script was still incorrectly producing single quoted strings.

I'm just about to patch the release script, but putting this fix in here
so that we can fix `master`.

cc @ob-stripe


brandur-stripe:
Thanks @ob-stripe!

ob-stripe:
I actually already fixed the release script, but not before running in once pre-fix, which resulted in the single-quoted string being committed. Sorry!

brandur-stripe:
> I actually already fixed the release script, but not before running in once pre-fix, which resulted in the single-quoted string being committed. Sorry!

Oh dang, no worries!

I think what might've happened here is that you got the string _matcher_, but not the string _replacer_. Unlike many other languages, Python's scripts in this area actually check that the value is present before doing anything with it. Either that, or there was some other open PR that was never merged.

PR #528, opened 2019-01-25
Allow failures on Pypy3

brandur:
Pypy3's been breaking pretty regularly due to an import error:

    ImportError: cannot import name 'pythonapi'
    make: *** [init] Error 1

Here we (temporarily) allow it to fail in the build matrix so that we
can get the general build back to green.

r? @ob-stripe This is obviously not great, but a failing master build
isn't either :) What do you think about bringing this in temporarily and
trying to get the Pypy3 build fixed later? It's not too obvious to me
why it's failing -- I'd have to spend some time digging in.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,16 @@ matrix:
     - python: "pypy"
     - python: "pypy3"
 
+  # There's a Pipenv issue that currently prevents stripe-python from building
+  # properly on Pypy3:
+  #
+  #     https://github.com/pypa/pipenv/pull/3322
+  #
+  # Olivier patched it, but we're still waiting on the Pipenv project for a new
+  # release. When that happens, we can remove this `allow_failures` section.
+  allow_failures:
+    - python: "pypy3"
+
 cache:
   directories:
     - stripe-mock


ob-stripe:
Yeah, it's annoying. I fixed the pipenv issue with PyPy 3 a while back (https://github.com/pypa/pipenv/pull/3322), but they haven't released any new version since then.

Mind adding a comment to explain this is just a pipenv/PyPy3 compatibility issue? Feel free to merge afterwards.

brandur-stripe:
Oh, right! Now that you mention it, I remember you talking about that. That is annoying :)

I added a comment explaining that it should be possible to remove Pypy3 from `allow_failures` soon with a link to your PR. Thanks!

jleclanche:
@brandur-stripe @ob-stripe A bit crass but you could apply [the patch](https://github.com/pypa/pipenv/commit/79da8e382564c7db69a366e51fd5ffac3026ca6a.patch) after pipenv is installed but before it's run. This would allow the build to run at least (and who knows if there's other issues in pipenv with pypy)

brandur-stripe:
@jleclanche Great point! My general feeling is that it's probably not worth it right now in that if we just wait a bit longer for a release we can save ~an hour's worth of work and spend it elsewhere. Let's revisit that if the release turns out to be a *really* long time coming though — I'm not opposed to giving the idea a shot at that point.

brandur-stripe:
@ob-stripe Yep. Up to you on that one! Might be worth it if you know you can do it quickly.

PR #531, opened 2019-01-29
Make `RequestsClient` thread-safe

brandur:
It came up in #526 that our implementation isn't quite thread-safe.
*Most* HTTP clients are by virtue of the fact that their underlying
libraries are, but the recommended client, `RequestsClient` for the
Requests library, creates and reuses only a single session for all the
requests it makes.

The Requests maintainers are non-committal when it comes to the thread
safety of a session, but the general sentiment of its userbase is that
sessions are not thread safe.

Here we tweak the `RequestsClient` implementation so that we create one
session per thread by leveraging Python's thread-local storage.

Unfortunately, this implementation still leaves a pitfall in that if you
explicitly pass in a session in, it will get used between all threads.
The ideal thing to do here I think would be to probably not accept a
session object anymore, but that'd be a breaking change. One possibility
is that we could start warning about the use of the session parameter,
but for the time being I've opted to just leave this as a known problem
(given that hopefully this patch will still be a strict improvement).

Note this will conflict with #530, but rebasing either branch should be
simple.

r? @ob-stripe
cc @jameshageman-stripe
cc @stripe/api-libraries

### DIFF
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -6,6 +6,7 @@ import warnings
 import email
 import time
 import random
+import threading
 
 import stripe
 from stripe import error, util, six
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Thanks OB!

PR #532, opened 2019-01-30
Add support for Subscription Schedules

remi-stripe:
This adds two new resources: Subscription Schedule and Subscription Schedule Revision. The latter is nested under the former.

This should be ready even though we won't merge until all libraries are ready and stripe-mock supports those new endpoints publicly.

r? @ob-stripe
cc @stripe/api-libraries @alexander-stripe

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ cache:
 env:
   global:
     # If changing this number, please also change it in `tests/conftest.py`.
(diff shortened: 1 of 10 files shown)


remi-stripe:
@ob-stripe stripe-mock has been released, tests are now passing. Can you review so that we can ship tomorrow (or today for you) please? Thank youuuuuuu

ob-stripe:
Released as 2.21.0.

PR #534, opened 2019-01-30
Fix session initialization for multi-threaded environments

### LINKED ISSUE #533: '_thread._local' object has no attribute 'session'
Since version 2.20.1 I get the following error when I call `stripe.Source.retrieve` (or any other function):

File "VIRTUALENV_PATH/lib/python3.5/site-packages/stripe/http_client.py" in request
  211.                 result = self._thread_local.session.request(

During handling of the above exception ('_thread._local' object has no attribute 'session'), another exception occurred:

[...]

File "VIRTUALENV_PATH/lib/python3.5/site-packages/stripe/http_client.py" in _handle_request_error
  288.         raise error.APIConnectionError(msg, should_retry=should_retry)

Going back to 2.20.0 works. Trying it from a Python shell works.

Stripe Python library: 2.20.1
Python: 3.5.3
requests: 2.21.0
Apache/2.4.25 (Debian) w. WSGI

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Fixes #533.

In #531, we initialized the session in a thread-local store in `RequestsClient`'s constructor. However, a single client instance is initialized for all threads, so only the first thread (that initializes the client) gets a session, and all other threads crash when trying to make a request because `self._thread_local.session` doesn't exist.

This PR fixes this issue by moving the session initialization in the `request()` method. A new session is only initialized if one doesn't already exist in the thread-local store. This way, each thread gets a session the first time it tries to make a request.

We should probably have an automated test for this, but in the meantime I've tested this with a trivial script:
```python
import threading

import stripe


stripe.api_key = "sk_test_..."

def worker():
    print(stripe.Account.retrieve().id)
    return

threads = []
for i in range(5):
    t = threading.Thread(target=worker)
    threads.append(t)
    t.start()
```

### DIFF
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -194,7 +194,6 @@ class RequestsClient(HTTPClient):
     def __init__(self, timeout=80, session=None, **kwargs):
         super(RequestsClient, self).__init__(**kwargs)
         self._timeout = timeout
-        self._thread_local.session = session or requests.Session()
 
     def request(self, method, url, headers, post_data=None):
         kwargs = {}
@@ -205,6 +204,9 @@ class RequestsClient(HTTPClient):
 
         if self._proxy:
             kwargs["proxies"] = self._proxy
+
+        if getattr(self._thread_local, "session", None) is None:
+            self._thread_local.session = requests.Session()
 
         try:
             try:
@@ -288,7 +290,7 @@ class RequestsClient(HTTPClient):
         raise error.APIConnectionError(msg, should_retry=should_retry)
 
     def close(self):
-        if self._thread_local.session is not None:
+        if getattr(self._thread_local, "session", None) is not None:
             self._thread_local.session.close()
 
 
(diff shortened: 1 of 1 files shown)


stripe/http_client.py
```diff
@@ -206,6 +205,9 @@ def request(self, method, url, headers, post_data=None):
         if self._proxy:
             kwargs["proxies"] = self._proxy
 
+        if getattr(self._thread_local, "session", None) is None:
+            self._thread_local.session = requests.Session()
```
jameshageman-stripe: Are we ignoring the `session` that was possibly passed to the constructor?

brandur-stripe: Yeah, it might be a good idea to store the `session` parameter as `_session` on the instance, and then use it here if available like:

``` python
self._thread_local.session = _session or requests.Session()
```

Still has the problem of not being thread safe if used, but should be fine in most situations.

brandur-stripe:
Oh man, that was really stupid of me. Thanks for the quick fix OB! It must be late there so LMK if you want me to take this over.

ob-stripe:
@brandur-stripe Yeah, would be great if you could pick this one up, thanks! Also take a look at the multithread tests James pushed in #535.

brandur-stripe:
@ob-stripe Sounds good! Just pushed a commit into your branch.

PR #538, opened 2019-02-28
Mention how to set API version per request

remi-stripe:
Improve the readme per feedback from @scherr

r? @rattrayalex-stripe 
cc @stripe/api-libraries

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,9 @@ stripe.Charge.retrieve("ch_1A2PUG2eZvKYlo2C4Rej1B9d")
 
 ### Per-request Configuration
 
-For apps that need to use multiple keys during the lifetime of a process, like
-one that uses [Stripe Connect][connect], it's also possible to set a
-per-request key and/or account:
+Configure individual requests with keyword arguments. For example, you can make
+requests with a specific [Stripe Version](https://stripe.com/docs/api#versioning)
+or as a [connected account](https://stripe.com/docs/connect/authentication#authentication-via-the-stripe-account-header):
 
 ```python
 import stripe
@@ -57,14 +57,16 @@ import stripe
 # list charges
 stripe.Charge.list(
     api_key="sk_test_...",
-    stripe_account="acct_..."
+    stripe_account="acct_...",
+    stripe_version="2019-02-19"
 )
 
 # retrieve single charge
 stripe.Charge.retrieve(
     "ch_1A2PUG2eZvKYlo2C4Rej1B9d",
     api_key="sk_test_...",
-    stripe_account="acct_..."
+    stripe_account="acct_...",
+    stripe_version="2019-02-19"
 )
 ```
 


README.md
```diff
@@ -47,24 +47,26 @@ stripe.Charge.retrieve("ch_1A2PUG2eZvKYlo2C4Rej1B9d")
 
 ### Per-request Configuration
 
-For apps that need to use multiple keys during the lifetime of a process, like
-one that uses [Stripe Connect][connect], it's also possible to set a
-per-request key and/or account:
+For apps that need to use multiple keys during the lifetime of a process, or
```
rattrayalex-stripe: hmm, what do you think about 

> For apps that need to use multiple keys during the lifetime of a process (eg; for [Connect][connect]), ...

rattrayalex-stripe: This way, the reader can quickly decide "this applies to me, pay attention and unpack this 'multiple keys' thing" or "this does not apply to me, can ignore this 'multiple keys' thing"

remi-stripe: Well Connect is not a valid example anymore since with Connect you have one key (yours) and a separate header for the account id, that's why I removed it. What do you think?

rattrayalex-stripe: Hmm, good point. We should probably change the first clause to not mention multiple keys, and instead say something more connect-specific. 

We should also remove `api_key="sk_test_..."` from the example here, though I do think we should add a paragraph just below mentioning something like 

> You can also set `api_key` per request, though this is deprecated in favor of `stripe_account`. 

Undocumenting deprecated things completely is an antipattern IMO since it can make porting off the deprecated behavior more confusing/intimidating, but we should also definitely not document deprecated things as if they were preferred 😅

remi-stripe: I'm not sure I agree. For me the readme should be an easy way to understand basic things and so this section is about the per-request configuration and not really about Connect itself.

rattrayalex-stripe: We shouldn't explain connect, just something to the effect of

> Configure individual requests with keyword arguments. For example, you can make requests with a specific [Stripe Version](...) or as a [connected account](...):

remi-stripe: Ah, but I think setting the API key is an important thing to keep personally. Let's not over-argue, I think it might be best to leave as is in that case and improve our official docs to cover all of this in details

rattrayalex-stripe: Ah, interesting. Curious what that's useful for outside of Connect. 

What do you think of using this text, but with `api_key=` still in the example?

> Configure individual requests with keyword arguments. For example, you can make requests with a specific [Stripe Version](...) or as a [connected account](...):

remi-stripe: Lots of merchants have more than one account and have to do this (or often hit the whole "no such token" error because 2 requests overlap)

remi-stripe: And yes I like your last proposal it sounds great. Do you mind fixing my PR directly?

remi-stripe:
Ended up making the update. PTAL @rattrayalex-stripe

rattrayalex-stripe:
Thanks! Sorry this slipped through the cracks.

LGTM

PR #539, opened 2019-03-04
Updating init to ensure all classes are imported to handle all Stripe responses

jackvz:
Hello there,

With a Python 2.7 configuration I get an error when working with Mezzanine Cartridge and the Stripe package. I have just tested it with the following configuration... 

Python 3.7.2
Mezzanine==4.2.3
Cartridge==0.13.0
cartridge-payments==0.97.0
stripe==2.21.0

...and get the same [error](https://github.com/jackvz/file-share-for-github-chat/blob/master/AttributeError%20at%20_shop_checkout_.pdf), or rather lack of an error response from Stripe, when, for example the API key is not correctly configured:

AttributeError at /shop/checkout/
module 'stripe' has no attribute 'CardError'

Here is a fix.

Thanks

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -45,6 +45,9 @@ from stripe.oauth import OAuth  # noqa
 # Webhooks
 from stripe.webhook import Webhook, WebhookSignature  # noqa
 
+# Errors
+from stripe.error import *
+
 
 # Sets some basic information about the running application that's sent along
 # with API requests. Useful for plugin authors to identify their plugin when


ob-stripe:
Hi @jackvz, thanks for the contribution.

We've purposefully removed exception classes from the top-level `stripe` namespace in version 2.0 (cf. #399), so unfortunately I'll have to decline this PR. Exception classes should now be accessed via `stripe.error`.

My guess is that either your code or one of the packages you're using (`cartridge-payments` sounds like the likely culprit) was not updated accordingly. If the error happens in your own code, you should simply update it to access exception classes via `stripe.error`, e.g. `stripe.error.CardError`. If the error happens in one of your dependencies, then you'd either have to ask the package maintainer to update their code, or pin the `stripe` package to the last 1.x version (1.84.2).

jackvz:
Hi Olivier. Great, the last 1.x version sorted it, thanks.

Issue #219, opened 2016-04-04
Get a dictionary representation of Stripe objects with only build-in Python types


vcatalano:
I'm currently using the `convert_to_stripe_object` in one of my libraries and this allows me convert a Python dictionary of my object to a Stripe object. Is there a way to basically do the reverse of this function in which the resulting object contains only Python built-in types (i.e. dict, list, string, boolean and number)?

In my case, I'm attempting to save a Customer object to a NoSQL database. The database only supports Python built-in types. However, when I attempt to create a create a dictionary from my customer object using `dict(customer_obj)`, the resulting object is not serializable to JSON since it contains lists of Subscription and Card objects. This has ultimately led me to use the following:

```python
    import json
    customer_obj = stripe.Customer.create(...)
    json.dumps(customer_obj, sort_keys=True, indent=2)
    customer_dict = json.loads(customer_json)
```

As you can see, this is not ideal. Is it possible that I'm just missing obvious in my transformation of a Customer object to a dictionary?

brandur:
Hi @vcatalano,

Unfortunately, to my knowledge there's no way of doing exactly what you're trying to do here given the current state of the bindings. They're written to only have to handle deserialization rather than go the other way around.

One thing that's pretty close is that `StripeObject` has a method called `serialize` which is used to turn a new or updated object into parameters to send back up to the API. Before encoding those parameters look like a basic dictionary. I think the only problematic difference compared to what you're trying to do is that `serialize` tries to be smart about things and only includes fields that have changed since the object is loaded.

We could consider adding an optional parameter to `serialize` along the lines of `force=false` that could be used to serialized all fields regardless of whether they're saved or unsaved. I'll doubt we'd tackle that one, but we'd certainly consider a well-tested patch!

vcatalano:
Thanks for the insight, @brandur! It looks like I'll be sticking with my serializing-deserializing solution for the time being.

duaneking:
## Please Reopen This
![please](https://user-images.githubusercontent.com/569341/53853481-769be900-3f7a-11e9-9316-723dc713a17b.gif)

Respectfully, I would like to request that this should be reopened, or even be linked to a new bug filed on the overall architecture of how result objects should be handled in the python library to make it higher quality and provide for the missed opportunities it currently presents.

In short, this python client API did not get the architectural attention it deserved and while its clear the person who did it came from an understanding of interference in OOP it can still be greatly improved.

[...]

PR #541, opened 2019-03-14
Add convert_to_dict method

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds a `stripe.util.convert_to_dict` method that recursively converts `StripeObject` instances to `dict`s.

Partial fix for #219.

### DIFF
--- a/stripe/util.py
+++ b/stripe/util.py
@@ -260,6 +260,26 @@ def convert_to_stripe_object(
         return resp
 
 
+def convert_to_dict(obj):
+    """Converts a StripeObject back to a regular dict.
+
+    Nested StripeObjects are also converted back to regular dicts.
+
+    :param obj: The StripeObject to convert.
+
+    :returns: The StripeObject as a dict.
+    """
+    if isinstance(obj, list):
+        return [convert_to_dict(i) for i in obj]
+    # This works by virtue of the fact that StripeObjects _are_ dicts. The dict
+    # comprehension returns a regular dict and recursively applies the
+    # conversion to each value.
+    elif isinstance(obj, dict):
+        return {k: convert_to_dict(v) for k, v in six.iteritems(obj)}
+    else:
+        return obj
+
+
 def populate_headers(idempotency_key):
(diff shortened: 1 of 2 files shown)


stripe/util.py
```diff
@@ -260,6 +260,15 @@ def convert_to_stripe_object(
         return resp
 
 
+def convert_to_dict(obj):
+    if isinstance(obj, list):
+        return [convert_to_dict(i) for i in obj]
+    elif isinstance(obj, dict):
```
brandur-stripe: Would you mind adding a comment here that this will work on a `StripeObject` because it will advertise itself as a `dict` and allow itself to be iterated (I think that's the reason anyway).

stripe/util.py
```diff
@@ -260,6 +260,15 @@ def convert_to_stripe_object(
         return resp
 
 
+def convert_to_dict(obj):
```
brandur-stripe: Minor note that if you send a list into `convert_to_dict`, it'll convert it to a list, not a dict.

I'd consider something like `convert_to_data`, but overall that probably reduces clarity, so `convert_to_dict` is probably our best bet.

stripe/util.py
ob-stripe: 👍

stripe/util.py
ob-stripe: Yeah, I initially named the method `convert_to_python_types`, but I think `convert_to_dict` is clearer since users are unlikely to pass anything but a `StripeObject` to this method.

brandur-stripe: Yeah, good call. +1.

ob-stripe:
I've added a comment as requested, as well as a docstring for the method to help advertise it. We don't use docstrings anywhere else in the library, but we probably should.

ptal @brandur-stripe

brandur-stripe:
Thanks OB. LGTM!

ob-stripe:
Released as 2.22.0.

PR #542, opened 2019-03-16
Add support for Terminal Location and Reader deletion

remi-stripe:
cc @stripe/api-libraries

### DIFF
--- a/tests/api_resources/test_account.py
+++ b/tests/api_resources/test_account.py
@@ -37,36 +37,6 @@ class TestAccount(object):
         assert isinstance(resource, stripe.Account)
         assert resource is account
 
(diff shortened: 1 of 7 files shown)


tests/api_resources/test_account.py
```diff
@@ -37,36 +37,6 @@ def test_is_saveable(self, request_mock):
         assert isinstance(resource, stripe.Account)
         assert resource is account
 
-    def test_is_saveable_with_additional_owners(self, request_mock):
```
remi-stripe: I had to remove the test as openapi does not support `legal_entity` anymore. We could mock if needed but I think it's fine to punt on that test.

PR #547, opened 2019-04-02
`custom_method` decorator for defining static methods for custom API requests

ob-stripe:
cc @stripe/api-libraries 

This is similar to https://github.com/stripe/stripe-ruby/pull/754. This PR defines a new `custom_method` decorator that can be applied to resource classes to automatically define class methods for custom (non-CRUDL) API requests.

The decorator checks if a method with the target name already exists, and if so it uses the new decorator `util.class_method_variant` from #543 to avoid overwriting the existing method (that we assume to be an instance method).

### DIFF
--- a/stripe/api_resources/charge.py
+++ b/stripe/api_resources/charge.py
@@ -4,8 +4,10 @@ from stripe import api_requestor, util
 from stripe.api_resources.abstract import CreateableAPIResource
 from stripe.api_resources.abstract import UpdateableAPIResource
 from stripe.api_resources.abstract import ListableAPIResource
+from stripe.api_resources.abstract import custom_method
 
 
+@custom_method("capture", http_verb="post")
 class Charge(
     CreateableAPIResource, ListableAPIResource, UpdateableAPIResource
 ):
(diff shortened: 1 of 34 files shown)


stripe/api_resources/charge.py
```diff
@@ -4,8 +4,10 @@
 from stripe.api_resources.abstract import CreateableAPIResource
 from stripe.api_resources.abstract import UpdateableAPIResource
 from stripe.api_resources.abstract import ListableAPIResource
+from stripe.api_resources.abstract import custom_method
 
 
+@custom_method("capture", "post")
```
brandur-stripe: Kind of like in Ruby, can we use the named variant for `http_verb` and `http_path` for a little extra clarity?

ob-stripe: Sure! We can't force it since Python handles keyword args differently, but we can just do it by convention when using `@custom_method`.

brandur-stripe: Ah yes, that's what I meant (by convention). Let's do it!

remi-stripe:
@ob-stripe I added all custom methods and tests and rebased on the latest branch for deletion

ob-stripe:
Released as 2.24.0.

PR #550, opened 2019-04-05
Fix multipart encoding of nested parameters

ob-stripe:
cc @stripe/api-libraries 

This one was a bit tricky. stripe-python rolls its own multipart encoder, which only supported two types of values: file-like (i.e. an object that has a `read` attribute) or strings. All other types caused an exception to be raised.

So there were actually two different issues to fix:
1. the parameters need to be flattened (which solves the problem for compound types -- lists and dicts are gone after this step)
2. handle non-string scalar types (booleans, integers, etc.)

The first issue was easily solved by calling `api_requestor._api_encode()` in `MultipartDataGenerator` to flatten the parameters.

~To solve the second issue, I changed `util.utf_8()` to stringify all non-file-like values. This could potentially have side effects for regular form-encoding, but I'm fairly sure that's not the case: form-encoding works by passing the dictionary of flattened params to `urlencode` which would stringify non-string values anyway.~

The second issue is solved by stringifying non-file-like values before adding them to the multipart buffer.

### DIFF
--- a/stripe/multipart_data_generator.py
+++ b/stripe/multipart_data_generator.py
@@ -3,7 +3,7 @@ from __future__ import absolute_import, division, print_function
 import random
 import io
 
-from stripe import six
+import stripe
 
 
 class MultipartDataGenerator(object):
@@ -14,7 +14,10 @@ class MultipartDataGenerator(object):
         self.chunk_size = chunk_size
 
     def add_params(self, params):
-        for key, value in six.iteritems(params):
+        # Flatten parameters first
+        params = dict(stripe.api_requestor._api_encode(params))
+
+        for key, value in stripe.six.iteritems(params):
             if value is None:
                 continue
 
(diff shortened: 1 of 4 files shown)


remi-stripe:
> To solve the second issue, I changed util.utf_8() to stringify all non-file-like values. This could potentially have side effects for regular form-encoding, but I'm fairly sure that's not the case: form-encoding works by passing the dictionary of flattened params to urlencode which would stringify non-string values anyway.

Could we special-case this just for file upload? My inner paranoia is tingling right now with that sentence :p

ob-stripe:
Not sure why the PR is failing on Travis for some Python versions -- I suspect a circular import issue, but I can't reproduce locally. I'll try shuffling some imports around anyway.

ob-stripe:
@remi-stripe I've updated the implementation to only stringify values in `MultipartDataGenerator`.

remi-stripe:
Thanks for making the change. This looks great to me and I like that you explicitly tested a nested dict!

ob-stripe:
Thanks both!

ob-stripe:
Released as 2.24.1.

PR #561, opened 2019-04-25
Replace pipenv with poetry

ob-stripe:
This PR gets rid of both `setup.py` and pipenv, and uses [Poetry](https://poetry.eustace.io/) to both manage dependencies and build the library.

Unfortunately, at the moment it looks like Poetry suffers from the same issue as pipenv: when setting up the development environment, it tries to install the `stripe` library in editable mode, which doesn't work with recent versions of pip.

I'm still not sure that Poetry will work better for us than pipenv did. For now I think we should just wait a few more weeks to see if Poetry fixes this issue.


brandur-stripe:
Well, that was fast! I don't have a great feeling for which dependency manager would be best, but this seems plausible.

> I'm still not sure that Poetry will work better for us than pipenv did. For now I think we should just wait a few more weeks to see if Poetry fixes this issue.

SG.

ob-stripe:
Just a note that pipenv is still broken and still hasn't released any new version since last year, but poetry fixed the issue. The build is now green on Travis without any changes to the PR.

I'm still not 100% sold on switching to Poetry, but at the very least the project seems more active.

brandur-stripe:
Looks good to me if you want to try it!

ob-stripe:
Okay, I've spent some time comparing the source distribution and wheel file built by `poetry build` in this branch vs. those produced by running `python setup.py sdist bdist_wheel` in the master branch, and the results are essentially the same. There are some differences to the resulting metadata files but nothing that seems significant.

We'll need to upgrade our internal tooling to update the version number inside `pyproject.toml` and run `poetry build` instead of `python setup.py sdist bdist_wheel` but that should be pretty straightforward.

Just in case you have final thoughts before I merge this: ptal @brandur-stripe

brandur-stripe:
Go for it OB.

ob-stripe:
Alright, going to merge this and attempt a release.

ob-stripe:
Released as 2.29.2 \o/

PR #565, opened 2019-05-08
Safe codegen transformations

dmitrykorolev-stripe:


ob-stripe:
@dmitrykorolev-stripe FYI, the Python library uses [Black](https://github.com/python/black) for code formatting. Some of these changes, while safe, result in broken builds because Black would have formatted them differently.

jleclanche:
@ob-stripe I recommend setting up an isort config that is compatible with black.

See here: https://github.com/timothycrosley/isort/issues/694

isort mode 3 is what you want.

ob-stripe:
@jleclanche Thanks, this is helpful!

However I think the issue here is not import order (which Black doesn't seem to care about with the current configuration) but changes in the formatting of parent classes like here: https://github.com/stripe/stripe-python/pull/565/files#diff-0b7eca4f345d4708e221dcf65baa0388.

Enforcing import order via Black seems like a net win though, I'll look into it.

jleclanche:
@ob-stripe note that black won't enforce import order on its own, but it will modify their style (eg. for long lines). isort does that too, so that is why if you use both tools you need to ensure they use a compatible mode.

ob-stripe:
Ah gotcha, thanks for the information!

rattrayalex-stripe:
This went through in #573 – thanks @dmitrykorolev-stripe !

PR #570, opened 2019-05-17
Enhanced support for DRF

anshengme:
The request.data obtained by the DRF is a dict type. 
Is not required to do json.loads

```python
class OrderCallBackViewSet(mixins.CreateModelMixin,
                           viewsets.GenericViewSet):
    ......
    def create(self, request, *args, **kwargs):
        # payload = request.data  => payload==dict Object
        ......
```

### DIFF
--- a/stripe/webhook.py
+++ b/stripe/webhook.py
@@ -20,7 +20,10 @@ class Webhook(object):
             payload = payload.decode("utf-8")
         if api_key is None:
             api_key = stripe.api_key
-        data = json.loads(payload)
+        if isinstance(payload, dict):
+            data = payload
+        else:
+            data = json.loads(payload)
         event = stripe.Event.construct_from(data, api_key)
 
         WebhookSignature.verify_header(payload, sig_header, secret, tolerance)


brandur-stripe:
Will leave this one to you OB :)

Out of curiosity though, does DRF = Django REST Framework?

anshengme:
> 将这一个留给你OB :)
> 
> 出于好奇，DRF = Django REST框架？

Yes, it is djangorestframework

anshengme:
> Will leave this one to you OB :)
> 
> Out of curiosity though, does DRF = Django REST Framework?

yes, drf==djangorestframework

ob-stripe:
Hi @anshengme, thanks for the contribution.

Regardless of the framework you use (Django or otherwise), you must pass the _raw_ payload (as in, the exact series of bytes sent by Stripe in the request's body) to `construct_event`. If the payload is already deserialized to a dict by the framework, then it is virtually guaranteed that the signature verification will fail (because JSON does not guarantee formatting, key order, etc.)

If you want to build a `stripe.Event` instance from a dict without verifying the signature (which is not recommended in production, but can be useful in tests), you can simply do:
```python
event = stripe.Event.construct_from(data, stripe.api_key) # where data is a dict
```

Hope this helps! Closing the PR, but feel free to reply if you have questions.

PR #573, opened 2019-05-20
Safe codegen formatting changes 

rattrayalex-stripe:
Supersedes https://github.com/stripe/stripe-python/pull/565

Just a rebase & squash

r? @ob-stripe 
cc @dmitrykorolev-stripe 
cc @stripe/api-libraries

### DIFF
--- a/stripe/api_resources/account.py
+++ b/stripe/api_resources/account.py
@@ -1,7 +1,8 @@
 from __future__ import absolute_import, division, print_function
 
 import stripe
-from stripe import oauth, six, util
+from stripe import oauth, six
+from stripe import util
 from stripe.api_resources.abstract import CreateableAPIResource
 from stripe.api_resources.abstract import custom_method
 from stripe.api_resources.abstract import DeletableAPIResource
--- a/stripe/api_resources/account.py
+++ b/stripe/api_resources/account.py
@@ -40,6 +40,9 @@ class Account(
         self.refresh_from(self.request("post", url, params, headers))
         return self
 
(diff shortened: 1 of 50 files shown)


stripe/api_resources/account.py
```diff
@@ -3,11 +3,11 @@
 import stripe
 from stripe import oauth, six, util
 from stripe.api_resources.abstract import CreateableAPIResource
+from stripe.api_resources.abstract import custom_method
```
ob-stripe: Nit: I think the sort order should be case-sensitive, i.e. `custom_method` should appear after `UpdateableAPIResource`.

No need to act specifically for this though, as I expect `custom_method` will disappear fairly soon and we'll just generate the methods directly instead.

rattrayalex-stripe:
Thanks ob! I think you're right about the sorting, gonna try to take a quick crack at that. 

Looks like the deepsource warnings are quite ignorable, I assume that's ok.

ob-stripe:
> Looks like the deepsource warnings are quite ignorable, I assume that's ok.

Yes, feel free to ignore -- I enabled it to play around with it and see what sort of issues it would report, but I didn't realize it would add a check on PRs.

rattrayalex-stripe:
Gonna hold off on this for a bit since I'm quite close to having nested methods working in Python

rattrayalex-stripe:
Okay, added two more commits which are small & independently reviewable. 

r? @ob-stripe

rattrayalex-stripe:
Done! Thanks ob!

PR #574, opened 2019-05-23
Fix quality issues reported by DeepSource

sauravsrijan:
Fixes:
- PYL-W0612: Rename unused variable to `_`
- PYL-C0123: Use `isinstance()` for checking type
- PAP-W0011: Use `dict.items()` to iterate over a dictionary
- PAP-W0012: Use `dict.get()` to access value from dictionary instead of an `if` query to check whether the key exists
- PYL-C0325: Remove superfluous brackets `()`
- PYL-R1701: Merge multiple consecutive `isinstance()` calls into one

### DIFF
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -412,7 +412,7 @@ class PycurlClient(HTTPClient):
         if self._proxy:
             # now that we have the parser, get the proxy url pieces
             proxy = self._proxy
-            for scheme, value in proxy.items():
+            for scheme, value in six.iteritems(proxy):
                 proxy[scheme] = urlparse(value)
 
     def parse_headers(self, data):
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -285,8 +285,9 @@ class RequestsClient(HTTPClient):
             err = "%s: %s" % (type(e).__name__, str(e))
             should_retry = False
         # Retry only timeout and connect errors; similar to urllib3 Retry
-        elif isinstance(e, requests.exceptions.Timeout) or isinstance(
-            e, requests.exceptions.ConnectionError
+        elif isinstance(
+            e,
+            (requests.exceptions.Timeout, requests.exceptions.ConnectionError),
         ):
             msg = (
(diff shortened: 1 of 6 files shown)


sauravsrijan:
Hi @ob-stripe, here are some more code quality issues fixed.
I would also suggest you to update the `.deepsource.toml` config file to make the analysis better and ignore issues which might not be important for you.

#### Suggestions: 
- Add `tests/**` to test-patterns to ignore running analysis on test files.
- Set `max-line-length` to 100 in the toml. You would need to add an `[analyzers.meta]` field and set it inside it. You can see an example [here.](https://deepsource.io/docs/analyzers/python.html#meta)

stripe/http_client.py
```diff
@@ -411,8 +412,8 @@ def __init__(self, verify_ssl_certs=True, proxy=None):
         if self._proxy:
             # now that we have the parser, get the proxy url pieces
             proxy = self._proxy
-            for scheme in proxy:
-                proxy[scheme] = urlparse(proxy[scheme])
+            for scheme, value in proxy.items():
```
ob-stripe: Nit: can you use `six.iteritems(proxy)` instead? It's slightly more efficient on Python 2 and is consistent with how we iterate over dictionaries in the rest of the library.

sauravsrijan: Sure. I'll make changes.

sauravsrijan:
Updated the PR. Good to go.

ob-stripe:
Awesome, thanks!

ob-stripe:
Released as 2.28.2.

ob-stripe:
> In the config, add `tests/**` in test-patterns to ignore running analysis on test files.

@sauravsrijan I've already added `tests/*` per the example in https://deepsource.io/docs/configuration/analysis.html#test-patterns. Do I need `**` so that it applies to all subdirectories as well?

sauravsrijan:
You have to do `tests/**` to discover files recursively.
`tests/*` will only match one level.

PR #578, opened 2019-05-31
Verify signature before deserializing event

### LINKED ISSUE #577: Check webhook signature before constructing `Event` from the payload
Hi!

I happened to be poking around in the details of your webhook signature-checking code, and I noticed that it does some processing on the event payload before checking the signature, in particular calling `stripe.Event.construct_from` on the not-yet-authenticated data:

  https://github.com/stripe/stripe-python/blob/15d03381b3a3019517e7a31e1607bb5af0bda5a6/stripe/webhook.py#L23-L26

I don't have any reason to believe that this can be used to trigger bad behaviour, but `Event.construct_from` seems to do quite a lot of dynamic work to hydrate the data into live objects. Calling `WebhookSignature.verify_header` before doing that work could be a nice defense-in-depth measure against bugs elsewhere in the codebase.

Just a small suggestion that I wanted to pass along; thanks for providing this nice library, and for an API that seems to strongly encourage the webhook-handling code down the right path security-wise!

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @rfk

Verify the signature from the `Stripe-Signature` header before deserializing the JSON payload into an `Event` instance.

Fixes #577.

### DIFF
--- a/stripe/webhook.py
+++ b/stripe/webhook.py
@@ -18,12 +18,11 @@ class Webhook(object):
     ):
         if hasattr(payload, "decode"):
             payload = payload.decode("utf-8")
-        if api_key is None:
-            api_key = stripe.api_key
-        data = json.loads(payload)
-        event = stripe.Event.construct_from(data, api_key)
 
         WebhookSignature.verify_header(payload, sig_header, secret, tolerance)
+
+        data = json.loads(payload)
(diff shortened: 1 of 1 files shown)


ob-stripe:
Released as 2.29.1.

PR #579, opened 2019-05-31
Move object_classes and some imports to other files for easier codegen

rattrayalex-stripe:
r? @ob-stripe 
cc @stripe/api-libraries


stripe/util.py
```diff
@@ -130,105 +130,22 @@ def secure_compare(val1, val2):
...
 def convert_to_stripe_object(
     resp, api_key=None, stripe_version=None, stripe_account=None
 ):
-    global OBJECT_CLASSES
-
-    if not OBJECT_CLASSES:
-        load_object_classes()
-    types = OBJECT_CLASSES.copy()
+    # Can we remove this copy?
+    types = get_object_classes().copy()
```
ob-stripe: Yeah, I think you can simply remove the `load_object_classes()` and `get_object_classes()` functions and just use `stripe.object_classes.OBJECT_CLASSES` in place of `types` in the `klass = types.get(...)` call below.

I wouldn't consider the removal of `load_object_classes()` a breaking change. It's technically public (but then again what isn't public in a Python package 🤔) but it's clearly an implementation detail that was never meant to be used by external users. 

We could simply have it return `stripe.object_classes.OBJECT_CLASSES` if you'd rather keep it for now though.

stripe/__init__.py
```diff
@@ -32,12 +32,6 @@
 
 # API resources
 from stripe.api_resources import *  # noqa
-from stripe.api_resources import checkout  # noqa
-from stripe.api_resources import issuing  # noqa
-from stripe.api_resources import radar  # noqa
-from stripe.api_resources import reporting  # noqa
-from stripe.api_resources import sigma  # noqa
-from stripe.api_resources import terminal  # noqa
```
ob-stripe: I think this is okay, but can you check that after importing the library, something like `stripe.issuing.Card` still works?

stripe/util.py
rattrayalex-stripe: cool I'm happy to remove `load_object_classes()`!

I'd rather keep `def get_object_classes()` though just because it's so strange to import within a function like that. Is that okay?

I think I can refactor away the `types` variable and the copy...

rattrayalex-stripe: (I don't think we can `return stripe.object_classes.OBJECT_CLASSES` without circular dep issues... when I tried, I got `AttributeError: module 'stripe' has no attribute 'object_classes'` ... I don't think we really want to deliberately expose `object_classes` on the root stripe object anyway)

stripe/__init__.py
rattrayalex-stripe: it does!

rattrayalex-stripe:
Updated!
ptal @ob-stripe

stripe/__init__.py
ob-stripe: 👍 Thanks!

ob-stripe:
LGTM, but you need to run `make fmt` to fix formatting errors. Feel free to self-approve afterwards.

ptal @rattrayalex-stripe

rattrayalex-stripe:
Oops, sorry I missed that. Done. Also added some reordering to the object_classes for alphabetization (some terminal resources hadn't been properly alphabetized previously). Given the nature of the change, self-approving: 

LGTM

PR #587, opened 2019-06-20
Enable request latency telemetry by default

ob-stripe:
cc @stripe/api-libraries 

Cf. https://github.com/stripe/stripe-ruby/pull/800.


ob-stripe:
Released as 2.31.0.

PR #588, opened 2019-06-21
[codegen] Update API Resources

rattrayalex-stripe:
r? @ob-stripe (feel free to change reviewer)
cc @stripe/api-libraries

- [ ] Sanity check for correctness.
- [ ] Ensure appropriate test coverage (ie; write tests).

### DIFF
--- a/stripe/object_classes.py
+++ b/stripe/object_classes.py
@@ -26,6 +26,7 @@ OBJECT_CLASSES = {
     api_resources.Coupon.OBJECT_NAME: api_resources.Coupon,
     api_resources.CreditNote.OBJECT_NAME: api_resources.CreditNote,
     api_resources.Customer.OBJECT_NAME: api_resources.Customer,
+    api_resources.CustomerBalanceTransaction.OBJECT_NAME: api_resources.CustomerBalanceTransaction,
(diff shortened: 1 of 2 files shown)


stripe/object_classes.py
```diff
@@ -26,6 +26,7 @@
     api_resources.Coupon.OBJECT_NAME: api_resources.Coupon,
     api_resources.CreditNote.OBJECT_NAME: api_resources.CreditNote,
     api_resources.Customer.OBJECT_NAME: api_resources.Customer,
+    api_resources.CustomerBalanceTransaction.OBJECT_NAME: api_resources.CustomerBalanceTransaction,
```
rattrayalex-stripe: err, @ob-stripe , do we need this?

rattrayalex-stripe:
This was a test.

stripe/object_classes.py
remi-stripe: Why wouldn't we need that line?

remi-stripe: oh wow I think it's missing in prod, this is a real bug.

rattrayalex-stripe: yeah I think that's correct 😬

PR #589, opened 2019-06-21
Fix CustomerBalanceTransaction to deserialize properly

remi-stripe:
I forgot to add this object to the list of objects to be deserialized when I rebased on master, this fixes it.

r? @rattrayalex-stripe @ob-stripe 
cc @stripe/api-libraries


rattrayalex-stripe:
LGTM

PR #592, opened 2019-07-07
Fix argument name conflict

### LINKED ISSUE #591: TypeError when updating webhooks
Code is basically:
```
stripe.WebhookEndpoint.modify(hook['id'], url=url)
```
Fails with:
```
  [...snip...]
  File "stripe/api_resources/abstract/updateable_api_resource.py", line 12, in modify
    return cls._static_request("post", url, **params)
TypeError: _static_request() got multiple values for keyword argument 'url'
```

That's because [stripe.api_resources.abstract.api_resource.APIResource._static_request](https://github.com/stripe/stripe-python/blob/master/stripe/api_resources/abstract/api_resource.py#L48) accepts a `url` argument which conflicts with the `url` kwarg required for hooks:
```
>>> def foo(bar, **kwargs):
...     pass
... 
>>> foo('a bar', bar='another bar')
TypeError: foo() got multiple values for keyword argument 'bar'
```

Fwiw, using stripe==2.31.0 on Python 2.7.10

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

The `APIResource._static_request` method defines some regular arguments plus the catch-all `**params` as the dictionary of actual parameters for the API request. The issue is that one cannot pass an API parameter with the same name as a regular argument.

This PR renames the `method` and `url` arguments to `method_` and `url_` respectively to avoid conflicting with actual API parameters.

Fixes #591.

### DIFF
--- a/stripe/api_resources/abstract/api_resource.py
+++ b/stripe/api_resources/abstract/api_resource.py
@@ -44,11 +44,13 @@ class APIResource(StripeObject):
         extn = quote_plus(id)
         return "%s/%s" % (base, extn)
 
+    # The `method_` and `url_` arguments are suffixed with an underscore to
+    # avoid conflicting with actual request parameters in `params`.
     @classmethod
     def _static_request(
         cls,
-        method,
-        url,
+        method_,
+        url_,
         api_key=None,
         idempotency_key=None,
         stripe_version=None,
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released as 2.32.1.

PR #605, opened 2019-08-12
Fix automatic retries of failed requests

Changaco:
Unless I'm missing something, the "less than" comparison here means that the request is retried `n - 1` times instead of `n` times, so setting `max_network_retries` to `1` doesn't work and setting it to `2` only retries once.

### DIFF
--- a/stripe/http_client.py
+++ b/stripe/http_client.py
@@ -493,6 +493,7 @@ class PycurlClient(HTTPClient):
                 "https://twitter.com/stripestatus, or let us know at "
                 "support@stripe.com."
             )
+            should_retry = True
         elif e.args[0] in [pycurl.E_SSL_CACERT, pycurl.E_SSL_PEER_CERTIFICATE]:
             msg = (
                 "Could not verify Stripe's SSL certificate.  Please make "
@@ -500,14 +501,16 @@ class PycurlClient(HTTPClient):
                 "If this problem persists, let us know at "
                 "support@stripe.com."
             )
+            should_retry = False
         else:
             msg = (
                 "Unexpected error communicating with Stripe. If this "
                 "problem persists, let us know at support@stripe.com."
             )
+            should_retry = False
 
         msg = textwrap.fill(msg) + "\n\n(Network error: " + e.args[1] + ")"
-        raise error.APIConnectionError(msg)
+        raise error.APIConnectionError(msg, should_retry=should_retry)
 
     def _get_proxy(self, url):
         if self._proxy:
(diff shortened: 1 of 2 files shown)


stripe/http_client.py
```diff
@@ -494,21 +494,24 @@ def _handle_request_error(self, e):
...
             )
+            should_retry = True
         elif e.args[0] in [pycurl.E_SSL_CACERT, pycurl.E_SSL_PEER_CERTIFICATE]:
             msg = (
                 "Could not verify Stripe's SSL certificate.  Please make "
                 "sure that your network is not intercepting certificates.  "
                 "If this problem persists, let us know at "
                 "support@stripe.com."
             )
+            should_retry = False
```
brandur-stripe: I don't think we should retry this one — if the SSL certificate couldn't be verified on the first request, it's not likely to be verified on the next either.

brandur-stripe:
@ob-stripe You're more familiar with the code here, so will leave to you. It does look like something is off — we increment the number of retries on the first request rather than waiting for an actual retry. The test _looks_ right, but I think it's measuring retries including the original request, which is why it works.

ob-stripe:
Ah yeah, looks like `num_retries` is incremented before the first request.

I think a slightly better fix would be to keep the comparison as is and move the `num_retries` increment from [here](https://github.com/stripe/stripe-python/blob/95166ac1c0532294ba7433f6d975cb7df878d302/stripe/http_client.py#L122) to [here](https://github.com/stripe/stripe-python/blob/95166ac1c0532294ba7433f6d975cb7df878d302/stripe/http_client.py#L135), because the first request is not a "retry" and should not be counted as such. It would also be slightly more consistent with the stripe-ruby implementation (see [here](https://github.com/stripe/stripe-ruby/blob/5b5b0f5722d89e94c0b0b34bd2b6c475347ddb79/lib/stripe/stripe_client.rb#L293)).

brandur-stripe:
> I think a slightly better fix would be to keep the comparison as is and move the num_retries increment from here to here, because the first request is not a "retry" and should not be counted as such. It would also be slightly more consistent with the stripe-ruby implementation (see here).

+1. Going off variable names, the code is pretty confusing at right now which is probably how this bug came in in the first place.

Changaco:
> I think a slightly better fix would be to keep the comparison as is and move the `num_retries` increment

Done.

stripe/http_client.py
Changaco: We're not retrying, `should_retry` is set to `False` here, not `True`.

Changaco:
Ping.

ob-stripe:
Released as 2.35.1.

PR #606, opened 2019-08-12
Clarify what `max_network_retries` does

Changaco:
It's not limited to network errors, it also triggers a retry when a 409 HTTP response is returned by the API.


Changaco:
@brandur-stripe I would rather not drop the second paragraph that this PR was made to add, so I reworded it instead. Let me know if it's still not good enough.

brandur-stripe:
Okay that works.

PR #607, opened 2019-08-12
[codegen] Add `SubscriptionItem.create_usage_record` method

ob-stripe:
r? @rattrayalex-stripe 
cc @stripe/api-libraries

- [x] Sanity check for correctness.
- [x] Ensure appropriate test coverage (ie; write tests).

### DIFF
--- a/stripe/api_resources/subscription_item.py
+++ b/stripe/api_resources/subscription_item.py
@@ -4,8 +4,10 @@ from stripe.api_resources.abstract import CreateableAPIResource
 from stripe.api_resources.abstract import DeletableAPIResource
 from stripe.api_resources.abstract import ListableAPIResource
 from stripe.api_resources.abstract import UpdateableAPIResource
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released as 2.35.0.

rogerpine:
Is action='set' supported in create_usage_record? I only get the increment behavior. Set is meant to overwrite the full usage amount during the billing period.
Thank you.

ob-stripe:
@rogerpine The library doesn't know about specific parameters, it simply passes parameters along to Stripe's API. So all parameters should be supported.

Are you observing an unexpected behavior when passing `action='set'`?

rogerpine:
Yes, when I passed action='set', it was continuing to increment.

remi-stripe:
@rogerpine `set` only works if you re-use the same timestamp for all your UsageRecords. If you pass a different timestamp each time it will still sum up to the total.

I'd recommend reaching out to our support team to discuss this instead: https://support.stripe.com/contact

PR #610, opened 2019-09-09
Add support for header parameters in `ListObject` request methods

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

`ListObject` provides three instance methods for issuing API requests on the list: `create`, `retrieve` and `list`.

Previously none of these supported "special" parameters for headers: `api_key`, `stripe_account`, `stripe_version`, and `idempotency_key` (for `create` only as the other two are GET requests). This PR fixes that.

The code is very much not DRY, but it's consistent with the rest of the library. The requestor layer in stripe-python is starting to show its age, we should refactor it at some point.

Fixes #609.

PR #612, opened 2019-09-11
Use ListObject properties as default values in request methods

### LINKED ISSUE #611: api_key param no longer passed to nested objects?
Hi, I'm hitting an issue with 2.36.0, it seems that the api_key param isn't passed to `Customer.sources` (and presumably other nested objects?).

Eg, this works with 2.35.1:

```python
c = stripe.Customer.retrieve(id="cus_xxxx", api_key="sk_test_123") 
c.sources.create(source="tok_visa")
```

But fails with 2.36.0

```
AuthenticationError: No API key provided. (HINT: set your API key using "stripe.api_key = <API-KEY>"). You can generate API keys from the Stripe web interface.  See https://stripe.com/api for details, or email support@stripe.com if you have any questions.
```

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

I caused a regression in #610. Previously, when calling a request method on a `ListObject`, the `ListObject`'s properties for headers (`api_key`, `stripe_account`, `stripe_version`) would be used.

In #610, I added parameters for headers to request methods, but if the parameters are not passed, we would default to `None` instead of using the `ListObject`'s properties for the headers.

This PR fixes this by using the parameter if provided, and falling back to the `ListObject` properties if not.

Fixes #611.

### DIFF
--- a/stripe/api_resources/list_object.py
+++ b/stripe/api_resources/list_object.py
@@ -12,13 +12,13 @@ class ListObject(StripeObject):
     def list(
         self, api_key=None, stripe_version=None, stripe_account=None, **params
     ):
-        requestor = api_requestor.APIRequestor(
-            api_key, api_version=stripe_version, account=stripe_account
-        )
-        url = self.get("url")
-        response, api_key = requestor.request("get", url, params)
-        stripe_object = util.convert_to_stripe_object(
-            response, api_key, stripe_version, stripe_account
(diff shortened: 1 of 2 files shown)


therefromhere:
I can confirm this fixes the issue for me, thanks.

ob-stripe:
Released as 2.36.1.

PR #616, opened 2019-09-13
Pass CFLAGS and LDFLAGS when running tests

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Use the `LDFLAGS` and `CFLAGS` environment variables when running tests.

Context for why this is useful:
- with Python 2, we install the `requests[security]` meta-package, which contains the `cryptography` package: https://github.com/stripe/stripe-python/blob/c5b9414b72ad298558b657b5990502a830fa5798/setup.py#L54
- PyPy 2 cannot use the prebuilt wheel and has to compile from source
- In order for the compilation to succeed on macOS, you need to pass the right `LDFLAGS` and `CFLAGS` value: https://cryptography.io/en/latest/installation/#building-cryptography-on-macos

With this PR, I can run this command on my laptop:

```sh
$ LDFLAGS="-L$(brew --prefix openssl@1.1)/lib" \
    CFLAGS="-I$(brew --prefix openssl@1.1)/include" \
    make test
```

and verify that tests pass with PyPy 2.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -17,6 +17,12 @@ skip_missing_interpreters = true
 [testenv]
 description = run the unit tests under {basepython}
 commands = python setup.py test -a "{posargs:-n auto}"
+# compilation flags can be useful when prebuilt wheels cannot be used, e.g.
+# PyPy 2 needs to compile the `cryptography` module. On macOS this can be done
+# by passing the following flags:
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Nice.

Would you mind just putting a brief comment on that line saying that it's there to enable successful crypto builds on Mac OS? Otherwise, LGTM.

ptal @ob-stripe

PR #619, opened 2019-09-24
Add ErrorObject to StripeError exceptions

ob-stripe:
cc @stripe/api-libraries 

Similar to https://github.com/stripe/stripe-ruby/pull/811 and https://github.com/stripe/stripe-php/pull/705.

### DIFF
--- a/stripe/api_resources/error_object.py
+++ b/stripe/api_resources/error_object.py
@@ -1,0 +1,69 @@
+from __future__ import absolute_import, division, print_function
+
+from stripe.util import merge_dicts
+from stripe.stripe_object import StripeObject
+
+
+class ErrorObject(StripeObject):
+    def refresh_from(
+        self,
+        values,
+        api_key=None,
+        partial=False,
+        stripe_version=None,
+        stripe_account=None,
+        last_response=None,
+    ):
+        # Unlike most other API resources, the API will omit attributes in
+        # error objects when they have a null value. We manually set default
+        # values here to facilitate generic error handling.
+        values = merge_dicts(
+            {
+                "charge": None,
+                "code": None,
+                "decline_code": None,
+                "doc_url": None,
+                "message": None,
+                "param": None,
+                "payment_intent": None,
+                "payment_method": None,
+                "setup_intent": None,
+                "source": None,
+                "type": None,
+            },
+            values,
+        )
+        return super(ErrorObject, self).refresh_from(
+            values,
+            api_key,
+            partial,
+            stripe_version,
+            stripe_account,
+            last_response,
+        )
+
+
+class OAuthErrorObject(StripeObject):
+    def refresh_from(
+        self,
+        values,
+        api_key=None,
+        partial=False,
+        stripe_version=None,
+        stripe_account=None,
+        last_response=None,
+    ):
+        # Unlike most other API resources, the API will omit attributes in
+        # error objects when they have a null value. We manually set default
+        # values here to facilitate generic error handling.
+        values = merge_dicts(
+            {"error": None, "error_description": None}, values
+        )
+        return super(OAuthErrorObject, self).refresh_from(
+            values,
+            api_key,
+            partial,
+            stripe_version,
+            stripe_account,
+            last_response,
+        )
--- a/stripe/error.py
+++ b/stripe/error.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
+import stripe
 from stripe.six import python_2_unicode_compatible
 
 
@@ -32,6 +33,7 @@ class StripeError(Exception):
         self.headers = headers or {}
         self.code = code
         self.request_id = self.headers.get("request-id", None)
+        self.error = self.construct_error_object()
 
     def __str__(self):
         msg = self._message or "<empty message>"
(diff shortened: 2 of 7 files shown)


ob-stripe:
Ugh there's a circular import issue. Will fix.

stripe/api_resources/error_object.py
```diff
@@ -0,0 +1,69 @@
...
+                "code": None,
+                "decline_code": None,
+                "doc_url": None,
+                "message": None,
+                "param": None,
+                "payment_intent": None,
+                "payment_method": None,
+                "setup_intent": None,
+                "source": None,
+                "type": None,
```
remi-stripe: Do we do the same in ruby? That feels like something we'll never remember to fill out unless it could be autogenerated from the spec

stripe/error.py
```diff
@@ -56,6 +59,14 @@ def __repr__(self):
             self.request_id,
         )
 
+    def construct_error_object(self):
+        if self.json_body is None or "error" not in self.json_body:
```
remi-stripe: What does that protect us from exactly?

stripe/api_resources/error_object.py
ob-stripe: We do something different but with the same goal in mind in stripe-ruby: https://github.com/stripe/stripe-ruby/blob/3cb9ad7fca5bfd372006d551d1ecc1089d3739aa/lib/stripe/error_object.rb#L8-L77

That approach wouldn't work in stripe-python because accessing a key that does not exist in Ruby returns `nil`, but in Python it raises a `KeyError`.

The approach here is the same one as in stripe-php: https://github.com/stripe/stripe-php/blob/9a58f033d9d126c3785f0d8ad47f2e1db269864a/lib/ErrorObject.php#L144-L160.

We can look at autogenerating these classes in all languages if maintenance becomes an issue.

stripe/error.py
ob-stripe: The first part is just run of the mill defensive programming, in case the `StripeError` is constructed manually, or if for some reason the API replied with a non-JSON body.

The second part is because for standard API errors (not OAuth errors), the error object is nested under the `"error"` key:

```json
{
  "error": {
    "type": "foo",
    "message": "Foo."
  }
}
```

so we're constructing the `ErrorObject` from the value of the `"error"` key instead of the top-level JSON object.

stripe/api_resources/error_object.py
remi-stripe: Ack. I just worry we will never remember this exists and the `error` object is being changed liberally lately.

ob-stripe:
Circular import should be fixed, ptal @remi-stripe.

stripe/util.py
```diff
@@ -205,6 +205,12 @@ def populate_headers(idempotency_key):
     return None
 
 
+def merge_dicts(x, y):
+    z = x.copy()
+    z.update(y)
+    return z
+
+
```
remi-stripe: How is that not a helper function in Python already 🙀

tests/test_error.py
```diff
@@ -52,6 +52,14 @@ def test_repr(self):
...
             )
 
+    def test_error_object(self):
+        err = error.StripeError(
+            "message", json_body={"error": {"code": "some_error"}}
+        )
+        assert err.error is not None
+        assert err.error.code == "some_error"
+        assert err.error.charge is None
+
```
remi-stripe: Is it worth adding tests for all 7 params in case we break something?

ob-stripe:
Released as 2.37.0.

tests/test_error.py
```diff
@@ -52,6 +52,14 @@ def test_repr(self):
                 "request_id='123')"
             )
 
+    def test_error_object(self):
+        err = error.StripeError(
+            "message", json_body={"error": {"code": "some_error"}}
+        )
```
thieman: @remi-stripe This PR was a breaking change for us since we had code similar to this in our own test suite. The difference was our `json_body.error` was a string, not an object:

```
stripe.error.StripeError('hi', json_body={"error": "Test json_body error"})
```

Starting on 2.37.0 this fails with the following:

```
    json_body={"error": "Test json_body error"},
/usr/local/lib/python3.7/site-packages/stripe/error.py:140: in __init__
    message, http_body, http_status, json_body, headers, code
/usr/local/lib/python3.7/site-packages/stripe/error.py:36: in __init__
    self.error = self.construct_error_object()
/usr/local/lib/python3.7/site-packages/stripe/error.py:66: in construct_error_object
    self.json_body["error"], stripe.api_key
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    @classmethod
    def construct_from(
        cls,
        values,
        key,
        stripe_version=None,
        stripe_account=None,
        last_response=None,
    ):
        instance = cls(
>           values.get("id"),
            api_key=key,
            last_response=last_response,
        )
E       AttributeError: 'str' object has no attribute 'get'
```

Just an FYI, not sure whether you'd like to fix this or publish a notice in the CHANGELOG or something. But the gist is that we updated a minor version and it was a breaking change for us.

thieman: Sorry, pinged the wrong person @ob-stripe

ob-stripe: @thieman Thanks for the feedback. I think we can fix with a slightly stricter check (i.e. check that `json_body['error']` is a `dict` before trying to use it to construct the `StripeError`).

ob-stripe: Should be fixed in 2.37.1.

PR #620, opened 2019-09-26
Check that `error` is a dict before trying to use it to create a `StripeError`

ob-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Check that `error` in the JSON data is a dict before trying to use it to initialize the `StripeError` instance. (This should never happen in production, but can happen in unit tests where `StripeError`s are manually constructed.)

Addresses this comment from @thieman: https://github.com/stripe/stripe-python/pull/619#discussion_r328761227

### DIFF
--- a/stripe/error.py
+++ b/stripe/error.py
@@ -59,7 +59,11 @@ class StripeError(Exception):
         )
 
     def construct_error_object(self):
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released as 2.37.1.

PR #623, opened 2019-10-12
Add support for reverse pagination

### LINKED ISSUE #585: Feature request: auto_paging_iter() using ending_before
It would be useful to support paginating backwards using `auto_paging_iter()` on listable resources. Currently, it assumes you're paginating forwards using `starting_after`. If you try to use it with `ending_before`, the library will attempt to use `starting_after` anyways on the last item's `id` and error out.

As a side note on consistency, the API returns an error message like "Received both starting_after and ending_before parameters. Please pass in only one." The documentation at https://stripe.com/docs/api/pagination claims that only ending_before is used in this case (although I think it's good that this is not the actual behavior).

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Add support for backwards auto-pagination, and generally improves `ListObject`'s code by making it closer to the Ruby / PHP implementations.

Fixes #585.

### DIFF
--- a/stripe/api_resources/list_object.py
+++ b/stripe/api_resources/list_object.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import, division, print_function
 
-from stripe import api_requestor, util
+from stripe import api_requestor, six, util
 from stripe.stripe_object import StripeObject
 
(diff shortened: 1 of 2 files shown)


ob-stripe:
Released as 2.38.0.

PR #627, opened 2019-11-08
Add list_usage_record_summaries and list_source_transactions

rattrayalex-stripe:
? @ob-stripe 
cc @stripe/api-libraries 

Used codegen for this (PR forthcoming), wrote tests by hand.

### DIFF
--- a/stripe/api_resources/source.py
+++ b/stripe/api_resources/source.py
@@ -35,8 +35,9 @@ class Source(CreateableAPIResource, UpdateableAPIResource, VerifyMixin):
                 "id",
             )
 
-    # DEPRECATED: use Source.list_source_transactions instead.
     def source_transactions(self, **params):
+        """source_transactions is deprecated, use Source.list_source_transactions instead.
+        """
         return self.request(
             "get", self.instance_url() + "/source_transactions", params
         )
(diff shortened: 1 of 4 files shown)


stripe/api_resources/source.py
```diff
@@ -33,6 +35,7 @@ def detach(self, idempotency_key=None, **params):
                 "id",
             )
 
+    # DEPRECATED: use Source.list_source_transactions instead.
```
ob-stripe: Nit: users would not see this unless they go through the trouble of reading the library's source code. The idiomatic way of marking deprecated functions in Python is to use the `warnings` module:

```python
import warnings

class Source(...):
    ...

    # TODO: remove this in next major version
    def source_transactions(self, **params):
        warnings.warn(
            "source_transactions is deprecated, use list_source_transactions instead",
             DeprecationWarning
        )
        return self.request(...)
```

We could also add a Python docstring to display the deprecation in most IDEs:

```python
    def source_transactions(self, **params):
        '''source_transactions is deprecated, use list_source_transactions instead
        '''
        ...
```

rattrayalex-stripe: Gotcha – will definitely use a docstring then. Thanks!
I'd lean away from a warning (feels like overkill, users will see the changelog when it comes time to do a major upgrade, and it'll be an easy transition) unless you feel strongly the other way?

ob-stripe: Works for me 👍

ob-stripe: Please do add an internal comment with a `TODO` so we can grep for it and remember to remove it in the next major version!

rattrayalex-stripe: Hmm, would we not grep for "deprecated"?

rattrayalex-stripe:
r? @ob-stripe 

Happy to add the comment too if you like, just feels redundant

PR #631, opened 2019-11-27
Update proxy.py

maxmokry:
if using stripe.proxy api requestor show warning "UserWarning: stripe.proxy was updated after sending a request - this is a no-op. To use a different proxy, set stripe.default_http_client to a new client configured with the proxy."


ob-stripe:
Thanks!

PR #636, opened 2019-12-30
Fix uploading files with Unicode names (Python 2.7)

illia-v:
It resulted in `UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-6: ordinal not in range(128)` before


ob-stripe:
Released as 2.41.1.

PR #640, opened 2020-01-14
Add support for `CreditNoteLineItem`

remi-stripe:
Codegen for openapi d663cdb
No method added in Python for now to be consistent with InvoiceLineItem.

r? @ob-stripe 
cc @stripe/api-libraries


ob-stripe:
Released as 2.42.0.

PR #646, opened 2020-03-23
Allow overriding API key in OAuth methods

### LINKED ISSUE #645: stripe.OAuth.token doesn't offer a way to override global api_key
Unlike most of the stripe object methods, [stripe.OAuth.token](https://github.com/stripe/stripe-python/blob/598566add896a571088c76acc073a8243afbbfb2/stripe/oauth.py#L39-L43) doesn't offer a way to pass in an `api_key` to override the global API key.

This capability is necessary to generate test access tokens from a production refresh token: https://stripe.com/docs/connect/testing#test-keys-for-livemode-applications

ob-stripe:
r? @brandur-stripe @richardm-stripe 
cc @stripe/api-libraries 

Allow overriding the global API key in OAuth methods.

Fixes #645.

### DIFF
--- a/stripe/oauth.py
+++ b/stripe/oauth.py
@@ -37,14 +37,18 @@ class OAuth(object):
         return url
 
     @staticmethod
-    def token(**params):
-        requestor = api_requestor.APIRequestor(api_base=connect_api_base)
+    def token(api_key=None, **params):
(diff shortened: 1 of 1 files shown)


ob-stripe:
Released as 2.44.0.

Issue #653, opened 2020-05-07
Clarify response type and handling of errors in the README?


mrcoles:
It’d be helpful if the README had info on the response object type and handling error responses. It seems that trying it in a Python REPL is the best way, but it’s still unexpected that they’re both objects with attributes and dicts.

**Response type:**

If the response is a success, then the response is always an instance of `stripe.stripe_object.StripeObject` or the subclass `stripe.api_resources.list_object.ListObject`, which is a dict or list matching the response body, but it has a couple other bells and whistles, including the ability to access and set values via attributes (instead of dictionary item lookup). It’s possible I have a blind spot in the reading the docs, but I don’t see this explained in the [Python API docs](https://stripe.com/docs/api?lang=python).

**Error handling:**

This one is covered really well in the sample code in the [Python API docs “Handling Errors” section](https://stripe.com/docs/api/errors/handling?lang=python), but a little redundancy having it in the README (or just a clear deep-link to that section in the docs) would be nice.

I think this would help people feel more confident using the library. Regardless, thanks for providing this and keep up the great work with the client and API!

Loosely related issues:

*   Are stripe objects dicts? https://github.com/stripe/stripe-python/issues/494
*   Type annotations feature request https://github.com/stripe/stripe-python/issues/650

remi-stripe:
@mrcoles Sorry for the delays here and thank you for providing detailed feedback! We definitely could improve the readme to make it clearer how to use stripe-python directly! We'll work on it next week :)

PR #656, opened 2020-05-12
README: response handling

richardm-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Attempt at clarifying how to handle responses, as suggested by @mrcoles in #653

I considered explaining
> If the response is a success, then the response is always an instance of stripe.stripe_object.StripeObject or the subclass stripe.api_resources.list_object.ListObject

as @mrcoles suggests,

But the details here are pretty complicated and it's been a challenge to describe them in a way that isn't either misleading or potentially confusing. My *hope* is that accessing some properties on the example will show users how they can grab properties, and will be enough of a hint to discerning users that these aren't plain dicts as you might otherwise expect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,11 +43,25 @@ import stripe
 stripe.api_key = "sk_test_..."
 
 # list customers
-stripe.Customer.list()
+customers = stripe.Customer.list()
 
-# retrieve single Customer
-stripe.Customer.retrieve("cus_123456789")
+# print the first customer's email
+print(customers.data[0].email)
+
+# retrieve specific Customer
+customer = stripe.Customer.retrieve("cus_123456789")
+
+# print that customer's email
+print(customer.email)
 ```
+
+### Handling exceptions
+
+Unsuccessful requests raise exceptions. The class of the exception will reflect
+the sort of error that occurred. Please see the [Api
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -43,12 +43,26 @@ import stripe
...
-# retrieve single Customer
-stripe.Customer.retrieve("cus_123456789")
+# print the first customer's email
+print(customers.data[0].email)
+
+# retrieve specific Customer
+customer = stripe.Customer.retrieve("cus_123456789")
+
+# print that customer's email
+print (customer.email)
```
remi-stripe, line 55: extra space

richardm-stripe:
ptal @remi-stripe

mrcoles:
Thanks, @richardm-stripe, this definitely improves the README!

I personally still like to have a little extra info about the specific response type (especially since Python doesn’t give you any additional hints in the code)—in this case a `stripe.stripe_object.StripeObject` which is the raw response dict or list, but it also allows you to read and set the data as attributes (and then other stuff that I haven’t really looked into, like the `request` method?) However, any improvement such as the one here makes me happy, and the more curious coder can always dive into REPL and source code.

PR #678, opened 2020-09-07
validate default value

ShreehariVaasishta:
Raise an error if stripe api key is None in [webhook.py](https://github.com/stripe/stripe-python/blob/master/stripe/webhook.py#L18)

### DIFF
--- a/stripe/webhook.py
+++ b/stripe/webhook.py
@@ -17,6 +17,9 @@ class Webhook(object):
     def construct_event(
         payload, sig_header, secret, tolerance=DEFAULT_TOLERANCE, api_key=None
     ):
+        if api_key is None:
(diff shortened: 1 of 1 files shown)


stripe/webhook.py
```diff
@@ -17,6 +17,9 @@ class Webhook(object):
     def construct_event(
         payload, sig_header, secret, tolerance=DEFAULT_TOLERANCE, api_key=None
     ):
+        if api_key is None:
+            raise ValueError("api key is None")
+
```
remi-stripe, line 22: @ShreehariVaasishta The API key should not be needed for a webhook handler so I'm not entirely sure why this change is necessary. Can you clarify why you are raising an error here and what issue you are fixing with that change?

PR #688, opened 2020-10-28
Add constants with all possible invoice's statuses

wreckah:
It allows to use these constants via autocomplete suggestions and avoid of typos.

### DIFF
--- a/stripe/api_resources/invoice.py
+++ b/stripe/api_resources/invoice.py
@@ -22,6 +22,12 @@ class Invoice(
 ):
     OBJECT_NAME = "invoice"
 
+    STATUS_DRAFT = "draft"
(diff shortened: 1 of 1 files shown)


remi-stripe:
@wreckah Thanks for the submission and sorry for the delay. I'll make sure someone on the team has a look and follows up with more details.

richardm-stripe:
Hi @wreckah, thanks for the contribution. I think if we were to merge this we would probably want to do it for all enums in the API, like we do with [stripe-php](https://github.com/stripe/stripe-php/blob/master/lib/Invoice.php#L130-L135), so I'm closing this PR for now, but I've taken the liberty of opening an Issue with this labeled as a feature request. https://github.com/stripe/stripe-python/issues/689

PR #696, opened 2020-11-25
Update to_dict_recursive to work with lists of StripeObjects #695

soltanianalytics:
See https://github.com/stripe/stripe-python/issues/695

### DIFF
--- a/stripe/stripe_object.py
+++ b/stripe/stripe_object.py
@@ -284,9 +284,7 @@ class StripeObject(dict):
                 d[k] = v.to_dict_recursive()
             if isinstance(v, list):
                 d[k] = [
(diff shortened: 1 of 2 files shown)


paulasjes-stripe:
Hi @soltanianalytics, thanks for opening this! Would you be able to add some tests that cover this edge case?

soltanianalytics:
Added a test, not sure why it's failing

ob-stripe:
Hi @soltanianalytics, thanks for the contribution! I went with a slightly different approach to keep the implementation consistent with what we do in other languages, but the behavior should be the same. I hope you don't mind!

soltanianalytics:
Happy to see that it was useful!

PR #710, opened 2021-02-11
github actions

richardm-stripe:
r? @ob-stripe 
Runs `make lint`, `make fmt`, and `make ci` in Github Actions.
Doesn't include support for "coveralls" but I think we've agreed to drop it.


ob-stripe (approved):
LGTM!

PR #711, opened 2021-02-17
Deprecate `python setup.py test`

ob-stripe:
r? @richardm-stripe 

Using `python setup.py test` is deprecated and produces this warning:
```
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
```

This PR removes support for `python setup.py test` and instead uses tox as the test entry point as recommended by the warning.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,12 +17,12 @@ test: venv
 test-nomock: venv
 	@${VENV_NAME}/bin/tox -p auto -- --nomock $(TOX_ARGS)
 
-ci: venv
-	@${VENV_NAME}/bin/python setup.py test -a "-n auto --cov=stripe"
+test-travis: venv
+	${VENV_NAME}/bin/python -m pip install -U tox-travis
+	@${VENV_NAME}/bin/tox -p auto $(TOX_ARGS)
 
 coveralls: venv
-	@${VENV_NAME}/bin/$(PIP) install --upgrade coveralls
-	@${VENV_NAME}/bin/coveralls
+	@${VENV_NAME}/bin/tox -e coveralls
 
 fmt: venv
 	@${VENV_NAME}/bin/tox -e fmt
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -17,8 +17,9 @@ test: venv
 test-nomock: venv
 	@${VENV_NAME}/bin/tox -p auto -- --nomock $(TOX_ARGS)
 
-ci: venv
-	@${VENV_NAME}/bin/python setup.py test -a "-n auto --cov=stripe"
+test-travis: venv
+	${VENV_NAME}/bin/python -m pip install -U tox-travis
```
ob-stripe, line 21: [`tox-travis`](https://github.com/tox-dev/tox-travis) is a helper module to ensure that tox will only run tests using the Python version from the current Travis VM.

tests/test_api_requestor.py
```diff
@@ -12,7 +12,7 @@
 from stripe import six
 from stripe.stripe_response import StripeResponse
 
-from six.moves.urllib.parse import urlsplit
+from stripe.six.moves.urllib.parse import urlsplit
```
ob-stripe, line 15: This was an actual bug in tests! The previous version used the `six` module installed in the venv, not the one bundled with stripe-python. Now tox creates a new venv for each test run and `six` is not installed in that venv, so this started failing.

ob-stripe:
Ah, it looks coverage is broken. coveralls report 0% coverage. 👀

ob-stripe:
Alright, coverage support is fixed. I had to make a few changes based on https://github.com/pytest-dev/pytest-cov/tree/master/examples/src-layout and make the coveralls upload run from tox as well, but that's probably better anyway.

ptal @richardm-stripe

richardm-stripe (approved):
These changes look good to me.

PR #717, opened 2021-03-21
Fix test TestCharge.test_is_saveable().

andrewnguonly:


yejia-stripe (approved):
@andrewnguonly Thank you for contributing to the Stripe Python library! This change looks good to me! And sorry we took so long to get back to you about it.

PR #719, opened 2021-05-02
Set daemon attribute instead of using setDaemon method that was deprecated in Python 3.10

### LINKED ISSUE #718: threading.Thread.setDaemon has been deprecated in favor of setting daemon attribute directly in Python 3.10
Ref : python/cpython#25174

https://github.com/stripe/stripe-python/blob/e70949b8b6961907620f01aaea60ecea9759558f/tests/test_integration.py#L64

tirkarthi:
Fixes #718


stianjensen:
Looks good. Since python 2.7 is the lowest target of this package, this should be safe to merge (supported since 2.6)

dcr-stripe (approved):
Thanks for the contribution here @tirkarthi and apologies for the extremely long delay.

This looks good. CI is passing and https://docs.python.org/2/library/threading.html confirms we can do this as of 2.6 + that `setDaemon` is the pre-2.6 API. Merging and this will go out with the next release.

PR #730, opened 2021-07-28
Add support for SearchResult.

dcr-stripe:
### Notify

r? @richardm-stripe 

### Summary

Adds support for `search_result` objects returned by the Stripe API, including adding a `SearchResultObject` class and a `SearchableAPIResource` class which can be added to resources to support invoking search APIs.

This behaves similarly to `ListObject` in that they both wrap around a list of objects and provide pagination. However the `SearchResultObject` object paginates by relying on a `next_page` token included in the response rather than using object IDs + `starting_before`/`ending_after`. Thus, only forward pagination is supported

### Test plan

Added unit tests. We'll want to add some more rigorous tests once stripe-mock has been updated to include at least one API that returns this resource.


dcr-stripe:
cc @susindaran-stripe

PR #736, opened 2021-10-07
Stop sending raw exception message as part of Stripe user agent.

dcr-stripe:
### Notify

r? @richardm-stripe 
cc @remi-stripe 

### Summary

Updates user agent generation to avoid sending the raw exception message when calling platform methods to determine user agent components. Instead just send `"(disabled)"` for the failed fields.

### Test plan

Added unit test.

### DIFF
--- a/stripe/api_requestor.py
+++ b/stripe/api_requestor.py
@@ -257,7 +257,7 @@ class APIRequestor(object):
             try:
                 val = func()
             except Exception:
-                val = None
+                val = "(disabled)"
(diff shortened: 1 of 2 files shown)


richardm-stripe:
`(disabled)` is what we send in stripe-php: https://github.com/stripe/stripe-php/blob/33317c93fb3634039dcf908e5b3a60c39769d1ff/lib/ApiRequestor.php#L332

Issue #755, opened 2021-12-20
Customer.list_payment_methods with auto_paging_iter does not respect initial parameters


AndreasHogstrom:
Similarly to this old issue; https://github.com/stripe/stripe-python/issues/221 , the `Customer.list_payment_methods` used in conjunction with `auto_paging_iter` does not persist any parameters when going to the next page.

Python version: 3.9.7
Stripe SDK version: 2.63.0

Example:
Both the `type` parameter and the customer ID is lost:
```python
# Generated URL: /v1/customers/cus_123/payment_methods?type=card
payment_methods = stripe.Customer.list_payment_methods("cus_123", type="card")

# Next page generated URL: /v1/customers/:customer/payment_methods?starting_after=pm_123
for payment_method in payment_methods.auto_paging_iter():
    pass
```

dcr-stripe:
Thanks for the report @AndreasHogstrom - I can reproduce this as well. The team is looking into fixing this and will report back.

dcr-stripe:
Fixes have been merged for this to pass the parameters forward. Thanks again for the report - great find.

 Unfortunately there still remains an issue with the `/payment_methods` endpoint where it returns `/v1/customers/:customer/payment_methods` in the `url` field instead of `/v1/customers/cus_123/payment_methods`. This will need to be fixed independently.

I'll keep this issue open until we've released a version of stripe-python with these fixes.

dcr-stripe:
https://github.com/stripe/stripe-python/releases/tag/v2.64.0 has been released including this fix.

As I mentioned, the `/payment_methods` endpoint is still experiencing a bug, but that is being worked on independently from the SDKs.

I'll close this issue since the initial parameters are now being respected, but I'll make sure to update here as well once the `/payment_methods` bug is fixed.

PR #756, opened 2021-12-20
Introduce custom listing methods on objects.

dcr-stripe:
r? @yejia-stripe 

### Summary

Updates the implementation for custom methods which return list objects to:

1. Create a new StripeObject rather than attempting to refresh - this is necessary as the returned object will be of a different type (now `list`) and fails to refresh.
2. Set `_retrieve_params` to make sure iteration continues with the same input set of parameters.

### Motivation

Partially fixes https://github.com/stripe/stripe-python/issues/755. We still need to do this for the static custom methods.

### Test plan

I added a test which previously failed and now passes.


yejia-stripe (approved):
Lgtm!

PR #761, opened 2022-01-08
Add Checkout Session Example

elibroftw:
Also, I strongly advise Stripe to delete `"editor.renderWhitespace": "all",` from `.vscode/settings.json` as it is irrelevant

### DIFF
--- a/examples/checkout.py
+++ b/examples/checkout.py
@@ -28,6 +28,7 @@ def load_dot_env():
                     key, value = line.split('=', 1)
                     os.environ[key] = value
 
+
 load_dot_env()
 stripe.api_key = os.environ['STRIPE_SECRET_KEY']
 
--- a/examples/checkout.py
+++ b/examples/checkout.py
@@ -1,0 +1,81 @@
+"""
+This is a thorough example using Stripe's Checkout Session API for customers paying with a credit card.
+    Using checkouts, Stripe handles the collection and processing of sensitive customer data,
+    meaning that you don't have to worry about PCI compliance, and customers don't have to trust you to secure their data.
+Relevant docs: https://stripe.com/docs/api/checkout/sessions
+Video tutorial: https://www.youtube.com/watch?v=8y9RNFc71DA
+
+Get API keys: https://dashboard.stripe.com/apikeys
+Credit card acceptance/blocking rules: https://dashboard.stripe.com/settings/radar/rules
+"""
+import time
+# 3rd party
+import stripe
+# not required in server code
+import webbrowser
+import os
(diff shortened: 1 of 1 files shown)


dcr-stripe:
Thanks for the contribution @elibroftw. It's really appreciated!

We currently have a number of examples that cover checkout both in the [docs](https://stripe.com/docs/api/checkout/sessions) and in the [stripe-samples](https://github.com/search?q=topic%3Astripe-checkout+org%3Astripe-samples&type=Repositories) repository. Particularly https://github.com/stripe-samples/checkout-one-time-payments/blob/master/server/python/server.py overlaps with this and has implementations for all of our languages.

Given that this is already covered by those samples and would add a non-trivial example that would require ongoing maintenance, I'll close without merging. Please feel free to re-open if you think the existing samples don't cover this flow.

Thanks!

elibroftw:
1. I read the docs to create my example. It took hours because I first had to figure out what it is that I was looking for (Sessions) and then I needed to figure out how to use it without requiring a price id. I don't want price IDs because my store is dynamic. 
2. The stripe-samples repository is not mentioned anywhere in this repo's `README.md`. If I missed it, so would many developers. 
3. The file you linked is using checkouts from one developers PoV. A generalized example is much easier to work with and won't impose as much opinions onto other devs. For example, in the session example, upon an error, customers are shown the actual Python exception! That is definitely not the way to handle an exception in a server route. 
4. In the sample you provided, there is no mention of checkout failures, or how the webhook is used. Developers would have to infer from that server sample that Stripe will send a request to the webhook once the client has made a payment. This was also not mentioned in the docs so that's why my example checks the payment manually rather than making a server example.
5. I almost forgot. There is no mention of testing failures. 
5. If maintenance is a concern, you can just fix the API version to what is is currently.

I'll end off with saying that this example is not for Stripe, it's for other developers who use Stripes' products.  I didn't make it for the heck of making it, it comes from actual experience. I made it because I personally sunk too much time for a such a simple task (dynamic checkouts).

elibroftw:
I definitely don't feel appreciated btw. Every reply takes 10s of minutes to make and what's worse is that the reward is almost none.

dcr-stripe:
Thanks for the examples of documentation gap.

Currently https://stripe.com/docs/payments is the go to for walking developers through integrating a payment flow. As an example, https://stripe.com/docs/payments/accept-a-payment?integration=checkout covers an end-to-end integration which does a checkout with inline price data and also provides failure card info.

I'll make sure to investigate the gaps you mentioned and see where we can add links to improve visibility of our existing documentation.

PR #764, opened 2022-01-19
Use tox-gh-actions

ob-stripe:
Currently GitHub Actions is configured to run `make test`, which simply runs `tox`, which will test all Python versions for which an interpreter is available. GitHub Actions has its own version matrix, so we end up testing N*N versions.

(Example output: https://github.com/stripe/stripe-python/runs/4872868620?check_suite_focus=true#step:8:397)

This PR uses [tox-gh-actions](https://github.com/ymyzk/tox-gh-actions) to ensure that for each version in the GitHub Actions version matrix, we only run tests for that version.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -54,12 +54,15 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
+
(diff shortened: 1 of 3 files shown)


richardm-stripe (approved):
Oof, that is a bit silly. Thank you @ob-stripe!

PR #767, opened 2022-01-24
Bump vendored six to 1.16.0

stianjensen:
1.16.0 includes support for Python 3.10


stianjensen:
Any interest in either bumping or getting rid of six?
The Stripe SDK is responsible for quite a lot of `ImportWarning: _SixMetaPathImporter.find_spec() not found; falling back to find_module()` warnings.

dcr-stripe (approved):
Thank you for the contribution and sorry for the delay here @stianjensen .

I've confirmed this vendored from https://github.com/benjaminp/six/blob/45f1a230f9cc8e48372e19627b91ac06a2013292/six.py.

Merging and this will go out with the next release.

PR #768, opened 2022-01-24
Update CA cert bundle

illia-v:
There has been a bunch of changes to the list of trusted CA certs since 2018. I updated the bundle using `mk-ca-bundle.pl`.


dcr-stripe:
Thanks for the PR @illia-v , and sorry for the incredibly late response. 

We have to do this across a few of our repositories, so I want to make sure that work is coordinated. We have a major version coming up shortly - I'll open PRs across all our repos to do this (and use the new April 26 version).

dcr-stripe:
I've created:

* https://github.com/stripe/stripe-ruby/pull/1079
* https://github.com/stripe/stripe-php/pull/1314
* https://github.com/stripe/stripe-python/pull/829

Going to close this PR out now since it's been superceeded by the above. Thank you again!

PR #769, opened 2022-01-25
Drop unsupported Python versions from testing 

stianjensen:


stianjensen:
Just realized there's already a PR being worked on for this. I started looking at this to be able to remove the vendored six code, which I also see is being done in https://github.com/stripe/stripe-python/pull/708/. So this may be closed, then! :+1:

PR #773, opened 2022-02-24
Add infrastructure for test-helper methods

pakrym-stripe:
When generating test_helper methods on a resource we group them into a TestHelpers nested class:

```python
        @test_helpers
        @custom_method("pdf", http_verb="get")
        class Foo(CreateableAPIResource):
          OBJECT_NAME = "foo"
          def pdf(self, idempotency_key=None, **params):
               #...

          @custom_method("boop", http_verb="post")
          class TestHelpers(APIResourceTestHelpers):
            def boop(self, idempotency_key=None, **params):
              url = self.instance_url() + "/boop"
              headers = util.populate_headers(idempotency_key)
              self.resource.refresh_from(self.resource.request("post", url, params, headers))
              return self.resource
```

We need a few changes to infrastructure to get this working:
1. `APIResourceTestHelpers` that contains `class_url` and `instance_url` generation logic (so `test_helpers` is included in URLs). 
2. `test_helpers` decorator to add test_helpers property to the resource class.
3. `custom_method` logic adjusted to support having static method on a nested class.


pakrym-stripe:
re-r? @richardm

tox.ini
```diff
@@ -39,7 +39,7 @@ deps =
     pytest-cov >= 2.8.1, < 2.11.0
     pytest-mock >= 2.0.0
     pytest-xdist >= 1.31.0
-commands = pytest --cov {posargs:-n auto}
+commands = pytest --cov {posargs:-n auto} --ignore stripe
```
richardm-stripe, line 42: Add a comment to explain `--ignore stripe`?

richardm-stripe (approved):
This looks good - let's wait to merge until the first `test_helper` method is released.

stripe/api_resources/abstract/test_helpers.py
```diff
@@ -0,0 +1,49 @@
...
+    def class_url(cls):
+        if cls == APIResourceTestHelpers:
+            raise NotImplementedError(
+                "APIResourceTestHelpers is an abstract class.  You should perform "
+                "actions on its subclasses (e.g. Charge, Customer)"
+            )
+        # Namespaces are separated in object names with periods (.) and in URLs
+        # with forward slashes (/), so replace the former with the latter.
+        base = cls._resource_cls.OBJECT_NAME.replace(".", "/")
+        return "/v1/test_helpers/%ss" % (base,)
```
dcr-stripe, line 21: Should there be a second `s` here (`%ss`)?

stripe/api_resources/abstract/test_helpers.py
```diff
@@ -0,0 +1,49 @@
...
+                "id",
+            )
+
+        id = util.utf8(id)
+        base = self.class_url()
+        extn = quote_plus(id)
+        return "%s/%s" % (base, extn)
+
+
+def test_helpers(cls):
```
dcr-stripe, line 40: Nit: Let's add a comment explaining what this does/is for?

tests/api_resources/abstract/test_test_helpers_api_resource.py
```diff
@@ -0,0 +1,61 @@
...
+        obj = self.MyTestHelpersResource.TestHelpers.do_stuff("mid", foo="bar")
+
+        request_mock.assert_requested(
+            "post",
+            "/v1/test_helpers/myresources/mid/do_the_thing",
+            {"foo": "bar"},
+        )
+        assert obj.thing_done is True
+
+    def test_call_custom_method_instance(self, request_mock):
```
dcr-stripe, line 45: Nit: Maybe we can rename this to `test_call_custom_method_instance_via_property` to make it clear what this is testing (that it's testing out the annotation)?

stripe/api_resources/abstract/test_helpers.py
```diff
@@ -0,0 +1,49 @@
+from __future__ import absolute_import, division, print_function
+
+from stripe import error, util, six
+from stripe.six.moves.urllib.parse import quote_plus
+
+
+class APIResourceTestHelpers:
```
dcr-stripe, line 7: Nit: Can we document that this is for?

stripe/api_resources/abstract/test_helpers.py
```diff
@@ -0,0 +1,49 @@
...
+        return "%s/%s" % (base, extn)
+
+
+def test_helpers(cls):
+    def test_helpers_getter(self):
+        return self.TestHelpers(self)
+
+    cls.TestHelpers._resource_cls = cls
+    cls.TestHelpers._static_request = cls._static_request
+    cls.TestHelpers._static_request_stream = cls._static_request_stream
```
dcr-stripe, line 46: I wonder if we should share some base class/mixin between this and `APIResource` so that we don't have to do this. In the interim, let's throw if cls isn't an instance of `APIResource`?

dcr-stripe (commented):
Thanks Pavel! Mainly all minor comments about adding documentation, otherwise this LGTM!

PTAL @pakrym-stripe

stripe/api_resources/abstract/test_helpers.py
pakrym-stripe: Yep, the `OBJECT_NAME` is singular: 

```
    OBJECT_NAME = "terminal.reader"
```

dcr-stripe: Oh I see this is just the same implementation as APIResource. Thanks for confirming!

pakrym-stripe:
PTAL @dcr-stripe ?

dcr-stripe (approved):
LGTM, this looks great! Thanks Pavel!

PR #777, opened 2022-03-16
Add support for SearchResult.

dcr-stripe:
### Notify

r? @pakrym-stripe 

### Summary

Adds support for `search_result` objects returned by the Stripe API, including adding a `SearchResultObject` class and a `SearchableAPIResource` class which can be added to resources to support invoking search APIs.

This behaves similarly to `ListObject` in that they both wrap around a list of objects and provide pagination. However the `SearchResultObject` object paginates by relying on a `next_page` token included in the response rather than using object IDs + `starting_before`/`ending_after`. Thus, only forward pagination is supported

This was branched from https://github.com/stripe/stripe-python/pull/730.

### Test plan

Added unit tests. We'll want to add some more rigorous tests once stripe-mock has been updated to include at least one API that returns this resource.

### DIFF
--- a/stripe/api_resources/__init__.py
+++ b/stripe/api_resources/__init__.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, division, print_function
 
 from stripe.api_resources.error_object import ErrorObject, OAuthErrorObject
 from stripe.api_resources.list_object import ListObject
+from stripe.api_resources.search_result_object import SearchResultObject
 
 from stripe.api_resources import billing_portal
 from stripe.api_resources import checkout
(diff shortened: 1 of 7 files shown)


pakrym-stripe:
Can you please add a test similar to https://github.com/stripe/stripe-python/blob/master/tests/api_resources/abstract/test_listable_api_resource.py that exercises the SearchableAPIResource?

dcr-stripe:
> Can you please add a test similar to https://github.com/stripe/stripe-python/blob/master/tests/api_resources/abstract/test_listable_api_resource.py that exercises the SearchableAPIResource?

Good catch - thank you! Added.

PTAL @pakrym-stripe

pakrym-stripe (approved):

pakrym-stripe (approved):

PR #784, opened 2022-03-28
Pin click dependency to 8.0.4 to avoid breakage in black

pakrym-stripe:
The new version of https://pypi.org/project/click/#history package broke the code formatter we use. Pin to a known good version.


dcr-stripe (approved):
LGTM.

PR title will show up in the changelog, so let's change the title to be more descriptive. Maybe "Pin `click` dependency to 8.0.4 to avoid breakage in `black`"?

PR #791, opened 2022-04-13
API Updates

richardm-stripe:
Codegen for openapi e0123a3.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `increment_authorization` method on resource `PaymentIntent`


dcr-stripe (approved):
LGTM!

PR #794, opened 2022-04-18
API Updates

richardm-stripe:
## Summary
Contains corrections to #792 which was merged but not released.

We are not releasing `GET /v1/customers/:id/funding_instructions` yet and are releasing only the `POST`, and are generating it as a "custom method" instead of the nested resource method.


stripe/api_resources/customer.py
```diff
@@ -43,6 +44,12 @@ class Customer(
 ):
     OBJECT_NAME = "customer"
 
+    def create_funding_instructions(self, idempotency_key=None, **params):
+        url = self.instance_url() + "/funding_instructions"
+        headers = util.populate_headers(idempotency_key)
+        self.refresh_from(self.request("post", url, params, headers))
```
dcr-stripe, line 50: This shouldn't refresh_from and instead return the funding instructions

dcr-stripe (changes requested):

stripe/api_resources/customer.py
richardm-stripe: Fixed!

stripe/object_classes.py
```diff
@@ -39,7 +39,6 @@
     api_resources.File.OBJECT_NAME: api_resources.File,
     api_resources.File.OBJECT_NAME_ALT: api_resources.File,
     api_resources.FileLink.OBJECT_NAME: api_resources.FileLink,
-    api_resources.FundingInstructions.OBJECT_NAME: api_resources.FundingInstructions,
```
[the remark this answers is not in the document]
richardm-stripe: Fixed!

kamil-stripe (approved):

stripe/api_resources/terminal/reader.py
```diff
@@ -53,7 +53,6 @@ class TestHelpers(APIResourceTestHelpers):
         def present_payment_method(self, idempotency_key=None, **params):
             url = self.instance_url() + "/present_payment_method"
             headers = util.populate_headers(idempotency_key)
-            self.resource.refresh_from(
-                self.resource.request("post", url, params, headers)
-            )
-            return self.resource
+            resp = self.resource.request("post", url, params, headers)
```
pakrym-stripe, line 56: This might be a breaking change, we are not refreshing the original resource anymore..

PR #797, opened 2022-04-22
Add python release action

pakrym-stripe:
Adds a github action to publish python package to PyPy.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -102,7 +102,7 @@ jobs:
 
   publish:
     name: Publish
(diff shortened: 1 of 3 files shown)


.github/workflows/python-package.yml
```diff
@@ -1,4 +1,4 @@
-name: Python package
+name: CI
```
pakrym-stripe, line 1: Consistency

.github/workflows/python-package.yml
```diff
@@ -1,36 +1,56 @@
...
 
 on:
+  workflow_dispatch: {}
   push:
     branches: [master]
```
dcr-stripe, line 6: Let's add the tag regex here.

dcr-stripe (approved):
LGTM - just need to add the tag filter and then good to go!

PR #801, opened 2022-04-28
Try to fix the gpg: signing failed: Inappropriate ioctl for device error

pakrym-stripe:
Solution from https://github.com/keybase/keybase-issues/issues/2798


dcr-stripe (approved):
LGTM!

PR #803, opened 2022-04-28
Manually sign python packages

pakrym-stripe:
Use manual signing for the python package. 

Also allow running the workflow on a custom branch so it's easier to iterate. I'll remove it when things work.


dcr-stripe (approved):
LGTM

PR #816, opened 2022-05-26
API Updates

pakrym-stripe:
Codegen for openapi v149.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `persons` method on resource `Account`
* Add support for `balance_transactions` method on resource `Customer`


kamil-stripe (changes requested):
Something is wrong with formatting. Skipping for now.

PR #818, opened 2022-05-31
fix: Update cash balance methods to no longer require nested ID.

dcr-stripe:
## Summary

Updates `Customer.retrieve_cash_balance` and `Customer.modify_cash_balance` methods to no longer require a nested ID.

Previously you always had to call with `None` as the second parameter (eg. `retrieve_cash_balance('cus_123', None)`). 
With this PR, the second parameter is `None` by default. Now you can call this just with the customer ID (eg. `retrieve_cash_balance('cus_123')`).

### DIFF
--- a/tests/api_resources/test_customer.py
+++ b/tests/api_resources/test_customer.py
@@ -188,3 +188,42 @@ class TestCustomerPaymentMethods(object):
             "get", "/v1/customers/%s/payment_methods" % TEST_RESOURCE_ID
         )
         assert isinstance(resource, stripe.ListObject)
+
+
+class TestCustomerCashBalanceMethods(object):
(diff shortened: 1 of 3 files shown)


richardm-stripe (approved):

tests/api_resources/test_customer.py
```diff
@@ -188,3 +188,42 @@ def test_is_listable_on_object(self, request_mock):
             "get", "/v1/customers/%s/payment_methods" % TEST_RESOURCE_ID
         )
         assert isinstance(resource, stripe.ListObject)
+
+
+class TestCustomerCashBalanceMethods(object):
+    # These tests are present for compatibility purposes. Previously the cash
```
richardm-stripe, line 194: thanks for these tests and the comment

PR #823, opened 2022-06-15
Trigger workflows on beta branches

pakrym-stripe:
Add `beta` branch to `push` and `pr` triggers. Also add support for beta version updates. Rename yml file for consistency with other repos.

r? @kamil-stripe


VERSION
```diff
@@ -1 +1 @@
-3.3.0
+39.119.0b2
```
kamil-stripe, line 1: Is that correct?

kamil-stripe: Seems like a big jump

pakrym-stripe: Nope, accidental.

kamil-stripe (approved):

PR #826, opened 2022-06-21
Use the generated API version

pakrym-stripe:
To support generating beta header values, use the generated ApiVersion value instead of the hardcoded one.

This change applies to beta only.

r? @dcr-stripe


dcr-stripe (approved):
LGTM

Apdlrcjafg19:
Zvhgzhtjgthdgk

PR #827, opened 2022-06-23
API Updates

dcr-stripe:
Codegen for openapi v157.
r? @pakrym-stripe
cc @stripe/api-libraries

## Changelog
Specs are identical


dcr-stripe:
Closing since this is effectively a no-op and can just go out with next release.

pakrym-stripe:
You can still merge it.

PR #832, opened 2022-06-30
Unify request sending and result parsing logic

pakrym-stripe:
Unify request sending logic into a single method and call it from all the places we make requests from.


stripe/stripe_object.py
```diff
@@ -233,18 +233,41 @@ def api_base(cls):
...
-        if params is None:
-            params = self._retrieve_params
+        return StripeObject._request(
+            self, method, url, headers=headers, params=params
+        )
+
+    def _request(
+        self,
+        method_,
+        url_,
```
dcr-stripe, line 243: Nit: Why the _ suffix here?

dcr-stripe: Ah just saw https://github.com/stripe/stripe-python/pull/832/files#diff-4656b48a9a21bc574de7c8bfb7e0aaa00241eb172da79110b30d6615685b64f3R101-R102 - can we carry over that comment here + in api_resource so we don't accidentally remove it?

dcr-stripe (approved):
LGTM overall! Just one comment!

stripe/stripe_object.py
pakrym-stripe: Done

pakrym-stripe:
Removed some outdated tests for APIs that were removed from the next major.

r? @dcr-stripe

stripe/api_resources/transfer.py
```diff
@@ -18,34 +17,3 @@ class Transfer(
     UpdateableAPIResource,
 ):
     OBJECT_NAME = "transfer"
-
-    @classmethod
-    def _cls_cancel(
```
dcr-stripe: Why was this removed in that commit? Should this be in the changelog?

stripe/api_resources/issuer_fraud_record.py
```diff
@@ -1,8 +0,0 @@
-# File generated from our OpenAPI spec
-from __future__ import absolute_import, division, print_function
-
-from stripe.api_resources.abstract import ListableAPIResource
-
-
-class IssuerFraudRecord(ListableAPIResource):
```
dcr-stripe, line 7: Same with this deletion

stripe/api_resources/transfer.py
pakrym-stripe: I can open a separate PR with those changes. This is the result of me running the latest generator against the latest  openapi from `sdk-release/next-major`

dcr-stripe (approved):
LGTM!

PR #835, opened 2022-07-06
API Updates

pakrym-stripe:
Codegen for openapi v162.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog

- Use auto-generation for credit_note and invoice methods.


yejia-stripe (approved):
Is there any functional change here and could you add to changelog if so?

PR #836, opened 2022-07-07
API Updates

pakrym-stripe:
Codegen for openapi v163.
r? @richardm-stripe
cc @stripe/api-libraries


kamil-stripe (approved):
Not sure if it is worth releasing it ;)

PR #839, opened 2022-07-12
Update the generated code

pakrym-stripe:
## Changelog
- Removed the `IssuerFraudRecord` resource. Please prefer the `EarlyFraudWarning` resource instead.
- Removed the deprecated `Transaction.cancel` method.

r? @kamil-stripe


dcr-stripe (approved):
LGTM

PR #847, opened 2022-08-01
API Updates

pakrym-stripe:
Codegen for openapi v174.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Update the generated code.


dcr-stripe (approved):
LGTM, though we probably want to tweak the changelog.

Issue #849, opened 2022-08-02
v4.0.0 `idempotency_key` parameter unknown


newenegue:
### Describe the bug

Any of our requests to `POST /v1/subscriptions`, `POST /v1/invoices` and `POST /v1/customers/:id` are failing with `Received unknown parameter: idempotency_key`

### To Reproduce

Make a request to create a customer and include the `idempotency_key` parameter
```
stripe.Customer.create(
  name="John Doe",
  email=email,
  idempotency_key="test-key-12345",
)
```

### Expected behavior

`idempotency_key` is an accepted parameter

### OS

macOS

### Language version

Python 3.7.3

### Library version

stripe-python v4.0.0

### API version

2019-05-16

remi-stripe:
@newenegue thanks a lot for the report. There is definitely something wrong where some API methods don't handle options like `idempotency_key` and `api_key` properly but others do. We're investigating but for now you should roll back to the previous major.

newenegue:
@remi-stripe thanks for the response!  We've rolled back to v3.5.0 and everything is working as expected.

ramya-stripe:
We have just released version 4.0.1 of the package which has the fix for this issue. Rolling back to previous major is no longer necessary.

PR #850, opened 2022-08-03
Fix incorrect handling of additional request parameters

kamil-stripe:
Fixes https://github.com/stripe/stripe-python/issues/849

Fixes issue where using special parameter like `api_key`, `idempotency_key`, `stripe_version`, `stripe_account`, `headers` can cause a `Received unknown parameter error`.

Before fix:
``` 
resource = stripe.Customer.modify(
    "cus_123", metadata={"key": "value"}, email="a@b.c", idempotency_key="abc"
)
```
resulted in 
```
stripe.error.InvalidRequestError: Request req_123: Received unknown parameter: idempotency_key
```
The code executes fine with the fix.

### DIFF
--- a/stripe/api_resources/abstract/api_resource.py
+++ b/stripe/api_resources/abstract/api_resource.py
@@ -116,19 +116,19 @@ class APIResource(StripeObject):
         params=None,
     ):
         if api_key is None and 'api_key' in params:
-            api_key = params['api_key'];
+            api_key = params['api_key']
             del params['api_key']
 
         if idempotency_key is None and 'idempotency_key' in params:
-            idempotency_key = params['idempotency_key'];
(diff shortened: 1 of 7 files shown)


stripe/api_resources/abstract/api_resource.py
```diff
@@ -137,6 +153,22 @@ def _static_request_stream(
         stripe_account=None,
         params=None,
     ):
+        if api_key is None and params and "api_key" in params:
+            api_key = params["api_key"]
+            del params["api_key"]
```
pakrym-stripe, line 158: we should always remove these from params even if it was overriden

dcr-stripe (approved):

dcr-stripe (approved):

dcr-stripe (approved):

PR #856, opened 2022-08-05
API Updates

pakrym-stripe:
Codegen for openapi v176.
r? @yejia-stripe
cc @stripe/api-libraries


yejia-stripe (approved):
Approved but there are no user-facing changes

PR #861, opened 2022-08-18
API Updates

kamil-stripe:
Codegen for openapi v183.
r? @pakrym-stripe
cc @stripe/api-libraries

## Changelog
* Add support for new resource `CustomerCashBalanceTransaction`

PR #866, opened 2022-08-23
API Updates for beta branch

pakrym-stripe:
Codegen for openapi v184.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
- Updated stable APIs to the latest version


stripe/api_resources/plan.py
```diff
@@ -13,4 +13,15 @@ class Plan(
     ListableAPIResource,
     UpdateableAPIResource,
 ):
+    """
+    You can now model subscriptions more flexibly using the [Prices API](https://stripe.com/docs/api#prices). It replaces the Plans API and is backwards compatible to simplify your migration.
+
+    Plans define the base price, currency, and billing cycle for recurring purchases of products.
+    [Products](https://stripe.com/docs/api#products) help you track inventory or provisioning, and plans help you track pricing. Different physical goods or levels of service should be represented by products, and pricing options should be represented by plans. This approach lets you change prices without having to change your provisioning scheme.
+
+    For example, you might have a single "gold" product that has plans for $10/month, $100/year, €9/month, and €90/year.
```
dcr-stripe, line 22: The euro symbols in this and price.py need to be manually changed to EUR: https://github.com/stripe/stripe-python/pull/861/commits/da09005a12e9b1483328a3de47d4aea090a8a4ec

Kamil is fixing this in the codegen layer but it's not in yet.

dcr-stripe (changes requested):
Just one thing to change and then LG!

stripe/api_resources/plan.py
pakrym-stripe: I'll regen when his fix is in.

pakrym-stripe, line 22:
```suggestion
    For example, you might have a single "gold" product that has plans for $10/month, $100/year, EUR 9/month, and EUR 90/year.
```

pakrym-stripe:
r? @dcr

Updated.

PR #872, opened 2022-09-06
Corrected mistyped OAuth exceptions

cosseyluke:
While looking over the APIRequestor, I noticed these two mistyped OAuth exception classes.

Hope this helps.

## Changelog
* Fix mistyped OAuth exceptions.


richardm-stripe:
Hello @cosseyluke. I appreciate the contribution.

This bug has been around long enough that we're going to consider the fix for this to be a breaking change, as there is some risk that fixing it would break users working around the typo and catching the `NameError` in order to detect these type of errors, and we are anticipating another major soon - we will merge this change then.

anniel-stripe (approved):

PR #873, opened 2022-09-06
Add abstract methods for SearchableAPIResource

yejia-stripe:
r? @anniel-stripe 

## Summary
Add `search` and `search_auto_paging_iter` "abstract methods" for `SearchableAPIResource`. This is functionally a no-op, but fixes the tests for the type annotation stubs here: https://github.com/python/typeshed/runs/8215124463


anniel-stripe (approved):
LGTM!

PR #557, opened 2019-04-18
Add deprecation warning to instance methods

ob-stripe:
r? @brandur-stripe @remi-stripe 
cc @stripe/api-libraries 

Adds a deprecation warning to all instance methods decorated with `@class_method_variant`.

The warnings will look like this:
```
  /Users/ob/workspace/python/stripe-python/tests/api_resources/test_customer.py:49: DeprecationWarning: This instance method is deprecated. Use the class method stripe.Customer.delete('cus_123', ...) instead.

  /Users/ob/workspace/python/stripe-python/tests/api_resources/terminal/test_reader.py:51: DeprecationWarning: This instance method is deprecated. Use the class method stripe.terminal.Reader.delete('rdr_123', ...) instead.
```

PR #887, opened 2022-10-19
Deprecate save()

anniel-stripe:
## Summary
Mark `save()` method as deprecated. Also update `NotImplementedError` messages in some resources to point to `modify()` instead of `save()`.

## Changelog
* Deprecate `save` method on resources in favor of `modify`.

### DIFF
--- a/stripe/api_resources/abstract/updateable_api_resource.py
+++ b/stripe/api_resources/abstract/updateable_api_resource.py
@@ -4,7 +4,6 @@ from stripe import util
 from stripe.api_resources.abstract.api_resource import APIResource
 from stripe.six.moves.urllib.parse import quote_plus
 
-import warnings
 
 class UpdateableAPIResource(APIResource):
     @classmethod
@@ -13,13 +12,11 @@ class UpdateableAPIResource(APIResource):
         return cls._static_request("post", url, params=params)
 
     def save(self, idempotency_key=None):
-        warnings.warn(
-            """
-            The `save` method is deprecated and will be removed in
-            a future major version of the library.
-            Use the class method `modify` on the resource instead.
-            """,
-            DeprecationWarning)
+        """
+        The `save` method is deprecated and will be removed in a future major version of the library.
+
+        Use the class method `modify` on the resource instead.
+        """
         updated_params = self.serialize(None)
         if updated_params:
             self._request_and_refresh(
--- a/stripe/api_resources/abstract/updateable_api_resource.py
+++ b/stripe/api_resources/abstract/updateable_api_resource.py
@@ -12,7 +12,11 @@ class UpdateableAPIResource(APIResource):
         return cls._static_request("post", url, params=params)
 
     def save(self, idempotency_key=None):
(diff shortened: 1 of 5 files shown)


stripe/api_resources/abstract/updateable_api_resource.py
```diff
@@ -12,6 +12,11 @@ def modify(cls, sid, **params):
         return cls._static_request("post", url, params=params)
 
     def save(self, idempotency_key=None):
+        """
+        The `save` method is deprecated and will be removed in a future major version of the library.
+
+        Use the class method `modify` on the resource instead.
```
ramya-stripe, line 18: Are there no decorators or something else in Python that will help mark the method as deprecated to tooling/editors which can then provide the appropriate warning?

anniel-stripe: There are deprecated decorators in Python, but they would require installing a new package, and I saw that the current practice seems to just be putting it in the docstring ([`api_version` example](https://github.com/stripe/stripe-python/blob/291d569d/stripe/api_resources/file.py#L35)). I also saw that in the [last major version release](https://github.com/stripe/stripe-python/pull/844/files) the removed deprecated methods had been marked the same way).

I agree that a decorator that throws a warning would be a better user experience though. We could consider using https://deprecation.readthedocs.io/en/latest/

yejia-stripe: You can use [`warnings.warn(message, DeprecationWarning)`](https://docs.python.org/2/library/warnings.html#warnings.warn)

yejia-stripe (approved):
Lgtm! Replied to the comment thread about deprecation.

I think you may also need to add a `##Changelog` section in the PR description if you want it to show up in the generated changelog in the next release, but I'm not sure about that.

stripe/api_resources/abstract/updateable_api_resource.py
anniel-stripe: Ah yep that would be better, thank you Yejia!!

ramya-stripe: That kind of warning shows up in run time right rather than in build time correct? Do we have precedence for that? It might be considered noisy.

If just using the docstring has been the precedent so far, then I am ok with that. We anyway have a task to set up proper deprecation workflows across languages beyond this where we can evaluate the new package

yejia-stripe: Yes at run-time. Build-time in this case meaning linting while writing the code? I don't think we have precedence for it, but `DeprecationWarning` is ignored by default.

Agree with being ok about just using the docstring.

anniel-stripe: The last time we used warnings was in 2019: https://github.com/stripe/stripe-python/pull/557 (description contains an example of what the warnings will look like)

But yes, this could be considered noisy. I'll revert back to docstring to align with more recent precedent.

ramya-stripe (approved):

PR #889, opened 2022-11-01
API Updates

kamil-stripe:
Codegen for openapi v204.
cc @stripe/api-libraries


richardm-stripe (approved):
No substantive changes but might as well merge master

PR #895, opened 2022-11-04
Next major release changes

anniel-stripe:
# Changelog

## 5.0.0 - 2022-11-15

Breaking changes that arose during code generation of the library that we postponed for the next major version. For changes to the Stripe products, read more at https://stripe.com/docs/upgrades#2022-11-15.

"⚠️" symbol highlights breaking changes.

### ⚠️ Changed
- Dropped support for Python version 3.4 and 3.5 (#881). We now support Python 2.7 or 3.6+.
- Fixed mixtyped names for two OAuth exceptions: `UnsupportedGrantTypError`->`UnsupportedGrantTypeError` and `UnsupportedResponseTypError`->`UnsupportedResponseTypeError` (#872).

### Deprecated
- Deprecate `save` method on resources (#887). Use `modify` instead.
   ```python
  # Before
  customer = stripe.Customer.retrieve("cus_123")
  customer.email = "example@test.com"
  customer.save()

  # After
  stripe.Customer.modify("cus_123", email="example@test.com")
   ```

### ⚠️ Removed
- Removed `Orders` resource (#882).
- Removed `SKU` resource (#883).


ramya-stripe:
Feedback for the changelog entries

> Removed deprecated save method on resources. Use modify instead

But we didnt really "remove" it. We only deprecated it...

> Dropped support for Python version 3.5.

To avoid any misunderstanding that everything below 3.5 is also dropped, lets change this to `Dropped support for Python version 3.4 and 3.5. We now support Python 2.7 or 3.6+.`

We need an entry for the error name typo fix - UnsupportedResponseTypError -> UnsupportedResponseTypeError and the other one with it

xavdid-stripe (approved):
Changelog looks good!

Only thought was around the renamed error- might be worth adding a blurb like

> if you were catching the incorrectly-named `UnsupportedResponseTypError` in your code, make sure to update your `except` statements with `UnsupportedResponseTypeError` instead.

Though IIRC, the typo error didn't exist, so it wouldn't have ever been thrown in the first place and there won't be user code depending on it. That could also be worth mentioning.

PR #909, opened 2022-12-22
API Updates

richardm-stripe:
Codegen for openapi v219.
cc @stripe/api-libraries

Update to latest generated code.


pakrym-stripe (changes requested):
formatting

anniel-stripe (approved):
Changelog needs updating

PR #913, opened 2023-01-11
Pass params into logger.{info,debug}

### LINKED ISSUE #776: JSON log formatting
Hello, 

It would be great to allow users to format logs in json format instead of the hardcoded `logfmt` format in the library. For those looking to make this work, this is how we did it:

```python
# strip_log_formatter.py
import datetime
import json

from pythonjsonlogger import jsonlogger


class StripeJsonLogFormatter(jsonlogger.JsonFormatter):  # pragma: no cover
    def add_fields(self, log_record, record, message_dict):
        """
        This method allows us to inject custom data into resulting log messages
        """
        for field in self._required_fields:
            log_record[field] = record.__dict__.get(field)
        log_record.update(message_dict)

        # Add timestamp and application name if not present
        if 'timestamp' not in log_record:
            now = datetime.datetime.utcnow()
            log_record['timestamp'] = now.isoformat()
        if 'application' not in log_record:
            log_record['application'] = 'myapp'

        if 'level' not in log_record:
            log_record['level'] = record.levelname

        jsonlogger.merge_record_extra(record, log_record, reserved=self._skip_fields)
```

Reference this class from `stripe_log_formatter.py` in log ini configuration file as follows:

```ini
[loggers]
keys=stripe

[handlers]
keys=stripe_file

[formatters]
keys=stripe_json

[logger_stripe]
level=INFO
handlers=stripe_file
qualname=stripe
propagate=0

[handler_stripe_file]
class=logging.handlers.WatchedFileHandler
formatter=stripe_json
args=('/var/log/myapp/stripe.json',)

[formatter_stripe_json]
class=app.stripe_log_formatter.StripeJsonLogFormatter
```

And then at import time, monkey patch stripe library:

```python
# Monkey patch Stripe logfmt formatter
# Watch for this code if it ever changes, things will break:
# https://github.com/stripe/stripe-python/blob/v2.67.0/stripe/util.py#L82

old_log_fmt = stripe.util.logfmt

def new_logfmt(props):
    return OrderedDict(sorted(props.items(), key=lambda x: x[0]))

stripe.util.logfmt = new_logfmt
```

Any chance we can get an ability to configure our own log formatter?

richardm-stripe:
## Summary
Causes `util.log_info` and `util.log_debug` to pass more information into the underlying python `logger.info` and `logger.debug` methods.

By default, this will do nothing, but if you define and attach your own `logger.Formatter` you will be able to access this information on `.args` property on the argument to the `format` method.

```python
class CoolFormatter(logging.Formatter):
    def format(self, record):
        # Just a proof of concept. This won't handle "bytes" nicely.
        return json.dumps(record.args)
```

Fixes #776.


neilpanchal:
This LGTM.

pakrym-stripe (approved):

asherf:
this change probably causes https://github.com/stripe/stripe-python/issues/922

PR #934, opened 2023-03-02
API Updates

pakrym-stripe:
Codegen for openapi v237.
cc @stripe/api-libraries


richardm-stripe:
Let's merge but no release.

richardm-stripe (approved):

PR #962, opened 2023-04-19
ci: Fix github action warnings

antonblr:
1) Update GH Actions. Fixes the following warnings:
    GitHub Actions Warnings

2) Use `actions/setup-python` builtin [cache](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages) functionality (would need to re-run actions to test it)
3) Update `make ci-test` command to not use [`--parallel` mode](https://tox.wiki/en/latest/user_guide.html#parallel-mode). Every node executes tests for single python version / tox environment. This also improves GH Actions logs - you can expand the tests and coverage data:
      Expandable Logs

4) Update `coveralls` step - I think it should only "upload coverage to coveralls.io" (as per description). tox.ini `depends` config suggests the same - it should run after the tests run. I've tested it in `coveralls debug` mode - it pick-ups coverage files generated by the previous step (`Test with pytest`). Let me know if you have strong opinion about this step, I can revert the change.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -73,10 +73,7 @@ passenv = GITHUB_*
 deps =
     coverage >= 4.5.3, < 5 # TODO: upgrade to coverage 5 when we drop support for Python 3.4
     coveralls
-    pytest
-    pytest-mock
 commands =
-    coverage run --source=stripe -m pytest tests/
     coverage combine
     coveralls --service=github
 depends = py{310,39,38,37,36,35,34,27,py3,py2}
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -73,10 +73,7 @@ passenv = GITHUB_*
 deps =
     coverage >= 4.5.3, < 5 # TODO: upgrade to coverage 5 when we drop support for Python 3.4
     coveralls
-    pytest
```
pakrym-stripe, line 76: Don't we need these anymore?

pakrym-stripe:
@antonblr , thank you for your contribution! One quick question above before we merge this.

pakrym-stripe (approved):

pakrym-stripe:
Thank you!

PR #998, opened 2023-07-28
Deprecate @test_helper decorator

richardm-stripe:
The `@test_helper` decorator is somewhat awkward and will make it difficult to add type annotations.

This PR deprecates the decorator and

1. adds a getter

```python
  @property
  def test_helpers(self):
      return self.TestHelpers(self)
```

2. sets `._resource_cls` directly
```
MyResource.TestHelper._resource_cls = MyResource
```
and
3. Puts `def _static_request` and `def _static_request_stream` in the APIResourceTestHelpers parent class instead of being added dynamically by the decorator.

### DIFF
--- a/stripe/api_resources/abstract/test_helpers.py
+++ b/stripe/api_resources/abstract/test_helpers.py
@@ -20,6 +20,14 @@ class APIResourceTestHelpers:
         self.resource = resource
 
     @classmethod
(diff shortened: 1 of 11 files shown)


Silvrbckw (commented):
Reviewed and Commit

pakrym-stripe (approved):

PR #1009, opened 2023-08-14
 _ApiVersion -> ApiVersion

richardm-stripe:
* Changes _ApiVersion to ApiVersion, as I think it's legitimate for users to want to ask what the pinned version of the library is.
* ~Updates the version.~ (apparently release script does this automatically)


remi-stripe:
Can we fix the title as the API version number is wrong

stripe/api_version.py
```diff
@@ -2,5 +2,5 @@
 # File generated from our OpenAPI spec
 
 
-class _ApiVersion:
-    CURRENT = "2022-11-15"
+class ApiVersion:
```
pakrym-stripe, line 5: Changes _ApiVersion to ApiVersion, as I think it's legitimate for users to want to ask what the pinned version of the library is.

They have `Stripe.api_version`  for that.

pakrym-stripe (changes requested):
Do we need two ways to get api_version?

stripe/api_version.py
richardm-stripe: Stripe.api_version is the version *currently in use*. It *defaults* to ApiVersion.CURRENT, but is not a constant describing the library's pinned version.

richardm-stripe: Also FWIW this is public in every other library.

pakrym-stripe: We also don't have a similar feature in node.

Issue #802, opened 2022-04-28
Setting a List Field to [] When Modifying a Resource Doesn't Actually Empty The List Field


jake-kent:
### Describe the bug

When trying to clear out the list of previously set blocked_categories on a cardholder, calling
`
stripe.issuing.Cardholder.modify(
    "ich_...",
    spending_controls={'blocked_categories': []},
    stripe_account="acct_...",
)
`
returns successfully but doesn't clear out the blocked_categories on the cardholder and returns a cardholder instance with the previous blocked_categories list still set.

Instead, it is necessary to replace [] with "":
`
stripe.issuing.Cardholder.modify(
    "ich_...",
    spending_controls={'blocked_categories': ""},
    stripe_account="acct_...",
)
`
and that successfully clears the list of blocked_categories—however, there is no mention of the need to make this distinction when calling stripe.issuing.Cardholder.modify or any other method in the Stripe API docs or this library's docs and setting a list to empty by setting its value to "" is not valid Python syntax.

### To Reproduce

1. create a stripe connect account via:
`
stripe.Account.create(
    type="custom",
    country="US",
    email="testaccount@emailprovider.com",
    requested_capabilities=["transfers", "card_payments", "card_issuing"],
)
`
2. complete any requested test KYB information to ensure the account is fully verified and approved
3. create a cardholder via:
`
stripe.issuing.Cardholder.create(
    name="Jane Doe",
    email="testcardholder@emailprovider.com",
    status="active",
    type="individual",
    billing={
        "name": "Jane Doe",
        "address": {
            "line1": "117 West St",
            "line2": None,
            "city": "New York",
            "state": "NY",
            "country": "US",
            "postal_code": "10006",
        },
    },
    spending_controls={'blocked_categories': ['floor_covering_stores']},
    stripe_account="acct_...",  # id of the account created above
)
`
4. attempt to update the cardholder via:
`
stripe.issuing.Cardholder.modify(
    "ich_...",  # id of the cardholder created above
    spending_controls={'blocked_categories': []},
    stripe_account="acct_...",  # id of the account created above
)
`

### Expected behavior

Setting a list to [] should set that list to empty. It is not intuitive or valid Python syntax to set a list to empty by setting its value to "".

### Code snippets

```Python
See code snippets in the "Describe the bug" and "To Reproduce" sections
```

### OS

macOS

### Language version

Python 3.10.2

### Library version

stripe-python v2.67.0

### API version

2019-10-17

[...]

PR #1018, opened 2023-08-20
Allowing [] input instead of '' in UpdateableAPIResource modify method

JoJo10Smith:
I define a new function `_format_values` which looks through the entire params input of the `modify `method and will change all final instances of [] to "" (empty string) so that the user can input [] to empty the list instead of having to input an empty string. 

This PR resolves issue #802 

Please feel free to give feedback on the implementation/ any changes I could make to have this merged. I look forward to hearing feedback.
Thanks
Jordan

### DIFF
--- a/stripe/api_resources/abstract/updateable_api_resource.py
+++ b/stripe/api_resources/abstract/updateable_api_resource.py
@@ -9,7 +9,27 @@ class UpdateableAPIResource(APIResource):
     @classmethod
     def modify(cls, sid, **params):
         url = "%s/%s" % (cls.class_url(), quote_plus(sid))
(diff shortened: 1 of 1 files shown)


anniel-stripe:
Hi @JoJo10Smith, unfortunately we won't be accepting this PR at this time - this is a breaking change, as users who explicitly set a param to `[]` may have been relying, intentionally or not, on the library's no-op behavior. We're also unlikely to prioritize this change in the next major version, as we'd like to investigate this a bit more before applying a disruptive fix.

JoJo10Smith:
Got it, thank you for the explanation @anniel-stripe

PR #1020, opened 2023-08-21
Run Pyright & begin adding type annotations

richardm-stripe:
## Summary

This PR begins to add type annotations to the library. The goal is a minimal amount of changes to to get `pyright` running and typechecking the library in CI with zero errors.

We add a `py.typed` to the repo but NOT to the `package_data` property of setup.py. This should ensure that the type annotations in the library don't get picked up by user typecheckers yet (we're not ready yet! we'll get there) but allows us to test our own library for consistency in CI.

We do, however, add `typing_extensions` as a dependency to the library.  We don't take adding dependencies lightly, but
this dependency is inescapable as we still target python 3.6, and are planning on using fairly recent type system features (`Unpack` for typed **kwargs). This dependency is lightweight and seems very common.

For now, the library is still compatible with `typing_extensions > 3.7.2` but this will narrow as we make use of newer type features.

Please let us know here on Github if you experience any problems as a result of this new dependency.


stripe/http_client.py
```diff
@@ -13,23 +13,27 @@
 from stripe import error, util
 from stripe.request_metrics import RequestMetrics
 
+from typing import NoReturn, Optional, cast
+from typing_extensions import TypedDict
```
pakrym-stripe, line 17: We need to figure out a pattern for when we import from typing_extensions or typing

stripe/http_client.py
```diff
@@ -494,24 +516,35 @@ def close(self):
...
         # consumes the proxy url in small pieces
         if self._proxy:
             # now that we have the parser, get the proxy url pieces
-            proxy = self._proxy
-            for scheme, value in proxy.items():
-                proxy[scheme] = urlparse(value)
+            # Note: self._proxy is actually dict[str, str] because this is the
+            # type on the superclass. Here, we reassign self._proxy into
+            # dict[str, ParseResult]
+            proxy_ = cast("dict[str, str]", self._proxy)
```
pakrym-stripe, line 545: https://typing-extensions.readthedocs.io/en/latest/#Dict ?

tests/test_http_client.py
```diff
@@ -11,13 +11,18 @@
 
 
 class StripeClientTestCase(object):
-    REQUEST_LIBRARIES = ["urlfetch", "requests", "pycurl", "urllib.request"]
+    REQUEST_LIBRARIES = [
```
pakrym-stripe, line 14: Is this change related to types?

tests/test_api_requestor.py
```diff
@@ -777,11 +777,14 @@ def setup_stripe(self):
         stripe.api_key = orig_attrs["api_key"]
         stripe.default_http_client = orig_attrs["default_http_client"]
 
-    def test_default_http_client_called(self, mocker):
-        hc = mocker.Mock(stripe.http_client.HTTPClient)
-        hc._verify_ssl_certs = True
-        hc.name = "mockclient"
-        hc.request_with_retries = mocker.Mock(return_value=("{}", 200, {}))
+    def test_default_http_client_called(self, http_client_mock):
```
pakrym-stripe, line 780: Unrelated?

stripe/oauth.py
```diff
@@ -43,15 +43,19 @@ def authorize_url(express=False, **params):
...
+            )
         )
         response, _ = requestor.request("post", "/oauth/token", params, None)
         return response.data
 
     @staticmethod
     def deauthorize(api_key=None, **params):
         requestor = api_requestor.APIRequestor(
-            api_key, api_base=connect_api_base
+            options=_RequestOptions.merge(  # type: ignore
```
pakrym-stripe, line 56: Unrelated?

stripe/http_client.py
richardm-stripe: Let's do typing_extensions everywhere

tests/test_http_client.py
richardm-stripe: Yes. See the changes in `stripe/http_client.py`. `urllib.request` isn't publicly defined in the types when you `import urllib`, so instead we are doing `import urllib.request as urllibrequest`.

setup.py
```diff
@@ -31,6 +31,8 @@
     package_data={"stripe": ["data/ca-certificates.crt"]},
     zip_safe=False,
     install_requires=[
+        'typing_extensions <= 4.2.0; python_version < "3.7"',
+        'typing_extensions; python_version >= "3.7"',
```
pakrym-stripe, line 35:
```suggestion
        'typing_extensions >= 4.1.1; python_version >= "3.7"',
```

setup.py
```diff
@@ -31,6 +31,8 @@
     package_data={"stripe": ["data/ca-certificates.crt"]},
     zip_safe=False,
     install_requires=[
+        'typing_extensions <= 4.2.0; python_version < "3.7"',
```
pakrym-stripe, line 34:
```suggestion
        'typing_extensions <= 4.8.0; python_version < "3.7"',
```

setup.py
```diff
@@ -31,6 +31,8 @@
...
+        'typing_extensions >= 4.1.1; python_version >= "3.7"',
```
pakrym-stripe, line 35: Should these rules be based on 3.6?

stripe/api_resources/abstract/listable_api_resource.py
```diff
@@ -11,12 +12,22 @@ def auto_paging_iter(cls, *args, **params):
...
+        result = cls._static_request(
             "get",
             cls.class_url(),
             api_key=api_key,
             stripe_version=stripe_version,
             stripe_account=stripe_account,
             params=params,
         )
+
+        # TODO: note, this is technically a breaking change of stripe-python-next vs.
```
pakrym-stripe, line 25: Remove note?

stripe/api_resources/abstract/test_helpers.py
```diff
@@ -15,6 +22,9 @@ class Foo(APIResource):
       class TestHelpers(APIResourceTestHelpers):
     """
 
+    _resource_cls: ClassVar[Any] = None
```
pakrym-stripe, line 25: ClassVar[T]?

stripe/api_resources/refund.py
```diff
@@ -54,6 +55,8 @@ def cancel(self, idempotency_key=None, **params):
         )
 
     class TestHelpers(APIResourceTestHelpers):
+        _resource_cls: Type["Refund"]
```
pakrym-stripe, line 58: Do we need to type these?

setup.py
richardm-stripe: I don't understand why you are suggesting this? 4.2.0 dropped support for 3.6

stripe/http_client.py
```diff
@@ -494,24 +516,35 @@ def close(self):
...
 
 
+class _Proxy(TypedDict):
+    http: Optional[ParseResult]
+    https: Optional[ParseResult]
+
+
 class PycurlClient(HTTPClient):
     name = "pycurl"
+    _proxy: _Proxy
```
pakrym-stripe, line 526: Should we have a separate field for parsed proxy instead of mutating its type?

setup.py
richardm-stripe: I don't understand what you mean. We have one rule for python_version 3.7+ and a different rule for < 3.7... IMO the rules should be as unrestrictive as we can make them.

stripe/util.py
```diff
@@ -199,7 +202,19 @@ def convert_to_dict(obj):
         return obj
 
 
-def populate_headers(idempotency_key):
+@overload
+def populate_headers(
+    idempotency_key: str,
+) -> "dict[Literal['Idempotency-Key'], str]":
```
pakrym-stripe, line 208: `typing_extensions.Dict`?

pakrym-stripe (changes requested):
Nice work! Some comments.

setup.py
pakrym-stripe: I mean that our lowest support version is 3.6. We know that the highest version of  typing_extensions we can rely on is 4.1.1 because only it supports 3.6.

Should the rules be 
`typing_extensions == 4.1.1; python_version == "3.6"`
`typing_extensions >= 4.1.1; python_version > "3.6"` 

?

stripe/api_resources/abstract/test_helpers.py
richardm-stripe: That would be lovely, but
> "ClassVar" type cannot include type variables

setup.py
pakrym-stripe: Ah, that's what you have.

pakrym-stripe: What you have works.

stripe/api_resources/abstract/test_helpers.py
```diff
@@ -3,8 +3,15 @@
...
+from typing import TypeVar, ClassVar, Any
+from typing_extensions import Protocol
 
-class APIResourceTestHelpers:
+from stripe.api_resources.abstract.api_resource import APIResource
+
+T = TypeVar("T", bound=APIResource)
+
+
+class APIResourceTestHelpers(Protocol[T]):
```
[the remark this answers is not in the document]
richardm-stripe: No! Generic[T] works!

stripe/api_resources/abstract/searchable_api_resource.py
```diff
@@ -13,14 +16,16 @@ def _search(
...
-        return cls._static_request(
+        ret = cls._static_request(
             "get",
             search_url,
             api_key=api_key,
             stripe_version=stripe_version,
             stripe_account=stripe_account,
             params=params,
         )
+        # TODO (types)
```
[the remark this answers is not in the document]
richardm-stripe: Will do

stripe/api_resources/abstract/listable_api_resource.py
richardm-stripe: Will do

stripe/http_client.py
richardm-stripe: Yes!

stripe/http_client.py
```diff
@@ -494,24 +516,35 @@ def close(self):
         pass
 
 
+class _Proxy(TypedDict):
```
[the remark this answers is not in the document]
richardm-stripe: Yes!

stripe/api_resources/abstract/test_helpers.py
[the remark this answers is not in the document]
richardm-stripe: Oh, no I remembered what this was for.

We want to enforce that everybody who extends from `APIResourceTestHelpers` provides an appropriate `_resource_cls` field. So it is an "interface". If we extend from `Protocol[T]` then you get a type error if you omit `_resource_cls` from the type definition of e.g. `Terminal.Reader.TestHelpers`.

It's not perfect because the implementor of `APIResourceTestHelpers` can still forget to do e.g.
`Reader.TestHelpers._resource_cls = Reader` after the classes have been defined (it's not possible to do this inline because a class variable isn't in scope inside its own definition and TestHelpers is nested inside the containing resource), but I do think it helps clarify the fact that e.g. Terminal.Reader.TestHelpers has the responsibility to set `_resource_cls`.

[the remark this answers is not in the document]
richardm-stripe: I did end up having to [Fix test_helpers](https://github.com/stripe/stripe-python/pull/1020/commits/b531c68e4ef5af5735a4bc0955639ea8b048c7a9) to actually give the Protocol teeth, though.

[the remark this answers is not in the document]
richardm-stripe: We could make this an [abstract base class](https://docs.python.org/3/library/abc.html), I guess. That might feel better given that APIResourceTestHelpers provides implementations, too, but if you google "abstract base class vs protocol python" I get the idea that it's kind of six of one, half dozen of the other

stripe/api_resources/refund.py
richardm-stripe: At the end of the day, it doesn't *really* matter as this is code generated, but I do think it helps communicate what `APIResourceTestHelpers` expects from its implementors.

stripe/http_client.py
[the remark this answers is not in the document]
richardm-stripe: Fixed with [_proxy -> _parsed_proxy](https://github.com/stripe/stripe-python/pull/1020/commits/ee66bb0a8721004455ab4e3f27571af9aba3d1fb)

stripe/http_client.py
richardm-stripe: Fixed with [_proxy -> _parsed_proxy](https://github.com/stripe/stripe-python/pull/1020/commits/ee66bb0a8721004455ab4e3f27571af9aba3d1fb)

stripe/util.py
richardm-stripe: Fixed with [Use Dict not "dict[...]"](https://github.com/stripe/stripe-python/pull/1020/commits/ab5f1e9547fe7e4b7aeab51c6ff6d845a284f1af)

stripe/api_resources/abstract/searchable_api_resource.py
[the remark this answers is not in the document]
richardm-stripe: Fixed with [Code review: list/search results](https://github.com/stripe/stripe-python/pull/1020/commits/f1b8ef8fbba64b0fcfdcd344db850ca87bf08ee4)

stripe/api_resources/abstract/listable_api_resource.py
richardm-stripe: Fixed with [Code review: list/search results](https://github.com/stripe/stripe-python/pull/1020/commits/f1b8ef8fbba64b0fcfdcd344db850ca87bf08ee4)

stripe/api_resources/abstract/test_helpers.py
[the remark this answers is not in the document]
pakrym-stripe: >We could make this an [abstract base class](https://docs.python.org/3/library/abc.html), I guess. That might feel better given that APIResourceTestHelpers provides implementations, too, but if you google "abstract base class vs protocol python" I get the idea that it's kind of six of one, half dozen of the other

Understood. Thank you for the explanation. I'll leave it to you to pick the option you prefer.

pakrym-stripe (approved):
This is great! 🥳 🎆 🍾

PR #1023, opened 2023-08-24
Pyright: fix more type suppressions

richardm-stripe:
Fixes type suppressions:

* Made `stripe/__init__.py` explicitly define things that were being referenced on it and existed implicitly but weren't picked up by pyright.
* ListObject/SearchResult types have pagination properties declared.
* Other sundry fixes.


stripe/util.py
```diff
@@ -253,14 +256,18 @@ def __call__(self, method):
...
         @functools.wraps(self.method)
         def _wrapper(*args, **kwargs):
             if obj is not None:
                 # Method was called as an instance method, e.g.
                 # instance.method(...)
                 return self.method(obj, *args, **kwargs)
-            elif len(args) > 0 and isinstance(args[0], objtype):  # type: ignore
+            elif (
+                len(args) > 0
+                and objtype is not None
```
richardm-stripe, line 268:
```
>>> isinstance("foo", None)
TypeError: isinstance() arg 2 must be a type or tuple of types
```
(so nobody was successfully calling with objtype=None)

stripe/api_resources/list_object.py
```diff
@@ -138,23 +142,25 @@ def next_page(
...
         params_with_filters.update(params)
 
-        return self.list(
+        result = self.list(
             api_key=api_key,
             stripe_version=stripe_version,
             stripe_account=stripe_account,
             **params_with_filters
         )
+        assert isinstance(result, ListObject)
```
richardm-stripe, line 165: We already did this for next_page but the previous_page analog got lost :(. Would have been nice to go in the major since it can technically break somebody testing against unrealistic mocks, but can't break actual requests against the API, so I think it's peaceful to ship now.

.gitignore
```diff
@@ -41,7 +41,7 @@ coverage.xml
 .pytest_cache/
 
 # pyenv
-.python_version
+.python-version
```
anniel-stripe, line 44: Why are we changing this back after https://github.com/stripe/stripe-python/commit/469b3d6cac85ad639d308c06241f93b2b017228c#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947 ?

richardm-stripe: Oh, didn't even notice this. `python-version` is correct and 
https://github.com/stripe/stripe-python/commit/469b3d6cac85ad639d308c06241f93b2b017228c#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947 was mistaken.

anniel-stripe (approved):
LGTM!

stripe/api_resources/list_object.py
```diff
@@ -7,6 +8,9 @@
 
 class ListObject(StripeObject):
     OBJECT_NAME = "list"
+    data: list
```
pakrym-stripe, line 11: `List[StripeObject]` ?

stripe/api_resources/search_result_object.py
```diff
@@ -1,10 +1,14 @@
 from __future__ import absolute_import, division, print_function
+from typing_extensions import Self
 
 from stripe.stripe_object import StripeObject
 
 
 class SearchResultObject(StripeObject):
     OBJECT_NAME = "search_result"
+    data: list
```
pakrym-stripe, line 9: `List[StripeObject]`?

pakrym-stripe (approved):
some nits

stripe/api_resources/search_result_object.py
richardm-stripe: Let's give it a shot!

richardm-stripe: I did that. That produced a type error when doing `data[0].id` below. What I really want is

```python
T = TypeVar("T", bound=Intersection[StripeObject, _HasId])
class SearchResultObject(StripeObject, Generic[T]):
   data: List[T]
```

but so far python doesn't have intersection types: (see https://github.com/python/typing/issues/213). So I just enforce the presence of id dynamically, not statically.

pakrym-stripe (approved):

PR #1026, opened 2023-08-24
Types: annotate resource properties

richardm-stripe:
## Reviewer
This should be ready to go right after https://github.com/stripe/stripe-python/pull/1032 (this is based off that rn)

## Summary
* ~This won't actually do much good without method return types being declared.~ (solved by https://github.com/stripe/stripe-python/pull/1032)
* Adds class definitions for types that did occur on responses, but did not have any methods defined on them.
* Adds basic type annotations for simple fields on API responses.
  * No nested objects. These are `Any` for now.
  * Same for union types.


stripe/api_resources/application_fee_refund.py
```diff
@@ -17,6 +21,14 @@ class ApplicationFeeRefund(UpdateableAPIResource):
     """
 
     OBJECT_NAME = "fee_refund"
+    amount: int
+    balance_transaction: Optional[Any]
```
pakrym-stripe, line 25:
```suggestion
    balance_transaction: Optional[StripeObject]
```
??

richardm-stripe: Fixed!

stripe/api_resources/charge.py
```diff
@@ -23,6 +28,54 @@ class Charge(
...
+    refunded: bool
+    refunds: Optional[Any]
+    review: Optional[Any]
+    shipping: Optional[StripeObject]
+    source: Optional[Any]
+    source_transfer: Optional[Any]
+    statement_descriptor: Optional[str]
+    statement_descriptor_suffix: Optional[str]
+    status: str
+    transfer: Any
```
pakrym-stripe, line 76:
```suggestion
    transfer: Union[str, StripeObject]
```

?

richardm-stripe:
I think it's
> Adds class definitions for types that did occur on responses, but did not
have any methods defined on them.

stripe/api_resources/charge.py
richardm-stripe: Out of scope? I will do expandables and union types separately so it easier to focus on the logic specific to the more involved cases.

stripe/object_classes.py
```diff
@@ -102,6 +111,7 @@
     api_resources.treasury.CreditReversal.OBJECT_NAME: api_resources.treasury.CreditReversal,
     api_resources.treasury.DebitReversal.OBJECT_NAME: api_resources.treasury.DebitReversal,
     api_resources.treasury.FinancialAccount.OBJECT_NAME: api_resources.treasury.FinancialAccount,
+    api_resources.treasury.FinancialAccountFeatures.OBJECT_NAME: api_resources.treasury.FinancialAccountFeatures,
```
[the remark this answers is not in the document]
richardm-stripe: > Adds class definitions for types that did occur on responses, but did not have any methods defined on them.

stripe/api_resources/application.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# File generated from our OpenAPI spec
+from __future__ import absolute_import, division, print_function
+
+from stripe.stripe_object import StripeObject
+from typing import Optional
+from typing_extensions import Literal
+
+
+class Application(StripeObject):
```
pakrym-stripe, line 10: This feels like a separate decision from "annotating resource properties". Are we saying that we'll start generating all top level resources in type-annotated libraries?

richardm-stripe: What's the alternative? How do we give a type to a field like

https://github.com/stripe/stripe-python/pull/1026/files/08adc64efa3d4721987cb919724375c070b3cd59#diff-e8276cf621b514c555d80becdf96a85735068568b3b7bdcce05964dbd66234e9R66

?

Here's the analogous definition in `stripe-node`. https://github.com/stripe/stripe-node/blob/master/types/Applications.d.ts

If you like, I can postpone the decision for how to handle resource references to another PR and do `any` in these places for now.

pakrym-stripe: >What's the alternative? How do we give a type to a field like

`StripeObject`? similar to how we type other fields that don't have generated types in Python. And what we already do in PHP today.

I don't want to postpone it. I want to explicitly discuss and make it and file followups (like doing the same thing in PHP).

pakrym-stripe (approved):

PR #1030, opened 2023-08-30
Explicitly define CRUDL methods in each resource

richardm-stripe:
Explicitly defines create, retrieve, update, delete, and list methods in the body of each resource, as an override of the "standard" implementation from `UpdateableApiResource` etc.

This makes the class definition serve as better documentation and removes indirection. Eventually it would be good to do this for the "nested_resource" methods, too.

Branched off #1046.

Compare the CRUDL methods in coupon.py (left) with the parent class methods (right). They should be identical except for concrete classes instead of the generic `T`.
image

image
image

image

image

### DIFF
--- a/stripe/api_resources/account.py
+++ b/stripe/api_resources/account.py
@@ -12,7 +12,7 @@ from stripe.api_resources.abstract import (
 )
 from stripe.api_resources.list_object import ListObject
 from stripe.stripe_object import StripeObject
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, cast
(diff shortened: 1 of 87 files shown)


stripe/api_resources/__init__.py
```diff
@@ -3,7 +3,6 @@
 from __future__ import absolute_import, division, print_function
 
 # flake8: noqa
-from . import abstract
```
richardm-stripe, line 6: What's this about?

pakrym-stripe (approved):

PR #1031, opened 2023-08-30
Enable some pyright rules

richardm-stripe:
## Summary
Adds
```
reportMissingTypeArgument=true
reportUnnecessaryCast=true
reportUnnecessaryComparison=true
reportUnnecessaryContains=true
reportUnnecessaryIsInstance=true
```
pyright configuration, and type modifications to make the type checker pass.

### DIFF
--- a/stripe/stripe_object.py
+++ b/stripe/stripe_object.py
@@ -4,6 +4,7 @@ import datetime
 import json
 from copy import deepcopy
 from typing_extensions import TYPE_CHECKING
+from typing import Any, Dict
 
 import stripe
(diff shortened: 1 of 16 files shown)


stripe/stripe_object.py
```diff
@@ -34,7 +35,7 @@ def _serialize_list(array, previous):
     return params
 
 
-class StripeObject(dict):
+class StripeObject(Dict[str, Any]):
```
richardm-stripe, line 38: Finding a better type than Any is out of scope.

pakrym-stripe (approved):

PR #1032, opened 2023-08-30
Types for CRUDL methods on parents

richardm-stripe:
Add a type parameter to APIResource and related classes, and set that as the return type of

* APIResource.retrieve
* CreateableAPIResource.create
* DeleteableAPIResource.delete
* ListableAPIResource.list
* SingletonAPIResource.retrieve
* UpdateableAPIResource.modify

### DIFF
--- a/stripe/api_resources/abstract/api_resource.py
+++ b/stripe/api_resources/abstract/api_resource.py
@@ -3,17 +3,19 @@ from __future__ import absolute_import, division, print_function
 from stripe import api_requestor, error, util
 from stripe.stripe_object import StripeObject
 from urllib.parse import quote_plus
-from typing import ClassVar
+from typing import ClassVar, Generic, TypeVar, cast
+
+T = TypeVar("T", bound="StripeObject")
 
 
-class APIResource(StripeObject):
+class APIResource(StripeObject, Generic[T]):
     OBJECT_NAME: ClassVar[str]
 
     @classmethod
-    def retrieve(cls, id, api_key=None, **params):
+    def retrieve(cls, id, api_key=None, **params) -> T:
         instance = cls(id, api_key, **params)
         instance.refresh()
-        return instance
+        return cast(T, instance)
 
     def refresh(self):
         return self._request_and_refresh("get", self.instance_url())
--- a/stripe/api_resources/abstract/api_resource.py
+++ b/stripe/api_resources/abstract/api_resource.py
@@ -5,7 +5,7 @@ from stripe.stripe_object import StripeObject
 from urllib.parse import quote_plus
 from typing import ClassVar, Generic, TypeVar, cast
 
-T = TypeVar("T", bound="StripeObject")
+T = TypeVar("T", bound=StripeObject)
 
 
 class APIResource(StripeObject, Generic[T]):
--- a/stripe/api_resources/abstract/createable_api_resource.py
+++ b/stripe/api_resources/abstract/createable_api_resource.py
@@ -1,9 +1,13 @@
 from __future__ import absolute_import, division, print_function
 
 from stripe.api_resources.abstract.api_resource import APIResource
+from typing import TypeVar, cast
+from stripe.stripe_object import StripeObject
+
+T = TypeVar("T", bound="StripeObject")
 
 
-class CreateableAPIResource(APIResource):
+class CreateableAPIResource(APIResource[T]):
     @classmethod
     def create(
         cls,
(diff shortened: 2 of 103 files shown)


stripe/api_resources/abstract/api_resource.py
```diff
@@ -3,17 +3,19 @@
...
+T = TypeVar("T", bound="StripeObject")
 
-class APIResource(StripeObject):
+
+class APIResource(StripeObject, Generic[T]):
     OBJECT_NAME: ClassVar[str]
 
     @classmethod
-    def retrieve(cls, id, api_key=None, **params):
+    def retrieve(cls, id, api_key=None, **params) -> T:
```
pakrym-stripe, line 15: While we are in these files should we type other methods that return `T`? Like `refresh`?

stripe/api_resources/abstract/createable_api_resource.py
```diff
@@ -1,9 +1,13 @@
 from __future__ import absolute_import, division, print_function
 
 from stripe.api_resources.abstract.api_resource import APIResource
+from typing import TypeVar, cast
+from stripe.stripe_object import StripeObject
 
+T = TypeVar("T", bound="StripeObject")
```
pakrym-stripe, line 7: Can we do
```suggestion
T = TypeVar("T", bound=StripeObject)
```

?

stripe/api_resources/abstract/deletable_api_resource.py
```diff
@@ -3,16 +3,23 @@
...
+from typing import TypeVar, cast
+from stripe.stripe_object import StripeObject
 
+T = TypeVar("T", bound="StripeObject")
 
-class DeletableAPIResource(APIResource):
+
+class DeletableAPIResource(APIResource[T]):
     @classmethod
     def _cls_delete(cls, sid, **params):
```
pakrym-stripe, line 14: Type this one as well?

pakrym-stripe: I'm not sure if it helps with all the `class_method_variant` magic.

stripe/api_resources/abstract/test_helpers.py
```diff
@@ -6,9 +6,7 @@
 from typing import TypeVar, ClassVar, Any
 from typing_extensions import Protocol
 
-from stripe.api_resources.abstract.api_resource import APIResource
-
-T = TypeVar("T", bound=APIResource)
```
pakrym-stripe, line 11: Did the bound not work out?

stripe/api_resources/account_link.py
```diff
@@ -5,7 +5,7 @@
 from stripe.api_resources.abstract import CreateableAPIResource
 
 
-class AccountLink(CreateableAPIResource):
+class AccountLink(CreateableAPIResource["AccountLink"]):
```
pakrym-stripe, line 8: Wrong quotes?

pakrym-stripe (changes requested):
minor Qs and nits.

stripe/api_resources/abstract/api_resource.py
richardm-stripe: I want to put a little more thought into `refresh` and would rather it stay out of scope here.

stripe/api_resources/abstract/deletable_api_resource.py
richardm-stripe: Can't hurt.
```suggestion
    def _cls_delete(cls, sid, **params) -> T:
```

stripe/api_resources/abstract/test_helpers.py
richardm-stripe: Sort of. The trouble is that APIResource needs a parameter but I guess `bound=APIResource[Any]` is good enough.

Stylistically, I consider unused bounds on type parameters to be a code smell, but in this case the only reason it's "unused" is because we can't properly type things and have that `_resource_cls: ClassVar[Any]` business, so I changed my mind.

stripe/api_resources/account_link.py
richardm-stripe: Oh 🤦 looks like our formatter converts these. I'll change codegen to match what our formatter is doing.

stripe/api_resources/abstract/test_helpers.py
richardm-stripe: (Fixed)

stripe/api_resources/abstract/api_resource.py
```diff
@@ -3,17 +3,19 @@
 from stripe import api_requestor, error, util
 from stripe.stripe_object import StripeObject
 from urllib.parse import quote_plus
-from typing import ClassVar
+from typing import ClassVar, Generic, TypeVar, cast
 
+T = TypeVar("T", bound="StripeObject")
```
pakrym-stripe, line 8: `bound=StripeObject` here and elsewhere?

pakrym-stripe (changes requested):
There are more occurrences of `bound=StripeObject`

pakrym-stripe (approved):

PR #1043, opened 2023-09-05
Types: ListObject

richardm-stripe:
More precise types for properties that are list objects.

### DIFF
--- a/tests/test_generated_examples.py
+++ b/tests/test_generated_examples.py
@@ -770,7 +770,7 @@ class TestGeneratedExamples(object):
             "/v1/application_fees/fee_xxxxxxxxxxxxx",
         )
 
-    def test_applicationfee_feerefund_list(self, request_mock):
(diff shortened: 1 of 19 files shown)


tests/test_generated_examples.py
```diff
@@ -770,7 +770,7 @@ def test_applicationfee_retrieve(self, request_mock):
             "/v1/application_fees/fee_xxxxxxxxxxxxx",
         )
 
-    def test_applicationfee_feerefund_list(self, request_mock):
+    def test_applicationfee_applicationfeerefund_list(self, request_mock):
```
anniel-stripe, line 773: Is this change expected too?

anniel-stripe (commented):
Makes sense to me! A comment about the generated example test case naming though

tests/test_generated_examples.py
richardm-stripe: Yes, see the codegen PR for an explanation.

anniel-stripe (approved):

PR #1055, opened 2023-09-15
Partially type resource methods (no **params)

richardm-stripe:
Add types for some of the parameters on API Methods (no type for **params) yet.


stripe/api_resources/application_fee.py
```diff
@@ -84,7 +88,7 @@ def _cls_refund(
         )
 
     @util.class_method_variant("_cls_refund")
-    def refund(self, idempotency_key=None, **params):
+    def refund(self, idempotency_key=None, **params: Any):
```
pakrym-stripe, line 91: types missing here.

pakrym-stripe (approved):

PR #1058, opened 2023-09-21
Add various type annotations

richardm-stripe:
Adds type annotations to certain infrastructure methods:
  * The goal: better type for `util.convert_to_stripe_object` (since the team was just talking about this method feels uncomfortable).
    * Had to leave in a couple of `cast` to make the types line up because it's weird and recursive and polymorphic.
  * Typed `.request`, `._request`, `.construct_from` and `.refresh_from` on`StripeObject`, since these are called from `convert_to_stripe_object`.
    * Had to add some asserts in `ListObject` `.list`, `.create`, `.retrieve`, `SearchObject.search` to stop complaints about the new typed `StripeObject._request`

### DIFF
--- a/stripe/util.py
+++ b/stripe/util.py
@@ -185,7 +185,7 @@ def convert_to_stripe_object(
     if isinstance(resp, list):
         return [
             convert_to_stripe_object(
-                cast("StripeObject", i),
+                cast("StripeResponse", i),
                 api_key,
                 stripe_version,
                 stripe_account,
--- a/stripe/util.py
+++ b/stripe/util.py
@@ -185,7 +185,7 @@ def convert_to_stripe_object(
     if isinstance(resp, list):
         return [
             convert_to_stripe_object(
-                cast("Union[StripeObject, Dict[str, Any]]", i),
+                cast("Union[StripeResponse, Dict[str, Any]]", i),
                 api_key,
                 stripe_version,
                 stripe_account,
--- a/stripe/util.py
+++ b/stripe/util.py
@@ -11,8 +11,21 @@ import re
 import stripe
 from urllib.parse import parse_qsl, quote_plus
 
-from typing_extensions import Type
-from typing import Union, overload, Dict, cast, Any, Optional
+from typing_extensions import Type, TYPE_CHECKING
+from typing import (
+    Union,
+    overload,
+    Dict,
+    List,
+    cast,
+    Any,
+    Optional,
+)
+
+from stripe.stripe_response import StripeResponse
+
+if TYPE_CHECKING:
+    from stripe.stripe_object import StripeObject
 
 STRIPE_LOG = os.environ.get("STRIPE_LOG")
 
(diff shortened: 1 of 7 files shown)


stripe/util.py
```diff
@@ -128,9 +141,38 @@ def get_object_classes():
...
+    stripe_version: Optional[str] = None,
+    stripe_account: Optional[str] = None,
+    params: Optional[Dict[str, Any]] = None,
+) -> "StripeObject":
+    ...
+
+
+@overload
+def convert_to_stripe_object(
+    resp: List[Resp],
```
pakrym-stripe, line 160: How is `List[Resp]` different from Resp that already is a `List[Resp]`?

stripe/util.py
```diff
@@ -143,7 +185,10 @@ def convert_to_stripe_object(
     if isinstance(resp, list):
         return [
             convert_to_stripe_object(
-                i, api_key, stripe_version, stripe_account
+                cast("StripeObject", i),
```
pakrym-stripe, line 188: Why are we casting to StripeObject? Isn't StripeObject the output of convert_to_stripe_object instead of the input?

pakrym-stripe (changes requested):
Some Qs

stripe/util.py
richardm-stripe: It's more specific. This overload means "if you pass a list of something in, you are guaranteed to get a list of something out", whereas the type signature in the implementation says "if you pass a Resp in (which may or may not be a list) you will get a Resp out (which may or may not be a list)".

stripe/util.py
richardm-stripe: Good catch, this was supposed to be "StripeResponse". At first glance I was mildly concerned that `StripeObject` was typechecking, but then I recalled that `StripeObject` extends from `Dict[str, Any]` which is one of the branches of `Resp`.

stripe/util.py
```diff
@@ -143,7 +185,10 @@ def convert_to_stripe_object(
     if isinstance(resp, list):
         return [
             convert_to_stripe_object(
-                i, api_key, stripe_version, stripe_account
+                cast("StripeResponse", i),
```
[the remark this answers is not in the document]
richardm-stripe: That won't typecheck because it needs to not come back as list. Could do
`Union[StripeObject, Dict[str, Any]]` though.

stripe/util.py
```diff
@@ -182,7 +227,7 @@ def convert_to_stripe_object(
 
         return obj
     else:
-        return resp
+        return cast("StripeObject", resp)
```
pakrym-stripe, line 230: Is there some test-only assert we can add? Would be nice to verify that this is true in tests but without affecting published library runtime.

stripe/util.py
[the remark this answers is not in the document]
richardm-stripe: (StripeResponse, lol)

pakrym-stripe (approved):

stripe/util.py
richardm-stripe: Hmm, I'll noodle on this. I think we have to build our own.

PR #1061, opened 2023-09-23
[WIP] Types: inner resource classes

richardm-stripe:


stripe/stripe_object.py
```diff
@@ -225,12 +223,31 @@ def refresh_from(
...
+                    k: util.convert_to_stripe_object(
+                        v,
+                        api_key,
+                        stripe_version,
+                        stripe_account,
+                        None,
+                        inner_class,
+                    )
+                    for k, v in v.items()
+                    if v is not None
```
pakrym-stripe, line 239: are we completely filtering out Nones?

stripe/stripe_object.py
```diff
@@ -225,12 +223,31 @@ def refresh_from(
...
-            super(StripeObject, self).__setitem__(
-                k,
-                util.convert_to_stripe_object(
-                    v, api_key, stripe_version, stripe_account
-                ),
-            )
+            inner_class = self._get_inner_class_type(k)
+            is_dict = self._get_inner_class_is_beneath_dict(k)
+            if is_dict:
+                obj = {
```
pakrym-stripe, line 229: Are we ok with this "breaking" change of going from StripeObject to dict?

stripe/stripe_object.py
```diff
@@ -417,3 +434,12 @@ def __deepcopy__(self, memo):
             super(StripeObject, copied).__setitem__(k, deepcopy(v, memo))
 
         return copied
+
+    _inner_class_types: ClassVar[Dict[str, Type["StripeObject"]]] = {}
```
pakrym-stripe, line 438: Optional/NonRequired?

stripe/stripe_object.py
```diff
@@ -417,3 +434,12 @@ def __deepcopy__(self, memo):
...
+    _inner_class_dicts: ClassVar[List[str]] = []
+
+    def _get_inner_class_type(self, field_name: str):
+        return self._inner_class_types.get(field_name)
+
+    def _get_inner_class_is_beneath_dict(self, field_name: str):
+        return field_name in self._inner_class_dicts
```
pakrym-stripe, line 445: Won't this fail if _inner_class_dicts is None?

stripe/stripe_object.py
```diff
@@ -417,3 +434,12 @@ def __deepcopy__(self, memo):
             super(StripeObject, copied).__setitem__(k, deepcopy(v, memo))
 
         return copied
+
+    _inner_class_types: ClassVar[Dict[str, Type["StripeObject"]]] = {}
+    _inner_class_dicts: ClassVar[List[str]] = []
+
+    def _get_inner_class_type(self, field_name: str):
+        return self._inner_class_types.get(field_name)
```
pakrym-stripe, line 442: Won't this fail if _inner_class_types is None?

pakrym-stripe (changes requested):
Qs

stripe/stripe_object.py
richardm-stripe: It won't be, `inner_class_types` defaults (on the parent class) to an empty dict.

stripe/stripe_object.py
richardm-stripe: No, required but always defined on the parent class leaves fewer edge cases, I think.

stripe/stripe_object.py
richardm-stripe: It won't be, `_inner_class_dicts` defaults to an empty array on the parent class.

pakrym-stripe (approved):
I should learn to read.

stripe/stripe_object.py
richardm-stripe: Oh, great point. The blast radius is sooo small here (there's only one of these) and in my estimation the chance of a user relying on StripeObjectness and not just using as a dict is miniscule.

stripe/stripe_object.py
richardm-stripe: Yes, probably doesn't make a difference but I will fix.

stripe/stripe_object.py
richardm-stripe: Added a note to the PR description.

richardm-stripe: (Not going to change now for the beta merge)

stripe/stripe_object.py
richardm-stripe: Fixed in [Don't filter out Nones](https://github.com/stripe/stripe-python/pull/1061/commits/8b1d5cd56c2818b601acc8d142e59c5af4153a9f).

PR #1069, opened 2023-10-11
Remove `__future__` imports and use generated sections in `object_classes.py`

anniel-stripe:
This PR:
1. Removes all `from __future__ import absolute_import, division, print_function` imports. These imports are no longer necessary, as they are used for forwards compatibility with Python 2.x versions and became the default behavior in Python 3.0 (see https://docs.python.org/3/library/__future__.html). Since stripe-python supports Python >=3.6 as of [v6.0.0](https://github.com/stripe/stripe-python/releases/tag/v6.0.0), this is a non-breaking change.
2. Updates `stripe/object_classes.py` to use generated sections, allowing us to manually maintain data structure objects like  `ListObject` (see [commit](https://github.com/stripe/stripe-python/pull/1069/commits/b6b42b2d083f0e120a8a6e13e2361b735e13b348)).


stripe/api_resources/bank_account.py
```diff
@@ -33,6 +31,139 @@ class BankAccount(
     """
 
     OBJECT_NAME = "bank_account"
+
```
pakrym-stripe, line 34: unrelated?

pakrym-stripe (changes requested):
A lot of unrelated changes.

pakrym-stripe (approved):

PR #1077, opened 2023-10-12
Add various type annotations

richardm-stripe:
Adds missing type annotations discovered while manually using types in example projects.


pakrym-stripe (approved):
❤️ ❤️

PR #1078, opened 2023-10-13
Parameter types

richardm-stripe:
Add `**params: Unpack[FooCreateParams]` and corresponding Params class definitions to resource methods.


.github/workflows/ci.yml
```diff
@@ -95,6 +95,11 @@ jobs:
 
       - name: Typecheck with pyright
         run: make pyright
+        if: matrix.python-version != '3.6'
+
+      - name: Typecheck with pyright
```
pakrym-stripe, line 100: Just skip it?

richardm-stripe: It's fairly useful to run this to make sure that we're not importing things from `typing` that need to be imported from `typing_extensions` in python 3.6. In most cases the test suite would crash, but this would be necessary for `if TYPE_CHECKING`

stripe/api_resources/apps/secret.py
```diff
@@ -10,6 +10,8 @@
 from typing import List, Optional, cast
 from typing_extensions import Literal, NotRequired, TypedDict, Unpack
 
+from typing_extensions import TYPE_CHECKING
```
pakrym-stripe, line 13: ungrouped imports?

pakrym-stripe (approved):
Lovely

.github/workflows/ci.yml
pakrym-stripe: Mind adding a comment about why we have that condition?

stripe/api_resources/apps/secret.py
richardm-stripe: fixed

PR #1087, opened 2023-10-18
More type annotations

richardm-stripe:
After this, `pyright --verifytypes stripe` doesn't report any of the "could be inferred differently by type checkers" messages, e.g. this message goes away:

```
error.py:27:14 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "Unknown | None"
```

Moves pyright's "Type completeness score" from 94.2% -> 94.3%

### DIFF
--- a/stripe/api_requestor.py
+++ b/stripe/api_requestor.py
@@ -3,7 +3,7 @@ import datetime
 import json
 import platform
 import time
-from typing import Tuple
+from typing import Optional, Tuple
 import uuid
 import warnings
 from collections import OrderedDict
@@ -13,6 +13,8 @@ from stripe import error, oauth_error, http_client, version, util
 from stripe.multipart_data_generator import MultipartDataGenerator
 from urllib.parse import urlencode, urlsplit, urlunsplit
 from stripe.stripe_response import StripeResponse, StripeStreamResponse
+
+from stripe.http_client import HTTPClient
 
 
 def _encode_datetime(dttime):
@@ -65,6 +67,11 @@ def _build_api_url(url, query):
 
 
 class APIRequestor(object):
+    api_key: Optional[str]
+    api_base: str
+    api_version: str
+    stripe_account: Optional[str]
+
     def __init__(
         self,
         key=None,
--- a/stripe/api_requestor.py
+++ b/stripe/api_requestor.py
@@ -282,8 +282,7 @@ class APIRequestor(object):
             headers["Content-Type"] = "application/x-www-form-urlencoded"
             headers.setdefault("Idempotency-Key", str(uuid.uuid4()))
 
-        if self.api_version is not None:
-            headers["Stripe-Version"] = self.api_version
+        headers["Stripe-Version"] = self.api_version
 
         return headers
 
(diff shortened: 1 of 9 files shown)


stripe/api_requestor.py
```diff
@@ -275,8 +282,7 @@ def request_headers(self, api_key, method):
             headers["Content-Type"] = "application/x-www-form-urlencoded"
             headers.setdefault("Idempotency-Key", str(uuid.uuid4()))
 
-        if self.api_version is not None:
-            headers["Stripe-Version"] = self.api_version
+        headers["Stripe-Version"] = self.api_version
```
richardm-stripe, line 285: See line 85. `stripe.api_version` is set by default (now because of version pinning), and so self.api_version will always be set. Pyright was complaining about this if statement being redundant, so I removed it.

pakrym-stripe: Do we support setting `stripe.api_version` to None?

pakrym-stripe (approved):
Consider typing constructors for the classes in which you've added fields.

stripe/api_requestor.py
richardm-stripe: * The types prohibit it.
* Even if it is `None` here, setting `headers["Stripe-Version"]` explicitly to None ultimately has the same result, I tested
```
>>> import stripe
>>> stripe.api_key = "xyz"
>>> stripe.api_version = None
>>> a = stripe.api_requestor.APIRequestor()
>>> a.api_base = "http://localhost:4444"
>>> a.request_raw("post", "/foo", {"bar": "baz"})
```

```
$ nc -l 4444
POST /foo HTTP/1.1
Host: localhost:4444
User-Agent: Stripe/v1 PythonBindings/7.0.0
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
X-Stripe-Client-User-Agent: ...
Authorization: Bearer xyz
Content-Type: application/x-www-form-urlencoded
Idempotency-Key: 8669fffb-835d-479d-be81-3c93bc26ea63
Content-Length: 7

bar=baz
```

stripe/util.py
```diff
@@ -178,7 +178,7 @@ def convert_to_stripe_object(
 
     if isinstance(resp, stripe.stripe_response.StripeResponse):
         stripe_response = resp
-        resp = stripe_response.data
+        resp = cast(Resp, stripe_response.data)
```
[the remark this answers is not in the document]
richardm-stripe: We're reassigning `resp` to `resp.data` basically. Even if `resp` is `stripe.stripe_response.StripeResponse` that doesn't guarantee that `resp.data` will be -- in fact, it's probably not.

stripe/stripe_response.py
```diff
@@ -1,8 +1,13 @@
...
 import json
 from collections import OrderedDict
+from typing import Dict
 
 
 class StripeResponseBase(object):
+    code: int
+    headers: Dict[str, str]
+
     def __init__(self, code, headers):
```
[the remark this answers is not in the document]
richardm-stripe: Done, except for `stripe_object` which I have in a follow-up.

pakrym-stripe (approved):

PR #1088, opened 2023-10-18
Even more type annotations

richardm-stripe:
In this one, the goal is to fully type `StripeObject`. I add `# pyright: strict` to the top, and also make sure every method has a return annotation. Now, `pyright --verifytypes` no longer complains about `Type of base class "stripe.stripe_object.StripeObject" is partially unknown` for the (many) classes that inherit from StripeObject.

The really hard thing, that I haven't been ready to tackle yet, is coming up with good types for `stripe.http_client`, so there are some pretty unsatisfying `Any` and casts in this one, but I'm hoping that adding types around the periphery will make it easier to tackle eventually.

Pyright reports
```
Type completeness score: 96.1%
```
after this (up from 94.3%)


stripe/api_requestor.py
```diff
@@ -286,16 +311,20 @@ def request_headers(self, api_key, method):
...
+        url: str,
+        params: Optional[Mapping[str, Any]] = None,
+        supplied_headers: Optional[Mapping[str, str]] = None,
+        is_streaming: bool = False,
+    ) -> Tuple[object, int, Mapping[str, str], str]:
         """
         Mechanism for issuing an API call
         """
 
+        supplied_headers_: Optional[Dict[str, str]] = (
```
pakrym-stripe, line 324: Should we type the parameter as dict?

pakrym-stripe, line 324:
```suggestion
        supplied_header_dict: Optional[Dict[str, str]] = (
```

stripe/api_requestor.py
```diff
@@ -419,6 +456,9 @@ def interpret_streaming_response(self, stream, rcode, rheaders):
                     "can be consumed when streaming a response."
                 )
 
-            return self.interpret_response(json_content, rcode, rheaders)
+            self.interpret_response(json_content, rcode, rheaders)
+            raise RuntimeError(
```
pakrym-stripe, line 460: `-> Never` in interpret_response?

stripe/api_requestor.py
richardm-stripe: `Mapping[str, str]` is really a better type. We want to guarantee that the implementation won't mutate the headers that the caller passes in. And also the caller should be able to pass immutable mappings too.

We have to create this temporary dict because (surprise!) we actually *were* mutating. So this does constitute a (almost certainly welcome) behavior change. I should test this out and call it out in the PR description, though.

stripe/stripe_object.py
```diff
@@ -1,17 +1,46 @@
...
+
+@overload
+def _compute_diff(
+    current: object, previous: Optional[Dict[str, Any]]
+) -> object:
+    ...
+
+
+def _compute_diff(
+    current: object, previous: Optional[Dict[str, Any]]
```
pakrym-stripe, line 40: You can make `_compute_diff` generic  based on type of `current`

pakrym-stripe: Don't know if it helps.

stripe/api_requestor.py
richardm-stripe: `intepret_response` doesn't throw always, only if `self._should_handle_code_as_error(rcode)` which we have checked here.

stripe/stripe_object.py
richardm-stripe: Yeah I went down that road a little bit, but hit my timeout. This is fairly internal so the `Any` shouldn't leak to users.

stripe/stripe_object.py
```diff
@@ -51,17 +84,17 @@ def default(self, obj):
...
-        stripe_version=None,
-        stripe_account=None,
-        last_response=None,
-        **params
+        id: Optional[str] = None,
+        api_key: Optional[str] = None,
+        stripe_version: Optional[str] = None,
+        stripe_account: Optional[str] = None,
+        last_response: Optional[StripeResponse] = None,
+        **params: Any
```
pakrym-stripe, line 92: _retrieve_params is Dict[str, Any] can we type params the same?

stripe/stripe_object.py
```diff
@@ -78,20 +111,22 @@ def __init__(
...
         return super(StripeObject, self).update(update_dict)
 
-    def __setattr__(self, k, v):
-        if k[0] == "_" or k in self.__dict__:
-            return super(StripeObject, self).__setattr__(k, v)
+    if not TYPE_CHECKING:
 
-        self[k] = v
-        return None
+        def __setattr__(self, k, v):
```
pakrym-stripe, line 124: I thought having these under `TYPE_CHECKING` didn't matter for type checking

stripe/stripe_object.py
```diff
@@ -152,13 +187,13 @@ def __delitem__(self, k):
...
     # string
-    def __setstate__(self, state):
+    def __setstate__(self, state: Dict[str, Any]) -> None:
         self.update(state)
 
     # Custom pickling method to ensure the instance is pickled as a custom
     # class and not as a dict, otherwise __setstate__ would not be called when
     # unpickling.
-    def __reduce__(self):
+    def __reduce__(self) -> Any:
```
pakrym-stripe, line 196:
```suggestion
    def __reduce__(self) -> Tuple[type, Tuple, dict]:
```

stripe/stripe_object.py
```diff
@@ -240,16 +275,16 @@ def refresh_from(
...
 
     def request(
         self,
         method: Literal["get", "post", "delete"],
         url: str,
         params: Optional[Dict[str, Any]] = None,
-        headers=None,
-    ):
+        headers: Optional[Dict[str, str]] = None,
+    ) -> "StripeObject":
```
pakrym-stripe, line 287: Self?

stripe/api_requestor.py
```diff
@@ -386,11 +415,17 @@ def request_raw(
...
+        self, rbody: object, rcode: int, rheaders: Mapping[str, str]
+    ) -> StripeResponse:
         try:
-            if hasattr(rbody, "decode"):
-                rbody = rbody.decode("utf-8")
-            resp = StripeResponse(rbody, rcode, rheaders)
+            resp = StripeResponse(
+                # TODO: should be able to remove this cast once self._client.request_with_retries
+                # returns a more specific type.
+                cast(Any, rbody).decode("utf-8"),
```
pakrym-stripe, line 425: Should we at least case to bytes?

pakrym-stripe (approved):

stripe/stripe_object.py
richardm-stripe: It does. I should have called this out in the PR description.

This will cause

`customer.this_property_does_not_exist = "foo".

to begin producing a type error. I think when we were originally working on the magic getter/setter we didn't think this would matter much, but this does matter for the "mutate and call `.save()`" use case, and we got some feedback from internal users about this (see #developer-sdks).

I'm not wedded to this but I don't see too much downside.

stripe/stripe_object.py
richardm-stripe: I think we should do something here, but you get 300 type errors if you try this, because you have to handle all the unpacked `Params` typedDicts here. I'll add a TODO.

stripe/stripe_object.py
richardm-stripe: `Tuple` needs type variables, so that won't quite work.

We can do 
```
    def __reduce__(self) -> Tuple[Any, ...]:
``` though, that's close to what they have in typeshed: https://github.com/python/typeshed/issues/6661

File: stripe/api_requestor.py (hunk 8 above)
Comment (richardm-stripe) [reply stripe/api_requestor.py]: Done in [Address code review](https://github.com/stripe/stripe-python/pull/1088/commits/067ab1bfd6a74acd7cb4bb8c4f7e878a2d0e4661)

File: stripe/stripe_object.py (hunk 6 above)
Comment (richardm-stripe) [reply stripe/stripe_object.py]: Fixed in [Address code review](https://github.com/stripe/stripe-python/pull/1088/commits/067ab1bfd6a74acd7cb4bb8c4f7e878a2d0e4661)

File: stripe/stripe_object.py [hunk 9]
```diff
@@ -51,17 +84,17 @@ def default(self, obj):
...
+        stripe_version: Optional[str] = None,
+        stripe_account: Optional[str] = None,
+        last_response: Optional[StripeResponse] = None,
+        **params: Any
     ):
         super(StripeObject, self).__init__()
 
-        self._unsaved_values = set()
-        self._transient_values = set()
+        self._unsaved_values: Set[Any] = set()
```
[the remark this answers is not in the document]
richardm-stripe: Fixed in [Address code review](https://github.com/stripe/stripe-python/pull/1088/commits/067ab1bfd6a74acd7cb4bb8c4f7e878a2d0e4661)

stripe/stripe_object.py
richardm-stripe: TODO added in [Address code review](https://github.com/stripe/stripe-python/pull/1088/commits/067ab1bfd6a74acd7cb4bb8c4f7e878a2d0e4661)

stripe/api_requestor.py
richardm-stripe: Added a comment in [Address code review](https://github.com/stripe/stripe-python/pull/1088/commits/067ab1bfd6a74acd7cb4bb8c4f7e878a2d0e4661)

stripe/api_requestor.py
richardm-stripe: Great idea, done in [Address code review](https://github.com/stripe/stripe-python/pull/1088/commits/067ab1bfd6a74acd7cb4bb8c4f7e878a2d0e4661)

richardm-stripe: Pulled it out to a separate PR https://github.com/stripe/stripe-python/pull/1091

But it looks like this might actually be load-bearing mutation... I should probably refresh my understanding of how multipart content-type works.

stripe/stripe_object.py
pakrym-stripe: This is a terrible method.

richardm-stripe: Truth

PR #1091, opened 2023-10-18
APIRequestor: don't mutate incoming multipart headers

richardm-stripe:
Pulled out from #1088 since it's a behavior change.

Also fixes the `file` tests. Unfortunately the special multipart form behavior is implemented in api_requestor, but our mocks mock at the api_requestor level so in order to actually make assertions about multipart you have to mock at the http_client level.

These tests were accidentally working because of the unintended mutation. Stopping the mutation broke them, but required the tests to change.

FileUpload is an alias for File, so I deleted `file_upload_test.py` as it's basically completely redundant.


tests/api_resources/test_file.py
```diff
@@ -28,7 +28,11 @@ def test_is_retrievable(self, request_mock):
         request_mock.assert_requested("get", "/v1/files/%s" % TEST_RESOURCE_ID)
         assert isinstance(resource, stripe.File)
 
-    def test_is_creatable(self, setup_upload_api_base, request_mock):
+    def test_is_creatable(self, mocker):
+        hc = mocker.Mock(stripe.http_client.HTTPClient)
```
pakrym-stripe, line 32: don't we have an existing fixture for this something like this?

richardm-stripe: Not in master :'(

tests/api_resources/test_file.py
```diff
@@ -38,14 +42,19 @@ def test_is_creatable(self, setup_upload_api_base, request_mock):
...
+        assert method == "post"
+        assert url.endswith("/v1/files")
+        assert (
+            headers["Content-Type"]
+            == "multipart/form-data; boundary=1234567890"
         )
+        parts = body.split(b"--1234567890")
+        assert len(parts) == 5
+        assert parts[0] == b""
+        assert parts[1].find(b'name="purpose"') > 0
```
pakrym-stripe, line 55: `assert b'name="purpose"' in parts[1]` ? here and elsewhere

richardm-stripe: Done!

pakrym-stripe (approved):

PR #1092, opened 2023-10-18
Beta: export raw request, deserialize, and preview in a better way

richardm-stripe:
These symbols were not considered to be exported by the type checker

https://github.com/microsoft/pyright/blob/cecb3e1e0cfd6dd504655ba678332f3399e218dc/docs/typed-libraries.md#library-interface

Imported symbols are not considered exported unless they are redundant (or use that odd `from . import ...` syntax). So you want to export an imported symbol under a different name, you have to explicitly define the new name outside of the import statement.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -47,6 +47,8 @@ from stripe.webhook import Webhook, WebhookSignature  # noqa
 
 from stripe.raw_request import _raw_request  # noqa
 
+# The type checker won't re-export imported symbols unless they are redundant,
+# so we need to re-export them manually.
 raw_request = _raw_request
 
 from stripe.raw_request import _deserialize  # noqa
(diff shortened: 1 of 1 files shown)


stripe/__init__.py
```diff
@@ -45,11 +45,16 @@
 # Webhooks
 from stripe.webhook import Webhook, WebhookSignature  # noqa
 
-from stripe.raw_request import _raw_request as raw_request  # noqa
+from stripe.raw_request import _raw_request  # noqa
 
-from stripe.raw_request import _deserialize as deserialize  # noqa
+raw_request = _raw_request
```
pakrym-stripe, line 50: Mind adding a comment as to why we are doing this?

pakrym-stripe: Can we add a test? Or enable typecheking in some test?

pakrym-stripe (approved):

stripe/__init__.py
richardm-stripe: Done!

richardm-stripe: (Oh, done for the comment. Re: test, I think that's a good idea, will think about that soon.)

PR #1094, opened 2023-10-24
Remove file that shouldn't have been committed

antoineleclair:
A git conflict `.orig` file was commited but shouldn't have.


anniel-stripe (approved):
Thank you for catching this @antoineleclair !

PR #1095, opened 2023-10-24
Ignore .orig files

anniel-stripe:
To prevent files like in https://github.com/stripe/stripe-python/pull/1094 from accidentally getting committed (we should also make sure this setting is false locally `git config --global mergetool.keepBackup false`).


richardm-stripe:
My bad!

richardm-stripe (approved):

PR #1100, opened 2023-10-25
Fix class method variant types

richardm-stripe:
`@util.class_method_variant("...")` was basically swallowing the types of custom methods and delete methods.

This PR:

* Changes the decorator to preserve the type of the method it wraps. (Suppresses some type errors, it is hard to type the implementation of the decorator in a satisfying way)
* Adds correctly-typed overloads to all the methods that use the decorator.
* Suppresses a type error on the implementation complaining about how it doesn't match the overloads. The type here shouldn't matter, users will only experience the type of the overloads.

### DIFF
--- a/stripe/util.py
+++ b/stripe/util.py
@@ -366,9 +366,13 @@ class class_method_variant(object):
     def __init__(self, class_method_name):
         self.class_method_name = class_method_name
 
-    def __call__(self, method):
+    T = TypeVar("T")
+
+    method: Any
+
+    def __call__(self, method: T) -> T:
         self.method = method
-        return self
+        return cast(T, self)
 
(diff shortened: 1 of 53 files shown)


stripe/util.py
```diff
@@ -366,9 +366,13 @@ class class_method_variant(object):
     def __init__(self, class_method_name):
         self.class_method_name = class_method_name
 
-    def __call__(self, method):
+    T = TypeVar("T")
+
+    method: Any
```
pakrym-stripe, line 371: Callable?

pakrym-stripe (approved):

stripe/util.py
richardm-stripe: That causes problems in the implementation of `__call__` and I timed out trying to figure out how to resolve.

PR #1101, opened 2023-10-26
Mark defunct and internal methods as deprecated

richardm-stripe:
## Summary
This PR adds deprecation annotations to classes and methods that, in my judgement, we should phase out -- at least for external users.

For things we want to keep internally but deprecate externally, I took two different approaches in different situations.
1. In some cases, I added a deprecation notice, introduced an internal (underscore-prefixed) version, and moved all internal usages to the underscore-prefixed copy.
2. In other cases, I just added the deprecation notice and added a `with warnings.catch_warnings():` clause around the internal use, so that errors won't bubble down to users.

Option 2 is more conservative. Both options preserve the behavior of the function itself. However, with Option 1 could cause a behavior difference if somebody had a child class, overrode the method in question, and relied on their overloaded version to be called by the internal code in the parent class. I've used option 2 in circumstances where that was a concern.

### Internal Classes
  * `StripeObject.ReprJSONEncoder` - Used in `StripeObject.__str__` for custom encoding datetimes. No external use case.

### Deprecated methods
  * `APIResource.retrieve`,  - Superceded by explicit overrides on the child class.
  * `ListObject.retrieve`, `ListObject.create` - odd methods that use the `url` from the ListObject to construct post and get requests. There is no guarantee that the API will actually support those methods, and I think it's better for users to be explicit about which API method they are calling.
  * `StripeObject.api_base` - I don't think this has been used since [2018](https://github.com/stripe/stripe-python/pull/477/files#diff-40fe656a0339a07dbeb1ba940bbaba6e663021442cf9d0addc00f73262752dd5L12)

### Internal methods
  * `APIRequestor.format_app_info` - internal helper used for formatting the user agent string.
  * `ListObject.list`, `SearchResultObject.search` - Like `ListObject.retrieve` and `ListObject.create` -- makes a request based off of the `url` property. This one is used internally by pagination, but for external users it's better if they either auto-paginate or call the specific `list` method explicitly.
  * `ListObject.empty_list`, `SearchResultObject.empty_search_result` - internal helper for pagination, no real legitimate external use.
  * `StripeError.construct_error_object` - internal helper for deserializing errors.
  * `StripeObject.to_dict_recursive` - internal helper for StripeObject.__str__
  * `StripeObject.stripe_id` - internal helper for APIRequestor._api_encode

### Not deprecated (yet)
  * `ListableAPIResource`, `CreateableAPIResource`, `DeleteableAPIResource`, `SearchableAPIResource`, `SingletonAPIResource`, `UpdateableAPIResource`, `@custom_method`, `@nested_resource_class_methods` - These classes provide CRUDL methods that have all been superceded by overrides on the resources that inherit from them, but they are still in use by other parts of the library, so it's difficult to prevent DeprecationWarning from being bubbled up when users use those other, non-deprecated parts of the library that use these features.
  * `ListableResource.auto_paging_iter` - I almost deprecated this one. `Customer.auto_paging_iter(...)` is just shorthand for `Customer.list(...).auto_paging_iter()` and feels unnecessary and non-standard, but I don't think preserving it is too harmful and it's not worth forcing users to change.


stripe/util.py
```diff
@@ -34,26 +35,32 @@
...
     "logfmt",
-    "deprecated",
+    "parse_qsl",
+    "_log_info",
+    "_log_debug",
+    "_dashboard_link",
+    "_deprecated",
 ]
 
-if hasattr(typing_extensions, "deprecated"):
```
richardm-stripe, line 46: This `hasattr` thing was actually interfering with IDE support, and it was too conservative. Even if you have an older version of `typing_extensions` without "deprecated"
* It won't crash, because `if TYPE_CHECKING`.
* It turns out that *typeshed* actually has a type definition for `typing_extensions.deprecated` which will be used even if your `typing_extensions` version is old. So as long as your typechecker version is relatively recent/includes a relatively recent version of typeshed, then you're good to go.

stripe/_api_requestor.py
```diff
@@ -19,6 +19,7 @@
 from stripe import oauth_error  # noqa: SPY101
 from stripe._encode import _api_encode
 from stripe._multipart_data_generator import MultipartDataGenerator
+
```
richardm-stripe, line 22:
```suggestion
```

stripe/_list_object.py
```diff
@@ -76,6 +84,9 @@ def create(
             ),
         )
 
+    @_util.deprecated(
+        "This will be removed in a future version of stripe-python. Please call the `retrieve` method on the corresponding resource directly, instead of using `create` from the list object."
```
richardm-stripe, line 88:
```suggestion
        "This will be removed in a future version of stripe-python. Please call the `retrieve` method on the corresponding resource directly, instead of using `retrieve` from the list object."
```

tests/test_api_requestor.py
```diff
@@ -11,7 +11,7 @@
 from stripe._stripe_response import StripeResponse, StripeStreamResponse
 from stripe._encode import _api_encode
 
-from urllib.parse import urlsplit
+from urllib.parse import urlsplit, parse_qsl
```
richardm-stripe, line 14: left over from before we made _util completely private. But there's no reason why we are re-exporting parse_qsl from util.

stripe/_util.py
```diff
@@ -25,16 +25,18 @@
...
+
+if TYPE_CHECKING:
+    from stripe._stripe_response import StripeResponse
+    from stripe._stripe_object import StripeObject
 
 STRIPE_LOG = os.environ.get("STRIPE_LOG")
 
 logger: logging.Logger = logging.getLogger("stripe")
 
-if hasattr(typing_extensions, "deprecated"):
```
richardm-stripe, line 35: This `hasattr(` thing actually prevents proper typechecking. `if TYPE_CHECKING` is enough to prevent runtime crashes, and it turns out that even in cases where `typing_extensions` is too old, Typeshed will still define it and allow the type checker to work.

stripe/_list_object.py
```diff
@@ -222,7 +251,7 @@ def previous_page(
         params_with_filters.update({"ending_before": first_id})
         params_with_filters.update(params)
 
-        result = self.list(
+        result = self.list(  # pyright: ignore[reportDeprecated]
```
pakrym-stripe, line 254: Will this fire DeprecatedWarning for users even though they use a non-deprecated method?

stripe/_listable_api_resource.py
```diff
@@ -5,6 +5,9 @@
 
 T = TypeVar("T", bound=StripeObject)
 
+# TODO(major): 1704 - remove this class and all internal usages. `.list` is already inlined into the resource classes.
+# Although we should inline .auto_paging_iter into the resource classes as well.
```
pakrym-stripe, line 9: Should we do this earlier than later?

stripe/_nested_resource_class_methods.py
```diff
@@ -4,8 +4,8 @@
 from stripe._api_resource import APIResource
 
 
-# TODO(major): Remove this. It is no longer used except for "nested_resource_url" and "nested_resource_request",
-# which are unnecessary ande deprecated.
+# TODO(major): 1704. Remove this. It is no longer used except for "nested_resource_url" and "nested_resource_request",
+# which are unnecessary and deprecated and should also be removed.
 def nested_resource_class_methods(
```
pakrym-stripe, line 9: Deprecated annotation?

stripe/_list_object.py
richardm-stripe: Oh, good catch. Need to wrap like above.

stripe/_listable_api_resource.py
richardm-stripe: Yes, I'll ticket.

stripe/_util.py
```diff
@@ -232,6 +234,10 @@ def convert_to_stripe_object(
     # the raw API response information
     stripe_response = None
 
+    # Imports here at runtime to avoid circular dependencies
+    from stripe._stripe_response import StripeResponse
```
pakrym-stripe, line 238: Is this a new circular dependency? Should we move deprecation into it's own file?

stripe/_error.py
```diff
@@ -43,7 +46,9 @@ def __init__(
         self.headers = headers or {}
         self.code = code
         self.request_id = self.headers.get("request-id", None)
-        self.error = self.construct_error_object()
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", DeprecationWarning)
+            self.error = self.construct_error_object()
```
pakrym-stripe, line 51: nit: create a private version and use here?

stripe/_list_object.py
```diff
@@ -190,13 +217,15 @@ def next_page(
...
-        result = self.list(
-            api_key=api_key,
-            stripe_version=stripe_version,
-            stripe_account=stripe_account,
-            **params_with_filters,
-        )
-        return result
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", category=DeprecationWarning)
+            result = self.list(  # pyright: ignore[reportDeprecated]
```
pakrym-stripe, line 222: What are we going to do here instead?

stripe/_error.py
richardm-stripe: Ahhh, you mean a private version that call the public version but swallows the error? I didn't think of that, I'll make that change throughout. It will make the follow-up easier.

pakrym-stripe: Hm, I meant what you did with empty_list

```
 def empty_list(
        cls,
        api_key: Optional[str] = None,
        stripe_version: Optional[str] = None,
        stripe_account: Optional[str] = None,
    ) -> Self:
        return cls._empty_list(
            api_key=api_key,
        )

    @classmethod
    def _empty_list(
        cls,
        api_key: Optional[str] = None,
        stripe_version: Optional[str] = None,
        stripe_account: Optional[str] = None,
    ) -> Self:
```

Add _construct_error_object with an implementation, call it from deprecated construct_error_object but call _construct_error_object internally.

stripe/_api_resource.py
```diff
@@ -28,6 +29,9 @@ class APIResource(StripeObject, Generic[T]):
     OBJECT_NAME: ClassVar[str]
 
     @classmethod
+    @_util.deprecated(
```
pakrym-stripe, line 32: What about `class_url`, `instance_url` ?

stripe/_error.py
richardm-stripe: Ok, I'll make the change. (This is option 1 vs. option 2 in my PR description, by the way) I don't think it makes a whole lot of sense to be worried about subclasses here.

stripe/_util.py
richardm-stripe: `util` and `stripe_object` are already circular but no errors happened because existing usages of util inside `stripe_object` were not top-level.

IMO `deprecated` belongs in `util` and it is `convert_to_stripe_object` that we should consider moving elsewhere. I will make a ticket.

stripe/_list_object.py
richardm-stripe: Rename it to `_list`, but in [Address review](https://github.com/stripe/stripe-python/pull/1101/commits/cc8e237402eb70696bad6cd98ecb21e2093db4db) I went ahead and just defined `_list` already.

stripe/_error.py
richardm-stripe: Done in [Address review](https://github.com/stripe/stripe-python/pull/1101/commits/cc8e237402eb70696bad6cd98ecb21e2093db4db)

stripe/_stripe_object.py
```diff
@@ -407,23 +427,26 @@ def __repr__(self) -> str:
...
         )
 
+    @_util.deprecated(
+        "Deprecated. The public interface will be removed in a future version."
+    )
     def to_dict(self) -> Dict[str, Any]:
         return dict(self)
 
-    def to_dict_recursive(self) -> Dict[str, Any]:
+    def _to_dict_recursive(self) -> Dict[str, Any]:
```
pakrym-stripe, line 442: no public overload?

stripe/_stripe_object.py
```diff
@@ -23,6 +23,9 @@
 from stripe import _util
 
 from stripe._stripe_response import StripeResponse, StripeStreamResponse
+import warnings
+
+_util.deprecated
```
pakrym-stripe, line 28: nit: rem?

stripe/_list_object.py
```diff
@@ -50,6 +71,9 @@ def list(
             ),
         )
 
+    @_util.deprecated(
+        "This will be removed in a future version of stripe-python. Please call the `create` method on the corresponding resource directly, instead of using `create` from the list object."
+    )
     def create(
```
pakrym-stripe, line 77: I didn't realize we are removing an entire pattern. Hm.

pakrym-stripe (approved):

stripe/_error.py
richardm-stripe: Oh, that [caused tests to fail](https://github.com/stripe/stripe-python/actions/runs/7174803441/job/19536938718?pr=1101#step:6:57) because *we* actually do subclass. So I went with "a private version that calls the public version but swallows the error". I think a user defining their own StripeError subclass is fairly unlikely but I feel more comfortable with the more conservative option.

stripe/_stripe_object.py
richardm-stripe: oh, good catch. This was for catching the circular import

richardm-stripe, line 28:
```suggestion
```

stripe/_list_object.py
richardm-stripe: Yes I think we should discourage their use because
1. They aren't guaranteed to point to API methods that actually exist.
2. We can't give them specific types.

stripe/_stripe_object.py
richardm-stripe: Line 463

pakrym-stripe: I'm blind.

PR #1102, opened 2023-10-26
Docstrings on fields

richardm-stripe:
Add docstrings to resource properties / method params.


setup.py
```diff
@@ -28,7 +28,7 @@
     license="MIT",
     keywords="stripe api payments",
     packages=find_packages(exclude=["tests", "tests.*"]),
-    package_data={"stripe": ["data/ca-certificates.crt"]},
+    package_data={"stripe": ["data/ca-certificates.crt", "py.typed"]},
```
pakrym-stripe, line 31: Accident?

pakrym-stripe (changes requested):
Lint is sad, strings need escaping.

setup.py
richardm-stripe: Fixed!

pakrym-stripe (approved):

PR #1104, opened 2023-10-26
Include `py.typed` and enable type annotations for the package

richardm-stripe:
## Summary
This PR changes `setup.py` so that inline type annotations are enabled for the `stripe-python` package. Up until now, the type annotations were just hanging out, existing, but would not be analyzed by type checkers without special effort. Now, anybody who does `pip install stripe` will experience the inline types. (subject to the [resolution order](https://peps.python.org/pep-0561/#type-checker-module-resolution-order)).

See more at the [wiki page](https://github.com/stripe/stripe-python/wiki/Inline-type-annotations).


pakrym-stripe (approved):
Can you please add a description for this one?

PR #1117, opened 2023-10-31
Make: move editable install into venv task

richardm-stripe:
This was previously done only in `make pyright` but you should be able to clone the repo, call `Make` and your `venv` should be in a correct state so that if you point your IDE to use `venv/bin/python` type checking and everything will work.


pakrym-stripe (approved):

richardm-stripe:
closing, we should try and avoid editable installs if possible

PR #1125, opened 2023-11-09
Add support for Python 3.11/3.12

Jamim:
Hello,

I believe it's worth having an explicit support and tests for Python 3.11/3.12, so I'd like to propose changes that:
  * enable testing for Python 3.11/3.12
  * update `classifiers` at `setup.py`
  * remove `pyright` from `tox.ini` since there's no corresponding `testenv` configuration
  * ~~bump `black` to `23.11.0` to get support for `py311`/`py312` targets~~
    [it looks like](https://github.com/Jamim/stripe-python/actions/runs/6817083959/job/18539818829#step:5:317) it's better to keep current version of `black` for now

Best regards!


pakrym-stripe:
Thank you for the contribution, @Jamim !

pakrym-stripe (approved):

PR #1126, opened 2023-11-10
Suppress type errors from latest pyright

richardm-stripe:
See commit-by-commit.

For [Fix redundant keyword params](https://github.com/stripe/stripe-python/commit/706b0b8d900809a80f1c5299c1ce7815e346cb59) -- I checked that this was safe by looking into the implementation of `_static_request` or `_request` in each case and observing that `util.read_special_variable` was called (eventually) to grab the value from `params.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 cryptography<40
 tox
 twine
-pyright <= 1.1.334
+pyright
 pytest-cov >= 2.8.1, < 2.11.0
 pytest-mock >= 2.0.0
 pytest-xdist >= 1.31.0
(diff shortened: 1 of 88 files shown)


pakrym-stripe:
Do we have some tests that exercise these parameters?

pakrym-stripe (changes requested):
Tests are red

richardm-stripe:
Turns out this is actually not safe. We don't use the little '*' trick to prevent keyword args from being passed positionally, so we have to suppress the type errors until we can do a major.

pakrym-stripe (approved):

PR #1129, opened 2023-11-10
Remove unneeded `if TYPECHECKING` guard

richardm-stripe:


stripe/api_resources/account_session.py
```diff
@@ -99,9 +91,7 @@ def create(
         idempotency_key: Optional[str] = None,
         stripe_version: Optional[str] = None,
         stripe_account: Optional[str] = None,
-        **params: Unpack[
```
pakrym-stripe, line 102: regression

PR #1134, opened 2023-11-15
Run pyright via tox

pakrym-stripe:
Updates our infrastructure to always run tools on python3.10 and only run tests on different versions of python.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   lint:
-    name: Lint
+    name: Typecheck and lint
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -27,6 +27,8 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: "3.10"
+      - name: Typecheck with pyright
+        run: make pyright
       - name: lint
         run: make lint
       - name: fmtcheck
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,20 +68,29 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version:
-          [
-            "3.6",
-            "3.7",
-            "3.8",
-            "3.9",
-            "3.10",
-            "3.11",
-            "3.12",
-            "pypy-3.7",
-            "pypy-3.8",
-            "pypy-3.9",
-            "pypy-3.10",
-          ]
+        include:
+            - python-version: "3.6"
+              env: py36
+            - python-version: "3.7"
+              env: py37
+            - python-version: "3.8"
+              env: py38
+            - python-version: "3.9"
+              env: py39
+            - python-version: "3.10"
+              env: py310
+            - python-version: "3.11"
+              env: py311
+            - python-version: "3.12"
+              env: py312
+            - python-version: "pypy-3.7"
+              env: py3.7
+            - python-version: "pypy-3.8"
+              env: py3.8
+            - python-version: "pypy-3.9"
+              env: py3.9
+            - python-version: "pypy-3.10"
+              env: py3.10
 
     steps:
       - uses: actions/checkout@v3
@@ -106,7 +115,7 @@ jobs:
         if: ${{ !contains(matrix.python-version, 'pypy') }}
 
       - name: Test with pytest
-        run: TOX_ARGS="-e ${{ matrix.python-version }}" make ci-test
+        run: TOX_ARGS="-e ${{ matrix.env }}" make ci-test
 
       - name: Calculate and publish coverage
         run: make coveralls
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,17 +95,12 @@ jobs:
     steps:
       - uses: actions/checkout@v3
 
-      - name: Set up Python ${{ matrix.python-version }}
+      - name: Set up Python ${{ matrix.python-version }} and 3.10
         uses: actions/setup-python@v4
         with:
-          python-version: ${{ matrix.python-version }}
-          cache: "pip"
-          cache-dependency-path: "setup.py"
-
-      - name: Set up Python 3.10
-        uses: actions/setup-python@v4
-        with:
-          python-version: "3.10"
+          python-version: |
+            ${{ matrix.python-version }}
+            3.10
 
       - uses: stripe/openapi/actions/stripe-mock@master
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,7 +60,6 @@ jobs:
           path: dist/
 
   test:
-    name: Test ${{ python-version }}
     # Specific ubuntu version to support python 3.6 testing
     # see https://github.com/actions/setup-python/issues/544#issuecomment-1332535877 for details
     # move to ubuntu-latest when we drop 3.6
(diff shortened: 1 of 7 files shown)


richardm-stripe (approved):

.github/workflows/ci.yml
```diff
@@ -66,6 +66,7 @@ jobs:
     # move to ubuntu-latest when we drop 3.6
     runs-on: ubuntu-20.04
     strategy:
+      fail-fast: false
```
pakrym-stripe, line 69: Allow all legs of tests to run

.github/workflows/ci.yml
```diff
@@ -92,23 +93,15 @@ jobs:
           cache: "pip"
           cache-dependency-path: "setup.py"
 
-      - name: Upgrade pip and virtualenv to latest
-        run: pip install --upgrade pip virtualenv
+      - name: Set up Python 3.10
```
pakrym-stripe, line 96: Always setup python 10, it's our base version to run all the tooling on

.github/workflows/ci.yml
```diff
@@ -19,14 +19,16 @@ on:
...
-    name: Lint
+    name: Typecheck and lint
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - name: Set up Python 3
         uses: actions/setup-python@v4
         with:
           python-version: "3.10"
+      - name: Typecheck with pyright
```
pakrym-stripe, line 30: Typecheck on py10 only

.github/workflows/ci.yml
```diff
@@ -92,23 +95,15 @@ jobs:
...
-        run: make pyright
-        # Pip won't install newer versions of pyright on python 3.6 (odd, because pyright is a node module)
-        # so we skip running pyright, and do double duty on python 3.7 using the --pythonversion flag.
-        if: matrix.python-version != '3.6'
-
-      - name: Typecheck with pyright
-        run: make pyright PYRIGHT_ARGS="--pythonversion 3.6"
-        if: matrix.python-version == '3.7'
-
       - name: Test with pytest
```
pakrym-stripe, line 105: Pass an explicit pip environment for the version we want to test on.

test-requirements.txt
```diff
@@ -0,0 +1,6 @@
+pytest-cov >= 2.8.1, < 2.11.0
```
pakrym-stripe, line 1: These requirements must be installable on all our supported versions

requirements.txt
```diff
@@ -1,14 +1,10 @@
 # cryptography 40.0.0 deprecates support for Python 3.6 and PyPy3 < 7.3.10
 cryptography<40
-tox
 twine
```
pakrym-stripe, line 3: These requirements must be installable only on py3.10 +

tox.ini
```diff
@@ -7,8 +7,8 @@
 envlist =
     fmt
     lint
+    pyright
     py{312,311,310,39,38,37,36,py3}
-skip_missing_interpreters = true
```
pakrym-stripe, line 11: Do not rely on skipping missing interpreters, always run a particular version and fail if interpreter is not available.

.github/workflows/ci.yml
```diff
@@ -92,23 +93,20 @@ jobs:
           cache: "pip"
           cache-dependency-path: "setup.py"
 
-      - name: Upgrade pip and virtualenv to latest
-        run: pip install --upgrade pip virtualenv
```
anniel-stripe, line 96: Why don't we want to upgrade pip / virtualenv to latest anymore?

pakrym-stripe: We use pip from python 3.10 and use venv instead of virtualenv that's bundled with python as well.

pyproject.toml
```diff
@@ -23,14 +23,10 @@ exclude = '''
 )
 '''
 [tool.pyright]
-executionEnvironments=[
-  {"root" = "stripe"},
-  {"root" = "tests"}
-]
```
anniel-stripe, line 29: Why don't we need executionEnvironments anymore?

pakrym-stripe: @richardm-stripe  will know better, some fix in pyright made it non-required

Makefile
```diff
@@ -1,6 +1,6 @@
 VENV_NAME?=venv
 PIP?=pip
-PYTHON?=python3
+PYTHON?=python3.10
```
richardm-stripe, line 3: What's the thinking here? When we want to use a version other than 3.10 this means we need to `make pyright PYTHON=<some/other/python>` (or whatever the syntax is for providing variables to Make, I forget) instead of being able to use pyenv

pakrym-stripe: We do https://github.com/stripe/stripe-python/pull/1134/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR104

.github/workflows/ci.yml
```diff
@@ -92,23 +93,20 @@ jobs:
...
-      - name: Typecheck with pyright
-        run: make pyright PYRIGHT_ARGS="--pythonversion 3.6"
-        if: matrix.python-version == '3.7'
+        run: PYRIGHT_ARGS="-- --pythonversion ${{ matrix.python-version }}" make pyright
+        # Skip typecheking in pypy legs
+        if: ${{ !contains(matrix.python-version, 'pypy') }}
 
       - name: Test with pytest
-        run: make ci-test
+        run: TOX_ARGS="-e ${{ matrix.python-version }}" make ci-test
```
richardm-stripe, line 109: Looking at some of the cI output:

https://github.com/stripe/stripe-python/actions/runs/6882099159/job/18719881529?pr=1134#step:7:28

I'm not actually convinced that `pytest` is being run with different python interpreters -- it looks like 3.10 is running the test suite always? Or at least, there are a lot of mentions of 3.10

image


Can you double check this and/or possible cause pytest to output something that unambiguously indicates what python version it is running under?

Makefile
richardm-stripe: I was thinking about the local dev environment and just pointing out that this will supersede pyenv.

pakrym-stripe: It won't supersede pyenv AFAIK, it will always create pyenv with python3.10 instead of whatever the current version of python in environment is

pakrym-stripe: But yeah, it's harder to create pyenv using a different version now

richardm-stripe: I think you are hearing `venv` when I am saying `pyenv`?

I guess part of my workflow when I want to locally run pyright or pytest on a particular python version is
```
$ echo "3.x" > .python-version
$ make 
$ . venv/bin/activate
$ pyright
$ python -m pytest tests/
```
and now `.python-version` will be ignored, unless I override the explicit hardcoding of 3.10
```
$ echo "3.x" > .python-version
$ make PYTHON=python
$ . venv/bin/activate
$ pyright
$ python -m pytest tests/
```

pakrym-stripe: >I think you are hearing venv when I am saying pyenv?
I do :/  Sorry.

How often do we use the ability to switch python versions?

tests/conftest.py
```diff
@@ -1,19 +1,14 @@
 import atexit
 import os
 import sys
-from distutils.version import StrictVersion
```
pakrym-stripe, line 4: distutils were [removed in 3.12.](https://github.com/nodejs/node-gyp/issues/2869). I'm not sure how tests ever passed there

.github/workflows/ci.yml
pakrym-stripe: Fixed

richardm-stripe (approved):

PR #1142, opened 2023-11-17
Move resource type exports to stripe.___

pakrym-stripe:
Deprecated modules are highlighted in pycharm and cause type errors in MyPy and Pyright, the warning is logged when importing deprecated module.
image
image

```
> python -c "import stripe.api_resources.card"
<string>:1: DeprecationWarning:
    The stripe.api_resources package is deprecated, please change your
    imports to import from stripe directly.
    From:
      from stripe.api_resources import ...
    To:
      from stripe import ...
```

## Changelog
- `stripe.error`, `stripe.stripe_object`, `stripe.api_requestor`, `stripe.stripe_response`, `stripe.request_options`, `stripe.api_resources.*`,  `stripe.api_resources.abstract.*` modules are deprecated. All types are available directly from `stripe` module now.
   Before:
   ```python
   from stripe.error import APIError
   # or
   stripe.error.APIError
   ````
   After:
   ```python
   from stripe import APIError
   # or
   stripe.APIError
   ```


stripe/_account.py
```diff
@@ -1,17 +1,16 @@
 # -*- coding: utf-8 -*-
```
anniel-stripe, line 1: So I know we talked about potentially putting all these resources into an `_api` subpackage (potentially cleaner / easier to navigate). Just to confirm, the reason not to have an `_api` subpackage is because we need namespace packages like `tax` to be on the top-level , so that `from stripe.tax import ...` will work?

pakrym-stripe: Correct.  But I also quite like how exports and the directory structure align it's easier to see type conflicts and in general think about public API this way.

richardm-stripe (approved):
Did the best I could spot-checking this and left some small questions. The approach in general looks good, and the import tests and the fact that the existing test suite passes gives me a lot of confidence.

stripe/api_resources/abstract/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+from warnings import warn
+
+warn(
+    """
+    The stripe.api_resources.abstract package is deprecated, please change your
+    imports to import from stripe.abstract directly.
+    From:
+      from stripe.api_resources.abstract import ...
+    To:
+      from stripe.abstract import ...
```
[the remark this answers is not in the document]
pakrym-stripe: The message is wrong.

flake8_stripe/flake8_stripe.py
```diff
@@ -114,3 +115,51 @@ def run(self) -> Iterator[Tuple[int, int, str, type]]:
...
+        "stripe.app_info",
+        "stripe.api_version",
+        "stripe.http_client",
+    ]
+
+    def run(self) -> Iterator[Tuple[int, int, str, type]]:
+        for node in ast.walk(self.tree):
+            if isinstance(node, ast.Import):
+                if any(alias.name == "stripe" for alias in node.names):
+                    msg = "IMP101 Do not `import stripe` unless you are accessing the global variable or breaking circular dependency"
```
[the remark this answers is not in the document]
pakrym-stripe: > Does this match just literally `import stripe`

Yep.

> What exactly is alias and why do we need to do any?

`node.names` are not strings instead they are pairs of `asname` and `alias` (`import {alias} as {asname}`)

stripe/__init__.py
```diff
@@ -18,6 +18,7 @@
 if TYPE_CHECKING:
     from stripe.http_client import HTTPClient
 
+_this_module = _sys.modules[__name__]
```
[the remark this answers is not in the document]
pakrym-stripe: Nothing, left over from experiments with exports.

stripe/api_resources/abstract/__init__.py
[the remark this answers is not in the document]
pakrym-stripe: Fixed

stripe/__init__.py
```diff
@@ -70,3 +63,226 @@ def set_app_info(
...
+from stripe._error import InvalidRequestError as InvalidRequestError
+from stripe._error import AuthenticationError as AuthenticationError
+from stripe._error import PermissionError as PermissionError
+from stripe._error import RateLimitError as RateLimitError
+from stripe._error import (
+    SignatureVerificationError as SignatureVerificationError,
+)
+
+# Backwards compatibility re-exports
+# TODO (major): remove in https://jira.corp.stripe.com/browse/DEVSDK-1706
```
anniel-stripe, line 134: nit: Should we have links to our private Jira in public repo?

stripe/_api_resource.py
```diff
@@ -190,3 +189,6 @@ def _static_request_stream(
 
         response, _ = requestor.request_stream(method_, url_, params, headers)
         return response
+
+
+from stripe import _api_requestor, _error, util  # noqa
```
anniel-stripe, line 194: What's the purpose of importing at the bottom of the file like this?

stripe/_encode.py
```diff
@@ -0,0 +1,45 @@
+import calendar
+import datetime
+import time
+from collections import OrderedDict
+from typing import Generator, Tuple, Any
+
+
+def _api_encode(data) -> Generator[Tuple[str, Any], None, None]:
```
anniel-stripe, line 8: Any reason for moving these functions out of _api_requestor.py (I'm a fan for cleanliness though)? Would this be breaking if a user were trying to do `from stripe.api_requestor import _encode_datetime` or something?

stripe/api_resources/recipient_transfer.py
```diff
@@ -1,6 +0,0 @@
-from stripe.stripe_object import StripeObject
-
-
-# This resource can only be instantiated when expanded on a BalanceTransaction
-class RecipientTransfer(StripeObject):
```
anniel-stripe, line 5: Where did this file go? I can't seem to find it if it was renamed

stripe/_encode.py
pakrym-stripe: These helpers don't have any dependencies and loading the entire `stripe.api_requestor ` module just to get the helper caused circular references.

All names that start with underscore are considered private so it feels like a fair game to move these helpers.

stripe/api_resources/recipient_transfer.py
pakrym-stripe: Nice catch, this class is unused and was lost in all the moves.

stripe/stripe_object.py
```diff
@@ -1,507 +1,21 @@
...
+# -*- coding: utf-8 -*-
+from typing_extensions import TYPE_CHECKING
+from warnings import warn
+
+warn(
+    """
+    The stripe_stripe_object package is deprecated, please change your
+    imports to import from stripe directly.
+    From:
+      from stripe_stripe_object import StripeObject
```
anniel-stripe, line 10: nit
```suggestion
      from stripe._stripe_object import StripeObject
```

stripe/_api_resource.py
pakrym-stripe: One of my attempts at resolving circular references, not needed anymore.

anniel-stripe (approved):
Aside from the nits and bringing back `recipient_transfer.py`, this looks great to me! Love the deprecation warnings / tests / linter rules. And thanks for breaking it up into easy-to-review commits ❤️

stripe/_api_resource.py
pakrym-stripe: Fixed

stripe/__init__.py
```diff
@@ -70,3 +62,226 @@ def set_app_info(
...
+
+        def __getattr__(name):
+            if name == "abstract":
+                import stripe.api_resources.abstract as _abstract
+
+                return _abstract
+            if name == "api_resources":
+                import stripe.api_resources as _api_resources
+
+                return _api_resources
```
andersk, line 157: This needs to `raise AttributeError` rather than falling off the end and returning `None`.

- #1158
- #1159

silviogutierrez:
Docs should probably be updated: https://docs.stripe.com/error-handling?lang=python

Currently uses `.error` and mypy / type checkers complain.

ramya-stripe:
Thanks for reporting that @silviogutierrez 
We will look into updating the docs

PR #1145, opened 2023-11-21
Refactor integration test

richardm-stripe:
[tests/test_integration.py](https://github.com/stripe/stripe-python/compare/richardm-refactor-integration-test?expand=1#diff-98bfe78f0e8d895dd454b5eef6f6485a80526776ad3b6525a14117029bd1eacb) is a test file that spins up an HTTP server running in a separate thread and is useful for tests that need to be written at the HTTP level, but I ran into some difficulties when attempting to add a new test here. This PR refactors the test helper a bit to make it more usable:
  - Instead of producing assertion errors inside the HTTP Server thread and doing a weird conversion into HTTP 400s, instead the HTTP server thread sends the request objects back to the test thread via a `Queue` and you can assert on the received requests inside the test.
  - There is a sensible default implementation that returns 200 with empty JSON.
  - Now, to define a test HTTP server, you just subclass `TestServer` and (optionally) implement `do_request(n: int)` to return [status_code, headers, body]`


tests/test_integration.py
```diff
@@ -116,19 +149,8 @@ def do_GET(self):
         assert MockServerRequestHandler.num_requests == 2
 
     def test_hits_proxy_through_custom_client(self):
```
pakrym-stripe, line 151: This test doesn't seem to test what it claims

tests/test_integration.py
```diff
@@ -17,6 +20,57 @@
     from http.server import BaseHTTPRequestHandler, HTTPServer
 
 
+class TestHandler(BaseHTTPRequestHandler):
+    num_requests = 0
+
+    requests = defaultdict(Queue)
+
+    @classmethod
+    def _add_request(cls, req):
```
pakrym-stripe, line 29: Is there any way to make this non-static?

pakrym-stripe: Possibly by returning something from self.setup_mock_server?

tests/test_integration.py
```diff
@@ -17,6 +20,57 @@
...
+    def do_GET(self):
+        return self._do_request()
+
+    def do_POST(self):
+        return self._do_request()
+
+    def _do_request(self):
+        n = self.__class__.num_requests
+        self.__class__.num_requests += 1
+        self._add_request(self)
```
pakrym-stripe, line 51: Is the handler created for each request?

tests/test_integration.py
```diff
@@ -141,101 +163,71 @@ def do_GET(self):
...
 
         stripe.Balance.retrieve()
         stripe.Balance.retrieve()
+
+        reqs = MockServerRequestHandler.get_requests(2)
         assert MockServerRequestHandler.num_requests == 2
+        # req 1
+        assert not reqs[0].headers.get("X-Stripe-Client-Telemetry")
+        # req 2
+        telemetry_raw = reqs[1].headers.get("x-stripe-client-telemetry")
```
pakrym-stripe, line 194: nit: use the same header name casing?

pakrym-stripe (approved):
I'm not sure why these tests need an actual HTTP server and can't be done on top of HTTP client mock.

tests/test_integration.py
richardm-stripe: I think this needs to stay a class method. Things are weird because the python standard library is weird: particularly `HTTPServer` takes in `handler` as a *class* instead of an instance, and so we don't actually control making instances of `TestServer`, and an *instance* of BaseHTTPRequestHandler actually represents a single incoming request. So the test case needs its own `.get_requests`, and `_add_request` scoped to the *subclass* that it defines, not to an instance.

tests/test_integration.py
richardm-stripe: Yes. An instance of BaseHTTPRequestHandler actually represents a single incoming request.

tests/test_integration.py
```diff
@@ -141,101 +163,71 @@ def do_GET(self):
...
         stripe.api_base = "http://localhost:%s" % self.mock_server_port
         stripe.enable_telemetry = True
 
         stripe.Balance.retrieve()
         stripe.Balance.retrieve()
+
+        reqs = MockServerRequestHandler.get_requests(2)
         assert MockServerRequestHandler.num_requests == 2
+        # req 1
+        assert not reqs[0].headers.get("X-Stripe-Client-Telemetry")
```
richardm-stripe, line 192:
```suggestion
        assert not reqs[0].headers.get("x-stripe-client-telemetry")
```

richardm-stripe:
> I'm not sure why these tests need an actual HTTP server and can't be done on top of HTTP client mock.

Some of the logic being tested here is logic that is implemented in `http_client.py` itself.

tests/test_integration.py
richardm-stripe:
```
        stripe.default_http_client = (
            stripe.http_client.new_default_http_client(
                proxy="http://localhost:%s" % self.mock_server_port
            )
        )
 ```
 is creating a new client with a proxy? In any case, I just want to refactor the file so that adding new tests is possible. Reimagining existing tests can stay out of scope.

PR #1146, opened 2023-11-21
Track usage of deprecated `save`

richardm-stripe:
Note: not to be merged until after https://github.com/stripe/stripe-python/pull/1142 so as to minimize merge conflicts. 
## Changelog
* Reports uses of the deprecated `.save` in `X-Stripe-Client-Telemetry`. (You can disable telemetry via `stripe.enable_telemetry = false`, see the [README](https://github.com/stripe/stripe-python/blob/master/README.md#telemetry).)

## Details
The unsatisfying thing about the implementation is how the telemetry header is added at the `http_client` layer of abstraction. I would have liked to change this and moved it to `api_requestor` -- but we are planning to fundamentally upheave api_requestor in the near future anyway, so I chose not to.

I added `_usage` as an optional keyword param to `ApiRequestor.request` and `HttpClient.request_with_retries` (along with streaming analogs, and to internal methods).

## Testing
I modified the existing test case for telemetry (in integration_test.py) to use a method that would trigger "save" usage to be included.

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -164,8 +164,6 @@ class TestIntegration(object):
 
     def test_passes_client_telemetry_when_enabled(self):
         class MockServerRequestHandler(TestHandler):
-            num_requests = 0
-
             def do_request(self, req_num):
                 if req_num == 0:
                     time.sleep(31 / 1000)  # 31 ms
@@ -174,7 +172,7 @@ class TestIntegration(object):
                     200,
                     {
                         "Content-Type": "application/json; charset=utf-8",
-                        "Request-Id": "req_1",
+                        "Request-Id": "req_%s" % (req_num + 1),
                     },
                     None,
                 ]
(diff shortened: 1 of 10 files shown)


stripe/http_client.py
```diff
@@ -134,21 +134,33 @@ def __init__(
...
+        _usage: Optional[List[str]] = None
     ) -> Tuple[Any, int, Any]:
         return self._request_with_retries_internal(
-            method, url, headers, post_data, is_streaming=True
+            method, url, headers, post_data, is_streaming=True, _usage=_usage
         )
 
     def _request_with_retries_internal(
-        self, method, url, headers, post_data, is_streaming
+        self, method, url, headers, post_data, is_streaming, *, _usage=None
```
anniel-stripe, line 163: We're adding `_usage` as a keyword-only parameter in some places and positional in others - is there a reason for which places have which? Or should we unify on one?

tests/test_integration.py
```diff
@@ -183,11 +181,15 @@ def do_request(self, req_num):
...
-        stripe.Balance.retrieve()
-        stripe.Balance.retrieve()
+        cus = stripe.Customer("cus_xyz")
+        cus.description = "hello"
+        cus.save()
+
+        stripe.Customer.retrieve("cus_xyz")
+        stripe.Customer.retrieve("cus_xyz")
+
+        reqs = MockServerRequestHandler.get_requests(3)
```
anniel-stripe, line 192: Bring back this assertion?
```suggestion
        assert MockServerRequestHandler.num_requests == 3
        
```

anniel-stripe (approved):
Left a couple comments but nothing blocking 👍

stripe/http_client.py
richardm-stripe: The only place it's positional-only I think is `_record_request_metrics` which is private and I think it should be required there.

Everywhere else's, it's a keyword argument and in some places I added the `*` but I kind of wanted to get the `*` thing figured out everywhere and I prepared https://github.com/stripe/stripe-python/pull/1144

tests/test_integration.py
richardm-stripe: It's actually covered by `reqs = MockServerRequestHandler.get_requests(3)` -- this is cleanup I forgot to do when I introduced the test server here: https://github.com/stripe/stripe-python/pull/1145/files#diff-98bfe78f0e8d895dd454b5eef6f6485a80526776ad3b6525a14117029bd1eacbR39

PR #1153, opened 2023-12-07
Move exports for more modules

pakrym-stripe:


tests/test_http_client.py
```diff
@@ -11,10 +11,10 @@
 
 class StripeClientTestCase(object):
     REQUEST_LIBRARIES = [
-        ("urlfetch", "stripe.http_client.urlfetch"),
-        ("requests", "stripe.http_client.requests"),
-        ("pycurl", "stripe.http_client.pycurl"),
-        ("urllib.request", "stripe.http_client.urllibrequest"),
+        ("urlfetch", "stripe._http_client.urlfetch"),
```
pakrym-stripe, line 14: As this and util tests show the change might be breaking if users were patching the sdk internals.

stripe/__init__.py
```diff
@@ -11,12 +11,13 @@
...
-from stripe.api_version import _ApiVersion
+from stripe._api_version import _ApiVersion
 
-from stripe.app_info import AppInfo
-
-if TYPE_CHECKING:
-    from stripe.http_client import HTTPClient
+# We must import the app_info module eagerly before defining the app_info global
+# otherwise the late import will overwrite the global
+import stripe.app_info
```
anniel-stripe, line 18: Wouldn't this overwriting behavior have been true before?

pakrym-stripe: No, because we were importing from app_info before so the app_info modules was already loaded before `app_info = None` line runs.

anniel-stripe (approved):
LGTM!

> oauth_errors is the only module that I kept as is. Types from it conflict with standard error types.
Do we have a plan to make this module private at some point in the future?

PR #1157, opened 2023-12-07
Add missing explicit reexport for `OAuth`, `Webhook`, `WebhookSignature`

andersk:
This is required for mypy to recognize these as exported with [`--no-implicit-reexport`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-no-implicit-reexport) (implied by `--strict`).


pakrym-stripe (approved):

pakrym-stripe:
Thank you @andersk !

PR #1159, opened 2023-12-07
Fix __getattr__ to raise AttributeError rather than returning None

### LINKED ISSUE #1158: [7.8.0 regression] `stripe.checkout` is `None` after `import stripe`
### Describe the bug

Previously, `import stripe` would cause `stripe.checkout` to be available. But after #1142 in 7.8.0, `stripe.checkout` is defined as `None`.

### To Reproduce

```pycon
>>> import stripe
>>> stripe.checkout.Session
AttributeError: 'NoneType' object has no attribute 'Session'
>>> print(stripe.checkout)
None
```

### Expected behavior

For compatibility, `stripe.checkout` should be available after `import stripe`, like it is after `import stripe.checkout`.

```pycon
>>> import stripe.checkout
>>> print(stripe.checkout)
<module 'stripe.checkout' from '/srv/zulip-py3-venv/lib/python3.8/site-packages/stripe/checkout/__init__.py'>
```

### Code snippets

_No response_

### OS

Linux

### Language version

Python 3.8.10

### Library version

stripe-python v7.8.0

### API version

N/A

### Additional context

_No response_

andersk:
- Fixes #1158.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -174,6 +174,7 @@ if not TYPE_CHECKING:
                 import stripe.api_resources as _api_resources
 
                 return _api_resources
+            raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
 
     else:
         with warnings.catch_warnings():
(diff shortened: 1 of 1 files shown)


pakrym-stripe:
Thank you, @andersk !

PR #1160, opened 2023-12-07
Correctly re-export nested packages

pakrym-stripe:
Fixes https://github.com/stripe/stripe-python/issues/1158

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -186,22 +186,20 @@ if not TYPE_CHECKING:
 # API resources
 
 # The beginning of the section generated from our OpenAPI spec
-from stripe import (
-    apps as apps,
-    billing_portal as billing_portal,
-    checkout as checkout,
-    climate as climate,
-    financial_connections as financial_connections,
-    identity as identity,
-    issuing as issuing,
-    radar as radar,
-    reporting as reporting,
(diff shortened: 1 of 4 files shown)


stripe/__init__.py
```diff
@@ -186,22 +186,20 @@ def __getattr__(name):
...
-    issuing as issuing,
-    radar as radar,
-    reporting as reporting,
-    sigma as sigma,
-    tax as tax,
-    terminal as terminal,
-    test_helpers as test_helpers,
-    treasury as treasury,
-)
+import stripe.apps as apps
```
anniel-stripe, line 189:
```suggestion
from . import apps as apps
```
?

andersk, line 189: This won’t work; mypy doesn’t treat this as an explicit reexport like it does `from stripe import apps as apps`.

The only reason `from stripe import apps as apps` appeared not to be working was that a bug in the module-level `__getattr__` was causing module-level variables of these names (indeed, every name) to be defined as `None` rather than undefined:

- #1159

pakrym-stripe: Yep, you are right. Thank you @andersk, we'll merge your change first and this PR after.

anniel-stripe (approved):

PR #1165, opened 2023-12-08
Beta: raw_request_async with HTTPX

richardm-stripe:
## Summary
Adds async infrastructure and `stripe.raw_request_async` to the beta channel.

## Details

Additions:

* `stripe.raw_request_async` - Gives a small surface area so that we can test async "end to end". We will codegen async methods for each API method in a follow-up.
* `stripe.HTTPClientAsync` - Defines the "interface" for defining an async http client.
* `stripe.HTTPXClient` - Default implementation of `stripe.HTTPClientAsync`. I picked httpx somewhat arbitrarily. It happens to be what [openai](https://github.com/openai/openai-python/tree/main/src/openai) used (a few months ago, openai-python was a fork of stripe-python with added async support, but this is no longer the case). We can add more or different defaults.
* `stripe.default_http_client_async` - A place for a user to set a custom async client
  (either a custom `stripe.HTTPXClient` or they can implement `HTTPClientAsync`.
* `ApiRequestor.request_async`, `ApiRequestor.request_stream_async`, `ApiRequestor.request_raw_async` - Infrastructure methods that lie between `raw_request` and `http_client`.

Notes: 

  - I omitted the ability to pass `http_client_async` into an ApiRequestor. You have to use the global http client. Mostly this is in the name of minimalism. Shouldn't be needed after StripeClient lands.
  - The `stripe.new_default_http_client_async` function exists but it is not exported publicly. It doesn't make too much sense to add because there's only one baked-in HTTPClientAsync: HTTPX, so you might as well just construct a `stripe.HTTPXClient`.
  - Since there are two http clients, users who want to use both sync and async interfaces might need to specify options such as `verify_ssl_certs` and `proxy` twice. In theory we could try and unify the interface for setting these settings on the global client, but I don't think this is an important enough use case to justify the additional complication.


flake8_stripe/flake8_stripe.py
```diff
@@ -31,6 +31,9 @@ class TypingImportsChecker:
         "NotRequired",
         "Self",
         "Unpack",
+        "Awaitable",
+        "Never",
+        "Awaitable",
```
akx, line 36: Duplicate Awaitable.

stripe/_api_requestor.py
```diff
@@ -78,6 +79,16 @@ def __init__(
...
 
+        if stripe.default_http_client_async:
+            self._client_async = stripe.default_http_client_async
+        else:
+            stripe.default_http_client_async = (
+                _http_client.new_default_http_client_async(
+                    verify_ssl_certs=verify, proxy=proxy
+                )
+            )
+            self._client_async = stripe.default_http_client_async
```
akx, line 90: Wouldn't 
```suggestion
        if not stripe.default_http_client_async:
            stripe.default_http_client_async = (
                _http_client.new_default_http_client_async(
                    verify_ssl_certs=verify, proxy=proxy
                )
            )
        self._client_async = stripe.default_http_client_async
```
be easier to read?

richardm-stripe: I do like that better!

stripe/_api_requestor.py
```diff
@@ -434,6 +507,52 @@ def request_raw(
...
+        else:
+            (
+                rcontent,
+                rcode,
+                rheaders,
+            ) = await self._client_async.request_with_retries(
+                method, abs_url, headers, post_data, _usage=_usage
+            )
+
+        _util.log_info(
```
pakrym-stripe, line 542: can this me shared?

richardm-stripe: [share logging](https://github.com/stripe/stripe-python/pull/1165/commits/ced75f85c5e0b54449eea8cf74958b52be1e82df)

stripe/_http_client.py
```diff
@@ -334,6 +357,106 @@ def close(self):
...
+                    self._record_request_metrics(
+                        response, request_start, usage
+                    )
+
+                    return response
+                else:
+                    assert connection_error is not None
+                    raise connection_error
+
+    async def request(self, method, url, headers, post_data=None):
```
pakrym-stripe, line 444: Should we suffix these method with async? In case there is a single HTTP client that can implement both?

pakrym-stripe: And for general consistency

richardm-stripe: Good idea!

stripe/_http_client.py
```diff
@@ -804,3 +927,76 @@ def _handle_request_error(self, e) -> NoReturn:
...
+class HTTPXClient(HTTPClientAsync):
+    def __init__(self, **kwargs):
+        super(HTTPXClient, self).__init__(**kwargs)
+
+        assert httpx is not None
+        self.httpx = httpx
+        self._client = httpx.AsyncClient()
+
+    def sleep(self, secs):
+        import asyncio
```
pakrym-stripe, line 952: Why import here?

stripe/_http_client.py
```diff
@@ -804,3 +927,76 @@ def _handle_request_error(self, e) -> NoReturn:
...
+            )
+        except self.httpx.HTTPError as e:
+            self._handle_request_error(e)
+
+        content = response.content
+        status_code = response.status_code
+        response_headers = response.headers
+        return content, status_code, response_headers
+
+    def _handle_request_error(self, e) -> NoReturn:
```
pakrym-stripe, line 971: Can this be shared?

stripe/_http_client.py
```diff
@@ -804,3 +927,76 @@ def _handle_request_error(self, e) -> NoReturn:
...
+
+
+class NoImportFoundAsyncClient(HTTPClientAsync):
+    def __init__(self, **kwargs):
+        super(NoImportFoundAsyncClient, self).__init__(**kwargs)
+
+    async def request(
+        self, method, url, headers, post_data=None
+    ) -> Tuple[bytes, int, Mapping[str, str]]:
+        raise_async_client_import_error()
```
pakrym-stripe, line 996: inline the method into this class?

test-requirements.txt
```diff
@@ -1,4 +1,10 @@
 # These requirements must be installable on all our supported versions
+
+# This is the last version of httpx compatible with Python 3.6
```
pakrym-stripe, line 3: do we need to add these to setup.py?

tests/test_http_client.py
```diff
@@ -51,6 +52,25 @@ def test_new_default_http_client_urllib2(self, request_mocks):
...
+            setattr(_http_client, lib, None)
+
+        inst = _http_client.new_default_http_client_async()
+
+        assert isinstance(inst, expected)
+
+    def test_new_default_http_client_httpx(self, request_mocks):
+        self.check_default((), _http_client.HTTPXClient)
+
+    def test_new_default_http_client_no_import_found(self, request_mocks):
```
pakrym-stripe, line 67: Does this check other behaviors like timeouts and retries?

stripe/__init__.py
```diff
@@ -48,6 +49,7 @@
     WebhookSignature as WebhookSignature,
 )
 from stripe._raw_request import raw_request as raw_request  # noqa
+from stripe._raw_request import raw_request_async as raw_request_async  # noqa
 from stripe._raw_request import deserialize as deserialize  # noqa
 
 from stripe._preview import preview as preview  # noqa
```
pakrym-stripe, line 55: we'll need preview overloads as well

stripe/_api_requestor.py
```diff
@@ -407,6 +463,37 @@ def request_raw(
             api_version=self.api_version,
         )
 
+        return abs_url, headers, post_data, my_api_key
+
+    @staticmethod
+    def log_response(abs_url, rcode, rcontent, rheaders):
```
pakrym-stripe, line 469:
```suggestion
    def _log_response(abs_url, rcode, rcontent, rheaders):
```

?

stripe/_http_client.py
```diff
@@ -264,11 +289,11 @@ def request_stream_with_retries(
         _usage: Optional[List[str]] = None
     ) -> Tuple[Any, int, Any]:
         return self._request_with_retries_internal(
-            method, url, headers, post_data, is_streaming=True, _usage=_usage
+            method, url, headers, post_data, is_streaming=True, usage=_usage
         )
 
     def _request_with_retries_internal(
-        self, method, url, headers, post_data, is_streaming, *, _usage=None
+        self, method, url, headers, post_data, is_streaming, usage
```
pakrym-stripe, line 296: nit: why the inconsistent `_usage` naming?

stripe/_http_client.py
```diff
@@ -334,6 +357,106 @@ def close(self):
...
+        raise NotImplementedError(
+            "HTTPClientAsync subclasses must implement `request`"
+        )
+
+    async def request_stream_async(self, method, url, headers, post_data=None):
+        raise NotImplementedError(
+            "HTTPClientAsync subclasses must implement `request_stream`"
+        )
+
+    async def close(self):
```
pakrym-stripe, line 454: close_async?

stripe/_http_client.py
```diff
@@ -334,6 +357,106 @@ def close(self):
...
+        post_data=None,
+        *,
+        _usage: Optional[List[str]] = None
+    ) -> Tuple[Any, int, Any]:
+        return await self._request_with_retries_internal_async(
+            method, url, headers, post_data, is_streaming=True, usage=_usage
+        )
+
+    @classmethod
+    async def sleep(cls: Type[Self], secs: float) -> Awaitable[None]:
```
pakrym-stripe, line 388: sleep_async

stripe/_http_client.py
pakrym-stripe: Is there a linter for this?

stripe/_http_client.py
richardm-stripe: It shouldn't be. `e` is a library-specific error object and the job of this function is to convert library-specific exceptions and standardize their handling. So far this function is pretty generic, but it's good for the extension point to be exposed.

stripe/_http_client.py
pakrym-stripe: oh, I didn't realize we used `_` to mark the parameter internal.

tests/test_http_client.py
richardm-stripe: No this is just a test for the behavior of the `new_default` function delegating to the correct constructors.

test-requirements.txt
richardm-stripe: I don't think so. This is like `requests` -- we don't *require* it but we use it if the user happens to have it installed.

stripe/_http_client.py
richardm-stripe: Fixed!

stripe/_http_client.py
richardm-stripe: All fixed!

stripe/_http_client.py
richardm-stripe: We'd have to write ourselves. The only thing I can find is https://pypi.org/project/flake8-async/

richardm-stripe: I can ticket

stripe/_http_client.py
pakrym-stripe: I see we do implement different messages based on e for other libraries

stripe/_http_client.py
richardm-stripe: Done

stripe/_http_client.py
richardm-stripe: Yes, I looked through https://www.python-httpx.org/exceptions/ to see if there were any exceptions that were clearly "this exception should never be retried" but I didn't see any obvious ones.

The conservative choice is to retry everything IMO. Choosing not to retry certain classes of exception is kind of an optimization.

richardm-stripe: Maybe UnsupportedProtocol but I don't think it's worth it.

stripe/__init__.py
richardm-stripe: Added in [preview async methods](https://github.com/stripe/stripe-python/pull/1165/commits/0c7f0224ff3857d7a587dd46461f8b24518bac5e)

tests/test_http_client.py
richardm-stripe: In https://github.com/stripe/stripe-python/pull/1165/commits/cd9c8bb3998c4eb5b0239a1461c9f7c9348b9005 I added more tests to mirror the ones elsewhere in `test_http_client`. This led me to add support for `proxies` and `verify` like Requests has, too.

requirements.txt
```diff
@@ -7,6 +7,6 @@ virtualenv<20.22.0
 pyright == 1.1.336
 black == 22.8.0
 flake8
-mypy == 1.7.0
+# mypy == 1.7.0
```
richardm-stripe, line 10:
```suggestion
mypy == 1.7.0
```

stripe/_http_client.py
```diff
@@ -316,14 +337,12 @@ def _request_with_retries_internal(
                     assert connection_error is not None
                     raise connection_error
 
-    def request(self, method, url, headers, post_data=None, *, _usage=None):
```
richardm-stripe, line 319: These were mistakenly added during a bad merge of master -> beta, (merge conflict in the beta branch between adding "usage" and async refactor).

richardm-stripe:
Ok, this should be ready again. The git history got messy, but since you last looked I

* [Added support for "proxy", "verify", and "timeout" to the HTTPX client](https://github.com/stripe/stripe-python/pull/1165/commits/88bdc7a3d49ac8d4c31f9082aa657492164ce3c1)
* [Addressed code review](https://github.com/stripe/stripe-python/pull/1165/commits/161d37dcd916b6a0d377d4b355462b1991d5202d)
* [Added preview overloads](https://github.com/stripe/stripe-python/pull/1165/commits/e5a27d601313caff274744daa3e8fc4ac0df2fba)
* [Added a lot of tests](https://github.com/stripe/stripe-python/pull/1165/commits/cef4396aacdb521797ae8bf56ab20e9e2491ebc4). This got tricky because the tests in `test_http_client` use `MagicMock`. For asynchronous functions you're supposed to use `AsyncMock` which exists in recent versions of our `pytest-mock`, but for python 3.6 and 3.7, we have to use an older version of `pytest-mock` which doesn't have `AsyncMock`, so I had to pull in another dependency.

tests/test_integration.py
```diff
@@ -54,9 +50,11 @@ def _do_request(self):
         status = provided_status or self.default_status
         headers = provided_headers or self.default_headers
         body = provided_body or self.default_body
+        content_length = len(body)
         self.send_response(status)
         for header_name, header_value in headers.items():
             self.send_header(header_name, header_value)
+        self.send_header("Content-Length", str(content_length))
```
pakrym-stripe, line 57: why is this needed?

tests/test_http_client.py
```diff
@@ -993,3 +1014,342 @@ def test_encode_array(self):
...
+
+            calls = [mocker.call(*args, **kwargs) for _ in range(times)]
+            request_mock.AsyncClient().request.assert_has_calls(calls)
+
+        return check_call
+
+    async def make_request_async(
+        self, method, url, headers, post_data, timeout=80
+    ):
+        client = self.REQUEST_CLIENT(
```
pakrym-stripe, line 1136: Why is this client-specific and not in the base test class?

tests/test_http_client.py
```diff
@@ -993,3 +1014,342 @@ def test_encode_array(self):
...
+            assert code == 200
+            assert body == '{"foo": "baz"}'
+
+            check_call(request_mock, method, abs_url, data, headers)
+
+    @pytest.mark.asyncio
+    async def test_request_stream(
+        self, mocker, request_mock, mock_response, check_call
+    ):
+        # TODO
```
pakrym-stripe, line 1179: File?

tests/test_http_client.py
```diff
@@ -993,3 +1014,342 @@ def test_encode_array(self):
...
+            "POST", self.valid_url, headers, data, timeout=5
+        )
+
+        check_call(None, "POST", self.valid_url, data, headers, timeout=5)
+
+    @pytest.mark.asyncio
+    async def test_request_stream_forwards_stream_param(
+        self, mocker, request_mock, mock_response, check_call
+    ):
+        # TODO
```
pakrym-stripe, line 1202: file?

tests/test_http_client.py
```diff
@@ -993,3 +1014,342 @@ def test_encode_array(self):
...
+    @pytest.mark.asyncio
+    async def test_timeout(self, request_mock, mock_response, check_call):
+        headers = {"my-header": "header val"}
+        data = {}
+        mock_response('{"foo": "baz"}', 200)
+        await self.make_request_async(
+            "POST", self.valid_url, headers, data, timeout=5
+        )
+
+        check_call(None, "POST", self.valid_url, data, headers, timeout=5)
```
pakrym-stripe, line 1196: This test method looks like it can be in the base test class as well.

tests/test_integration.py
richardm-stripe: Otherwise the client can't tell whether the request is finished or not and it will hang.
I tried `self.wfile.close()` to close the connection, but this seems to fight with another place that `BaseHTTPHandler` wants to close things, so setting content-length is more expedient. Besides, content-length is standard to include.

tests/http_client_mock.py
```diff
@@ -185,17 +185,27 @@ def assert_post_data(self, expected, is_json=False):
 
 
 class HTTPClientMock(object):
-    def __init__(self, mocker, is_streaming=False):
-        self.mock_client = mocker.Mock(
-            wraps=stripe.http_client.new_default_http_client()
-        )
+    def __init__(self, mocker, is_streaming=False, is_async=False):
```
pakrym-stripe, line 188: I wonder if we should keep adding more HTTPClientMock instances of store the information about the call in one instance of mocker (e.g. `(verb, path, ..., streaming, async)`).

pakrym-stripe: That way the same test can mix multiple kinds of call as well.

tests/test_http_client.py
richardm-stripe: It isn't for the non-async clients. Timeout isn't in the base HTTPClient class so I think it shouldn't be in the base test, either.

pakrym-stripe (changes requested):
Some questions about http client tests

tests/test_http_client.py
pakrym-stripe: Do all async clients we are planning for support timeout? Maybe timeout should be on the base client if it's supported everywhere.

richardm-stripe: What is the advantage of including it on the base client?

tests/http_client_mock.py
richardm-stripe: I can file this as a potential follow-up.

tests/test_http_client.py
pakrym-stripe: 1. We can expose timeout setting on client options and silently create an appropriate client in the background knowing that all our clients support timeout
2. We won't have to copy-paste timeout tests for every client and then accidentally miss one

richardm-stripe: Not all sync clients support timeout, and I have a slight dislike for inconsistency between the async world and sync world here.

I'm also afraid that there will be subtle differences between what "timeout" means from http client to http client. Also in terms of missing things, there's the possibility that we will add another async client, where "timeout" is settable on the base client but we forget to pipe "timeout" through to the underlying client so the client will appear to support timeout when it doesn't.

It feels more conservative to me to keep timeout client-specific. I don't think this is a trapdoor, once more async clients are added we can look to see if timeout actually does work the same between them and choose to put them in the base class then. I would prefer that to trying to make a forecast now.

pakrym-stripe: >I'm also afraid that there will be subtle differences between what "timeout" means from http client to http client.
The base test will help catch these differences.

Can you file an issue for this? We do have a requirement for timeout option to be exposed in client options. So we'll need to solve this somehow.

richardm-stripe: > Overall, I'd say that our sync client timeout implementation is not something to draw inspiration from.

Ha, very true.

I will file an issue -- I like the idea of using the AsyncClient as an opportunity to get timeouts right, but this is better to tackle separately.

tests/test_http_client.py
richardm-stripe: Filed!

tests/test_http_client.py
richardm-stripe: Filed!

tests/test_http_client.py
richardm-stripe: Oh, the reason turns out to be that `timeout` is not in the base class. Per our other discussion, I will fix this, but separately.

pakrym-stripe (approved):

PR #1168, opened 2023-12-11
Do not raise a DeprecationWarning in `stripe.app_info`

### LINKED ISSUE #1167: Importing stripe==7.8.1 raises a deprecation warning
### Describe the bug

See reproduction notes.

When `DeprecationWarning`s are set to raise and not just warn (e.g. in our test setup), this causes an immediate test failure.

### To Reproduce

```
root@ed5bfe3d7af7:/# pip install stripe
Successfully installed certifi-2023.11.17 charset-normalizer-3.3.2 idna-3.6 requests-2.31.0 stripe-7.8.1 typing-extensions-4.9.0 urllib3-2.1.0
root@ed5bfe3d7af7:/# python -Xdev -c 'import stripe'
/usr/local/lib/python3.12/site-packages/stripe/app_info.py:5: DeprecationWarning:
    The stripe.app_info package is deprecated, please change your
    imports to import from stripe directly.
    From:
      from stripe.app_info import AppInfo
    To:
      from stripe import AppInfo

  warn(
root@ed5bfe3d7af7:/#
```

### Expected behavior

No deprecation warning simply from importing the package.

### Code snippets

```Python
import stripe
```

### OS

Any

### Language version

Any

### Library version

stripe-python v7.8.1

### API version

Doesn't matter

### Additional context

_No response_

akx:
Fixes #1167.

Regressed in 0c7ad9613c43f90350fbe7e05bdcf0e84b4f75c8 (#1153).

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -27,6 +27,10 @@ per-file-ignores =
     stripe/oauth_error.py: IMP102
     # setup.py is required for tooling
     setup.py: IMP102
+    # should not raise a deprecation warning since it needs
+    # to be imported early in `stripe/__init__.py` to avoid
+    # a name conflict
+    stripe/app_info.py: IMP102
 
 [flake8:local-plugins]
 extension =
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -13,8 +13,9 @@ import os
 # Configuration variables
 from stripe._api_version import _ApiVersion
 
-# We must import the app_info module eagerly before defining the app_info global
-# otherwise the late import will overwrite the global
+# We must import the app_info module early to populate it into
+# `sys.modules`; otherwise doing `import stripe.app_info` will end up
+# importing that module, and not the global `AppInfo` name from below.
 import stripe.app_info
 from stripe._app_info import AppInfo as AppInfo
 from stripe._version import VERSION as VERSION
--- a/stripe/app_info.py
+++ b/stripe/app_info.py
@@ -1,18 +1,18 @@
 # -*- coding: utf-8 -*-
+"""
+The stripe.app_info package is deprecated, please change your
+imports to import from stripe directly.
+From:
+  from stripe.app_info import AppInfo
+To:
+  from stripe import AppInfo
+"""
 from typing_extensions import TYPE_CHECKING
-from warnings import warn
 
-warn(
-    """
-    The stripe.app_info package is deprecated, please change your
-    imports to import from stripe directly.
-    From:
-      from stripe.app_info import AppInfo
-    To:
-      from stripe import AppInfo
-    """,
-    DeprecationWarning,
-)
+# No deprecation warning is raised here, because it would happen
+# on every import of `stripe/__init__.py` otherwise. Since that
+# module declares its own `app_info` name, this module becomes
+# practically impossible to import anyway.
 
 if not TYPE_CHECKING:
     from stripe._app_info import (  # noqa


akx:
cc @pakrym-stripe

pakrym-stripe:
I'm afraid we need that line for the reason the comment outlined. 
Will wrapping that import with 

```
        with warnings.catch_warnings():
            warnings.filterwarnings("ignore", category=DeprecationWarning)
```
work?

akx:
@pakrym-stripe What is the late import the comment refers to? I tried figuring it out, but no luck.

As I see it, the line

https://github.com/stripe/stripe-python/blob/5ab12c625287a93bb6ec84d116045919153b7b87/stripe/__init__.py#L18

ends up importing `app_info`, so `app_info` is `<module 'stripe.app_info' from '/usr/local/lib/python3.12/site-packages/stripe/app_info.py'>` after that (and a deprecation warning was emitted).

Then, the name is overwritten with `None` some lines later:
https://github.com/stripe/stripe-python/blob/5ab12c625287a93bb6ec84d116045919153b7b87/stripe/__init__.py#L31

pakrym-stripe:
Yes, and `app_info` variable is used for a different purpose. If you don't import app_info module early, the late import will overwrite the `stripe.app_info`.

akx:
I'm still not following the terminology "late import" here, unless you mean this corner-case situation (which is not really what I'd call a late import), when an user imports `stripe`, then separately does `import stripe.app_info` and _not_ `from stripe import app_info` (note how the import order has an effect too):

```
(no-deprecation-warning-on-import) $ python -c 'import stripe; import stripe.app_info as ai; from stripe import app_info as ai2; print(ai, ai2)'
<module 'stripe.app_info'> <module 'stripe.app_info'>
(no-deprecation-warning-on-import) $ python -c 'import stripe; from stripe import app_info as ai2; import stripe.app_info as ai; print(ai, ai2)'
<module 'stripe.app_info'> None
```

```
(master) $ python -c 'import stripe; import stripe.app_info as ai; from stripe import app_info as ai2; print(ai, ai2)'
None None
(master) $ python -c 'import stripe; from stripe import app_info as ai2; import stripe.app_info as ai; print(ai, ai2)'
None None
```

akx:
Also, with the current setup there's actually no point in `stripe.app_info` even having the `DeprecationWarning` in the first place; since it's implicitly imported by `import stripe`, trying to re-import it would not emit the warning again (because the module will have already been executed).

On that note, if you like, @pakrym-stripe, I can rework this PR to just remove the DeprecationWarning from there.

pakrym-stripe:
> unless you mean this corner-case situation (which is not really what I'd call a late import), when an user imports stripe, then separately does import stripe.app_info

Yep, that.

```
python -c 'import stripe; import stripe.app_info; print(stripe.app_info)'
<module 'stripe.app_info' from '/Users/pakrym/stripe/stripe-python/stripe/app_info.py'>
```

If `stripe.app_info` if not eagerly imported in __init__, the separate import of `stripe.app_info` will overwrite `stripe.app_info`. 

>Also, with the current setup there's actually no point in stripe.app_info even having the DeprecationWarning in the first place; since it's implicitly imported by import stripe, trying to re-import it would not emit the warning again (because the module will have already been executed).

You are right here, but we'd like to keep all backward compatibility re-exporting files uniform. We also have have a linter that depend on the warning being there (https://github.com/stripe/stripe-python/blob/master/flake8_stripe/flake8_stripe.py#L136)

akx:
I think the suggested `warnings.catch_warnings()` filter thing is a bit of a silly waste of users' cycles (especially since it's not thread-safe, etc., etc.), if the simplest option is to get rid of the deprecation warning (with a suitable comment explaining why it's not there) and adding an exemption for that file – after all, there are `per-file-ignores` for `IMP102`s already.

pakrym-stripe:
Alright, but please leave the 
```
  The stripe.app_info package is deprecated, please change your
    imports to import from stripe directly.
    From:
      from stripe.app_info import AppInfo
    To:
      from stripe import AppInfo
``` 
part as a comment

akx:
@pakrym-stripe Okiedoke, changed. Flake8 seems to be happy locally too.

pakrym-stripe (approved):

pakrym-stripe:
Great work! Thank you for your help @akx !

PR #1170, opened 2023-12-13
HTTPClient refactor

richardm-stripe:
## Summary
Contains refactoring changes from https://github.com/stripe/stripe-python/pull/1165 that we can merge into master to minimize how much master and beta diverge.

In particular, it extracts an `HTTPClientBase` parent class from `HTTPClient`, and removes some dead code related to warning that would only fire if Python2 (which this library is no longer compatible with) is used.

### DIFF
--- a/flake8_stripe/flake8_stripe.py
+++ b/flake8_stripe/flake8_stripe.py
@@ -160,7 +160,6 @@ class StripeImportsChecker:
                     )
             if isinstance(node, ast.ImportFrom):
                 # Forbid: from stripe...module import Type
-                assert node.module is not None
                 parts = node.module.split(".")
                 if (
                     len(parts) > 1
--- a/flake8_stripe/flake8_stripe.py
+++ b/flake8_stripe/flake8_stripe.py
@@ -31,6 +31,8 @@ class TypingImportsChecker:
         "NotRequired",
         "Self",
         "Unpack",
+        "Awaitable",
+        "Never",
     ]
 
     allowed_typing_imports = [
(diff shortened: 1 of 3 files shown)


flake8_stripe/flake8_stripe.py
```diff
@@ -158,6 +160,7 @@ def run(self) -> Iterator[Tuple[int, int, str, type]]:
                     )
             if isinstance(node, ast.ImportFrom):
                 # Forbid: from stripe...module import Type
+                assert node.module is not None
```
pakrym-stripe, line 163: hm, why?

stripe/_http_client.py
```diff
@@ -315,16 +311,8 @@ def close(self):
...
-        self,
-        timeout: int = 80,
-        session: Optional["Session"] = None,
-        verify_ssl_certs: bool = True,
-        proxy: Optional[Union[str, HTTPClient._Proxy]] = None,
-    ):
-        super(RequestsClient, self).__init__(
-            verify_ssl_certs=verify_ssl_certs, proxy=proxy
-        )
+    def __init__(self, timeout=80, session=None, **kwargs):
```
pakrym-stripe, line 314: types lost

pakrym-stripe: here and else where

pakrym-stripe (approved):

flake8_stripe/flake8_stripe.py
richardm-stripe: Fixed a pyright info I think. I can remove.

PR #1175, opened 2023-12-14
Typecheck test file

richardm-stripe:


pyproject.toml
```diff
@@ -40,7 +40,7 @@ reportUnnecessaryTypeIgnoreComment = true
 [tool.mypy]
 follow_imports = "silent"
 python_version = "3.10"
-files = ["tests/test_generated_examples.py", "tests/test_exports.py"]
+files = ["tests/test_generated_examples.py", "tests/test_exports.py", "tests/test_http_client.py"]
```
pakrym-stripe, line 43: Should we do pyright instead? Considering we use it to typecheck internal code.

richardm-stripe: Good catch! I just saw a list of test files and added. Forgot there were two.

pakrym-stripe (approved):

PR #1179, opened 2023-12-18
Replace `request_mock` with `http_client_mock` in tests

anniel-stripe:
Replace `RequestMock` with `HTTPClientMock` in tests. This mocking interface introduces an easier, more robust assertion solution that does not rely on the arguments to `APIRequestor.request(....)`. Instead, it directly checks the contents of each HTTP request for stubs / assertions.


tests/api_resources/abstract/test_api_resource.py
```diff
@@ -7,54 +7,44 @@ class TestAPIResource(object):
...
 
-        request_mock.assert_requested("get", url, {"myparam": 5}, None)
+        http_client_mock.assert_requested(
+            "get", path=path, query_string=query_string
+        )
         assert res.frobble == 5
         with pytest.raises(KeyError):
             res["bobble"]
 
-    def test_request_with_special_fields_prefers_explicit(self, request_mock):
```
pakrym-stripe, line 39: Why remove?

anniel-stripe: Oops, I added this back locally but never pushed the commit

tests/api_resources/abstract/test_custom_method.py
```diff
@@ -30,87 +30,65 @@ def do_stream_stuff(self, idempotency_key=None, **params):
...
         @classmethod
-        def _cls_do_stuff_new_codegen(
-            cls,
-            id,
-            api_key=None,
-            stripe_version=None,
-            stripe_account=None,
-            **params
-        ):
+        def _cls_do_stuff_new_codegen(cls, id, **params):
```
pakrym-stripe, line 33: Are we going to make this generator change as well?

tests/api_resources/abstract/test_custom_method.py
```diff
@@ -121,26 +99,28 @@ def test_call_custom_list_method_class_paginates(self, request_mock):
...
 
-    def test_call_custom_stream_method_class(self, request_mock):
-        request_mock.stub_request_stream(
+    def test_call_custom_stream_method_class(self, http_client_mock_streaming):
+        http_client_mock_streaming.stub_request(
             "post",
-            "/v1/myresources/mid/do_the_stream_thing",
-            "response body",
+            path="/v1/myresources/mid/do_the_stream_thing",
+            rbody=util.io.BytesIO(str.encode("response body")),
```
pakrym-stripe, line 114: Why is this needed?

tests/api_resources/abstract/test_deletable_api_resource.py
```diff
@@ -5,66 +5,73 @@ class TestDeletableAPIResource(object):
...
             rheaders={"request-id": "req_id"},
         )
 
         obj = self.MyDeletable.construct_from({"id": "mid"}, "mykey")
 
         self.MyDeletable.delete(obj)
 
-        request_mock.assert_requested("delete", "/v1/mydeletables/mid", {})
+        http_client_mock.assert_requested(
+            "delete", path="/v1/mydeletables/mid", query_string=""
```
pakrym-stripe, line 40: is query_string a required parameter?

tests/api_resources/abstract/test_custom_method.py
anniel-stripe: Ah I included too many changes, reverted with https://github.com/stripe/stripe-python/pull/1179/commits/f2441927ad1cfd7337ab3331f0c59014d911a77f

tests/test_util.py
```diff
@@ -152,4 +149,6 @@ def test_convert_to_stripe_object_and_back(self):
 
     def test_sanitize_id(self):
         sanitized_id = util.sanitize_id("cu  %x 123")
+        if isinstance(sanitized_id, bytes):
```
pakrym-stripe, line 152: Where did we start getting bytes from?

tests/test_oauth.py
```diff
@@ -35,46 +35,37 @@ def test_authorize_url(self, request_mock):
...
+            rbody='{"stripe_user_id":"acct_test_deauth"}',
         )
 
         resp = stripe.OAuth.deauthorize(stripe_user_id="acct_test_deauth")
-        request_mock.assert_requested(
+        http_client_mock.assert_requested(
             "post",
-            "/oauth/deauthorize",
-            {"client_id": "ca_123", "stripe_user_id": "acct_test_deauth"},
+            api_base=stripe.connect_api_base,
```
pakrym-stripe, line 67: nice

tests/test_integration.py
```diff
@@ -204,16 +202,6 @@ def do_request(self, req_num):
         # latency shouldn't be outside this range.
         assert 30 < duration_ms < 300
 
-        usage = telemetry["last_request_metrics"]["usage"]
```
pakrym-stripe, line 207: accidental?

tests/test_integration.py
```diff
@@ -93,9 +93,9 @@ def setup_stripe(self):
         stripe.api_base = "http://localhost:12111"  # stripe-mock
         stripe.api_key = "sk_test_123"
         stripe.default_http_client = None
+        stripe._default_proxy = None
```
pakrym-stripe, line 96: intended?

tests/api_resources/abstract/test_deletable_api_resource.py
anniel-stripe: No, it isn't. `http_client_mock .assert_requested()` will not check `query_string` if it's not passed in, so I specified it because I thought we might as well be stricter about the checks.

tests/api_resources/test_file_upload.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class TestFileUpload(object):
+    @pytest.fixture(scope="function")
+    def setup_upload_api_base(self):
+        stripe.upload_api_base = stripe.api_base
+        stripe.api_base = None
+        yield
+        stripe.api_base = stripe.upload_api_base
+        stripe.upload_api_base = "https://files.stripe.com"
```
pakrym-stripe, line 18: Considering we are replacing the entire HTTP client. Why do we even need to mess with API bases? Can we always keep them default?

tests/test_util.py
anniel-stripe: Good catch, this really shouldn't be necessary. `sanitize_id` calls `quote_plus` which returns a string.

pakrym-stripe (approved):

tests/api_resources/abstract/test_custom_method.py
anniel-stripe: `request_mock` would [convert the string to bytes internally](https://github.com/stripe/stripe-python/blob/3a3454408974a76c7dcbcc1a63951793142b3d8e/tests/request_mock.py#L202) when stubbing the request. I pulled that logic out so that the mock HTTP client directly returns bytes.

tests/test_integration.py
anniel-stripe: You're right, none of the changes in this file are relevant. Removing.

tests/api_resources/test_file_upload.py
anniel-stripe: You're right, we can keep them default. Updated in https://github.com/stripe/stripe-python/pull/1179/commits/54caa2faa8b457a49d909f5eb4759a02354e4e4e

anniel-stripe: Nevermind, just realized `setup_upload_api_base` is necessary because it makes it so that `stripe.api_base = None`, so if `stripe.api_base` is mistakenly used and API base is asserted to be `stripe.upload_api_base`, the test will fail.

anniel-stripe: go/j/DEVSDK-1430

PR #1181, opened 2023-12-20
Add request setting forwarding tests

anniel-stripe:
To validate how request settings on a resource are flowed in different contexts (class vs. instance methods).

### DIFF
--- a/tests/api_resources/abstract/test_api_resource.py
+++ b/tests/api_resources/abstract/test_api_resource.py
@@ -10,6 +10,9 @@ class TestAPIResource(object):
     def test_retrieve_and_refresh(self, http_client_mock):
         path = "/v1/myresources/foo%2A"
         query_string = "myparam=5"
+        key = "sk_test_123"
+        stripe_version = "2018-02-28"
+        stripe_account = "acct_foo"
         http_client_mock.stub_request(
             "get",
             path,
@@ -29,15 +32,15 @@ class TestAPIResource(object):
             "get",
             path=path,
             query_string=query_string,
-            api_key="sk_test_123",
-            stripe_version="2018-02-28",
-            stripe_account="acct_foo",
+            api_key=key,
+            stripe_version=stripe_version,
+            stripe_account=stripe_account,
         )
         assert res.bobble == "scrobble"
         assert res.id == "foo2"
-        assert res.api_key == "sk_test_123"
-        assert res.stripe_version == "2018-02-28"
-        assert res.stripe_account == "acct_foo"
+        assert res.api_key == key
+        assert res.stripe_version == stripe_version
+        assert res.stripe_account == stripe_account
 
         assert res.last_response is not None
         assert res.last_response.request_id == "req_id"
@@ -51,7 +54,12 @@ class TestAPIResource(object):
         res = res.refresh()
 
         http_client_mock.assert_requested(
-            "get", path=path, query_string=query_string
+            "get",
+            path=path,
+            query_string=query_string,
+            api_key=key,
+            stripe_version=stripe_version,
+            stripe_account=stripe_account,
         )
         assert res.frobble == 5
         with pytest.raises(KeyError):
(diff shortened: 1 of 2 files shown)


tests/http_client_mock.py
```diff
@@ -248,7 +248,10 @@ def get_last_call(self) -> StripeRequestCall:
         return StripeRequestCall.from_mock_call(self.func.call_args)
 
     def get_all_calls(self) -> List[StripeRequestCall]:
-        return [StripeRequestCall.from_mock_call(call) for call in self.func]
+        return [
+            StripeRequestCall.from_mock_call(call_args)
+            for call_args in self.func.call_args_list
+        ]
```
anniel-stripe, line 254: I used the `get_all_calls` and `get_raw_header` methods for debugging, and fixed some bugs in them. But these 2 methods aren't used anywhere.

tests/api_resources/abstract/test_api_resource.py
```diff
@@ -101,3 +113,65 @@ def test_raise_on_incorrect_id_type(self):
         for obj in [None, 1, 3.14, dict(), list(), set(), tuple(), object()]:
             with pytest.raises(stripe.error.InvalidRequestError):
                 self.MyResource.retrieve(obj)
+
+    @pytest.fixture
+    def resource(self):
```
pakrym-stripe, line 118: nit: I would create a new resource for every test, it's not that much writing and you get to vary parameters like  `stripe_version`.

tests/api_resources/abstract/test_api_resource.py
```diff
@@ -101,3 +113,65 @@ def test_raise_on_incorrect_id_type(self):
...
+        return self.MyResource.construct_from(
+            {"id": "foo"},
+            key="newkey",
+            stripe_version="2023-01-01",
+            stripe_account="acct_foo",
+        )
+
+    @pytest.fixture
+    def check_request_options(self, http_client_mock):
```
pakrym-stripe, line 127: It might be simpler to directly assert `http_client_mock.assert_requested(` inside a test.

tests/api_resources/abstract/test_api_resource.py
```diff
@@ -101,3 +113,65 @@ def test_raise_on_incorrect_id_type(self):
...
+                "get",
+                path="/v1/myresources/foo",
+                api_key=api_key,
+                stripe_version=stripe_version,
+                extra_headers=extra_headers,
+            )
+
+        return check_request_options
+
+    def test_class_method_does_not_forward_options(
```
pakrym-stripe, line 149: Do we need a scenario where parameters are passed into the method call? `Account.create(, stripe_version = ...)

anniel-stripe: Good call, I'll add that

tests/api_resources/abstract/test_api_resource.py
```diff
@@ -101,3 +113,118 @@ def test_raise_on_incorrect_id_type(self):
...
+
+        http_client_mock.assert_requested(
+            "get",
+            path="/v1/myresources/foo",
+            api_key=key,
+            stripe_version=stripe_version,
+            stripe_account=stripe_account,
+        )
+
+    def test_instance_method_prefers_method_arguments(
```
pakrym-stripe, line 203: What a bout an instance call on a resource returned by a method call with extra parameters.
```python
acc = Account.retrieve(, stripe_account)
acc.delete()
```

pakrym-stripe: I guess that's covered by refresh example above.

pakrym-stripe (approved):

tests/api_resources/abstract/test_api_resource.py
anniel-stripe: Good point, I went and added a few more assertions to test_retrieve_and_refresh to make it more complete.

PR #1195, opened 2024-01-17
Use sys.executable in test_exports

s-t-e-v-e-n-k:
Not all systems will have 'python' in the path to execute to check exports -- and indeed, it may not even be the interpreter that is under test and can import the module. Since we are already running under the test suite, use sys.executable to check.


anniel-stripe (approved):
Good catch @s-t-e-v-e-n-k. Thanks for the contribution!

PR #1196, opened 2024-01-17
Add request mocking to two test methods

s-t-e-v-e-n-k:
These two test methods do not mock, which also means they are not skipped if the mocking server is not running. Sprinkle in the mocking.

### DIFF
--- a/tests/api_resources/test_invoice.py
+++ b/tests/api_resources/test_invoice.py
@@ -137,7 +137,7 @@ class TestInvoice(object):
         )
         assert isinstance(resource, stripe.Invoice)
 
-    def test_can_iterate_lines(self):
+    def test_can_iterate_lines(self, http_client_mock):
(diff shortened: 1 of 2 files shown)


anniel-stripe (changes requested):
>if the mocking server is not running

Our test suite depends on stripe-mock as specified in [our README](https://github.com/stripe/stripe-python?tab=readme-ov-file#development), and it should be running in the background when running the tests.

The `http_client_mock` fixture is used when we need to assert properties of the request, like in https://github.com/stripe/stripe-python/blob/c1a7c80160bd335e004fb6a788532b6f63961848/tests/api_resources/test_invoice.py#L68-L70

The tests you've changed don't assert anything yet. But if you add some assertions using `http_client_mock` about the expected method / path / query for the methods tested here, we can take those changes.

s-t-e-v-e-n-k:
They do assert things, just not about the request. But fair point.

PR #1200, opened 2024-01-19
Add StripeClient

anniel-stripe:
## StripeClient
This PR introduces a `StripeClient` and service-based pattern. This new interface allows you to easily call Stripe APIs and has many benefits over the existing resource-based pattern:

* No global config: you can simultaneously use multiple clients with different configuration options (such as API keys)
* No extra API calls. All API endpoints can be accessed with a single method call. You don't have to call `retrieve` before doing a `modify`.
* No static methods for easier mocking.

To migrate from a resource-based to service-based pattern:
1. Initialize a `StripeClient` instance.

   **Before**
   ```python
   ```

   **After**
   ```python
   client = stripe.StripeClient("sk_test_123")
   ```
2. Convert class resource method calls to `StripeClient` calls. On resource-based calls, request parameters and request options were each individually passed in as a mix of positional and keyword arguments. But for service-base calls, request parameters and request options are passed in as separate dictionary arguments.

   **Before**
   ```python
   customer = stripe.Customer.create(email="jenny.rosen@example.com", stripe_account="acct_123")
   ```

   **After**
   ```python
   customer = client.customers.create(params={"email": "jenny.rosen@example.com"}, options={"stripe_account": "acct_123"})
   ```
3. Convert instance resource method calls to `StripeClient` calls. As with class calls, request parameters and request options become separate dictionary arguments in the service-based pattern.

   **Before**
   ```python
   customer = stripe.Customer.retrieve("cus_123", options={});
   customer.delete();
   ```

   **After**
   ```python
   client.customers.delete("cus_123");
    ```
4. Convert nested resource operations to `StripeClient` calls. As with class and instance calls, request parameters and request options become separate dictionary arguments in the service-based pattern.

   **Before**
   ```python
   stripe.Customer.list_balance_transactions(
     "cus_123",
     limit=3,
     stripe_version="2022-11-15",
   )
   ```
   
   **After**
   ```python
   customer = client.customers.balance_transactions.list(
     "cus_123",
     params={"limit": 3},
     options={"stripe_version": "2022-11-15"},
   );
   ```

## Changelog

"⚠️" symbol highlights breaking changes

### ⚠️ Changed
* ⚠️ **Request options like `api_key`, `stripe_account`, `stripe_version`, and `idempotency_key` can no longer be passed in positionally on resource methods. Please pass these in as keyword arguments.**

**BEFORE**
```python
stripe.Customer.create(
  "sk_test_123",  # api key
  "KG5LxwFBepaKHyUD",  # idempotency key
  "2022-11-15",  # stripe version
  "acct_123",  # stripe account
)
```

**AFTER**
```python
stripe.Customer.create(
  api_key="sk_test_123",
  idempotency_key="KG5LxwFBepaKHyUD",
  stripe_version="2022-11-15",
  stripe_account="acct_123",
)
```
* ⚠️ Methods that turn a response stream (`Quote.pdf`) now returns a single value of type `StripeResponseStream` instead of a tuple containing `(StripeResponseStream, api_key)`.
* ⚠️ `APIRequestor.request()` and `APIRequestor.request_stream()` parameter changes:
    * Remove `headers` positional parameter
    * Add optional `options` positional parameter
    * Add required `base_address` and `api_mode` keyword-only parameter
* ⚠️ Update `APIRequestor.request()` to return a `StripeObject` (previously `Tuple[StripeResponse, str]`)
* ⚠️ Update `APIRequestor.request_stream()` to return a `StripeResponseStream` (previously `Tuple[StripeStreamResponse, str]`)
* ⚠️ `APIRequestor.request_headers()` parameters list changed from ` def request_headers(self, api_key, method)` to `def request_headers(self, method, options: RequestOptions)`. Please migrate any existing calls by passing in an API key via a `RequestOptions` dict: `self.request_headers(method, {"api_key": api_key})`.

### ⚠️ Removed
* ⚠️ Remove `api_version` from `File.create` parameters. Please use `stripe_version` instead.
* ⚠️ Remove `api_base` and `api_version` attributes on `APIRequestor`
* ⚠️ Remove `util.read_special_variable()` utility method (importing directly from `stripe.util` is deprecated as of [v7.8.0](https://github.com/stripe/stripe-python/blob/master/CHANGELOG.md#780---2023-12-07))
* ⚠️ Remove `APIRequestor.interpret_response()` and `APIRequestor.interpret_streaming_response()` - these are now private methods.


stripe/__init__.py
```diff
@@ -36,6 +43,44 @@
     os.path.dirname(__file__), "data", "ca-certificates.crt"
 )
 
+# Lazily initialized stripe.default_http_client
+default_http_client = None
+_default_proxy = None
+
+
+def ensure_default_http_client():
```
pakrym-stripe, line 51: The API requestor is the only thing that calls this method. Worth inlining the default client logic into it?

stripe/__init__.py
```diff
@@ -2,6 +2,7 @@
 from typing import Optional
 import sys as _sys
 import os
+import warnings
```
pakrym-stripe, line 5: Any particular reason to move?

stripe/_api_requestor.py
```diff
@@ -8,75 +8,113 @@
...
-    return urlunsplit((scheme, netloc, path, query, fragment))
+HttpVerb = Literal["get", "post", "delete"]
 
 
 class APIRequestor(object):
-    api_key: Optional[str]
-    api_base: str
-    api_version: str
-    stripe_account: Optional[str]
+    _instance = None
```
pakrym-stripe, line 54:
```suggestion
    _instance: ClassVar["APIRequestor"] = None
```

stripe/_api_requestor.py
```diff
@@ -8,75 +8,113 @@
...
-        else:
-            # If the stripe.default_http_client has not been set by the user
-            # yet, we'll set it here. This way, we aren't creating a new
-            # HttpClient for every request.
-            stripe.default_http_client = _http_client.new_default_http_client(
-                verify_ssl_certs=verify, proxy=proxy
-            )
-            self._client = stripe.default_http_client
-            self._default_proxy = proxy
+        self.options = options
```
pakrym-stripe, line 61: nit: private field naming (options vs _options)

stripe/_api_requestor.py
```diff
@@ -99,36 +137,51 @@ def request(
...
-        headers: Optional[Mapping[str, str]] = None,
+        options: Optional[RequestOptions] = None,
         *,
+        base_address: BaseAddress,
+        api_mode: ApiMode,
         _usage: Optional[List[str]] = None,
-    ) -> Tuple[StripeResponse, str]:
-        rbody, rcode, rheaders, my_api_key = self.request_raw(
+    ) -> "StripeObject":
+        requestor = self._replace_options(options or {})
```
pakrym-stripe, line 146: nit: can we handle options being null in  _replace_options ?

stripe/_api_requestor.py
```diff
@@ -333,75 +388,97 @@ def request_raw(
...
-                supplied_headers_dict is not None
-                and supplied_headers_dict.get("Content-Type")
-                == "multipart/form-data"
-            ):
+            if api_mode == "V1FILES":
                 generator = MultipartDataGenerator()
                 generator.add_params(params or {})
                 post_data = generator.get_post_data()
-                supplied_headers_dict[
+                headers[
```
pakrym-stripe, line 411: consider doing this in request_headers

stripe/_api_requestor.py
```diff
@@ -333,75 +388,97 @@ def request_raw(
...
+                link=dashboard_link(request_id),
             )
 
-        return rcontent, rcode, rheaders, my_api_key
+        return rcontent, rcode, rheaders
 
     def _should_handle_code_as_error(self, rcode):
         return not 200 <= rcode < 300
 
     def interpret_response(
```
pakrym-stripe, line 476:
```suggestion
    def _interpret_response(
```

stripe/_api_requestor.py
```diff
@@ -414,19 +491,25 @@ def interpret_response(
...
                 rheaders,
             )
         if self._should_handle_code_as_error(rcode):
-            self.handle_error_response(rbody, rcode, resp.data, rheaders)
+            self.handle_error_response(
+                rbody, rcode, resp.data, rheaders, api_mode
+            )
         return resp
 
     def interpret_streaming_response(
```
pakrym-stripe, line 507:
```suggestion
    def _interpret_streaming_response(
```

stripe/_api_requestor.py
```diff
@@ -99,36 +137,51 @@ def request(
...
+            params=params,
+            requestor=requestor,
+            api_mode=api_mode,
+        )
 
     def request_stream(
         self,
         method: str,
         url: str,
         params: Optional[Mapping[str, Any]] = None,
```
pakrym-stripe, line 170: I think we talked about it but why didn't we make params and options keyword only?

stripe/_encode.py
```diff
@@ -29,17 +46,10 @@ def _api_encode(data) -> Generator[Tuple[str, Any], None, None]:
...
 
 
-def _encode_datetime(dttime: datetime.datetime):
-    if dttime.tzinfo and dttime.tzinfo.utcoffset(dttime) is not None:
-        utc_timestamp = calendar.timegm(dttime.utctimetuple())
-    else:
-        utc_timestamp = time.mktime(dttime.timetuple())
-
-    return int(utc_timestamp)
+def _build_api_url(url, query):
```
pakrym-stripe, line 49: this is only used in API requestor. Consider inlining.

stripe/_http_client.py
```diff
@@ -568,6 +595,11 @@ def close(self):
         pass
 
 
+class _Proxy(TypedDict):
```
pakrym-stripe, line 598: There is another copy of this type in the file https://github.com/stripe/stripe-python/blob/8f1168417aa0815e1e9f6229dca031a9ae72d34c/stripe/_http_client.py#L107

tests/test_util.py
```diff
@@ -135,7 +135,7 @@ def test_convert_to_stripe_object_and_back(self):
             "livemode": False,
         }
 
-        obj = util.convert_to_stripe_object(resp)
+        obj = util.convert_to_stripe_object(resp, api_mode="V1")
```
pakrym-stripe, line 138: is this now required? Can we keep the test as it was to validate that the behavior didn't change?

tests/test_util.py
```diff
@@ -149,4 +149,6 @@ def test_convert_to_stripe_object_and_back(self):
 
     def test_sanitize_id(self):
         sanitized_id = util.sanitize_id("cu  %x 123")
+        if isinstance(sanitized_id, bytes):
```
pakrym-stripe, line 152: when is this ever true?

tests/api_resources/test_search_result_object.py
```diff
@@ -52,7 +52,7 @@ def test_empty_search_result(self):
 
     def test_iter(self):
         arr = [{"id": 1}, {"id": 2}, {"id": 3}]
-        expected = stripe.util.convert_to_stripe_object(arr)
+        expected = stripe.util.convert_to_stripe_object(arr, api_mode="V1")
```
pakrym-stripe, line 55: consider leaving unchanged

tests/api_resources/test_quote.py
```diff
@@ -142,7 +142,7 @@ def test_can_list_computed_upfront_line_items_classmethod(
 
     def test_can_pdf(self, setup_upload_api_base, http_client_mock_streaming):
         resource = stripe.Quote.retrieve(TEST_RESOURCE_ID)
-        stream, _ = resource.pdf()
+        stream = resource.pdf()
```
pakrym-stripe, line 145: add to a list of breaking changes

tests/api_resources/test_file.py
```diff
@@ -69,7 +69,7 @@ def test_create_respects_api_version(
     ):
         test_file = tempfile.TemporaryFile()
         stripe.File.create(
-            purpose="dispute_evidence", file=test_file, api_version="foo"
+            purpose="dispute_evidence", file=test_file, stripe_version="foo"
```
pakrym-stripe, line 72: breaking change?

tests/api_resources/test_customer.py
```diff
@@ -197,7 +197,7 @@ def test_is_listable(self, http_client_mock):
     def test_is_listable_on_object(self, http_client_mock):
         resource = stripe.Customer.retrieve(
             TEST_RESOURCE_ID
-        ).list_payment_methods(TEST_RESOURCE_ID, type="card")
+        ).list_payment_methods(type="card")
```
pakrym-stripe, line 200: What happened here?

tests/api_resources/abstract/test_test_helpers_api_resource.py
```diff
@@ -16,9 +15,12 @@ def __init__(self, resource):
 
             def do_stuff(self, idempotency_key=None, **params):
                 url = self.instance_url() + "/do_the_thing"
-                headers = util.populate_headers(idempotency_key)
                 self.resource.refresh_from(
-                    self.resource.request("post", url, params, headers)
+                    self.resource.request(
```
pakrym-stripe, line 19: should this method become private?

tests/api_resources/abstract/test_custom_method.py
```diff
@@ -20,43 +20,36 @@ class MyResource(stripe.api_resources.abstract.APIResource):
 
         def do_stuff(self, idempotency_key=None, **params):
             url = self.instance_url() + "/do_the_thing"
-            headers = util.populate_headers(idempotency_key)
-            self.refresh_from(self.request("post", url, params, headers))
+            self.refresh_from(
```
pakrym-stripe, line 23: _request_and_refresh is what we use in the generated implementation

pakrym-stripe (changes requested):
Great work! Some questions and comments.

tests/api_resources/test_customer.py
anniel-stripe: This was never actually supported syntax since it's being called on a customer instance - `TEST_RESOURCE_ID` was being use as the `idempotency_key` here. Caught this by changing request options to be keyword-only.

tests/api_resources/test_file.py
anniel-stripe: Yep! Adding to Changelog

stripe/__init__.py
anniel-stripe: Done in https://github.com/stripe/stripe-python/pull/1200/commits/458614d69533618d18ed0dc6629caaca31ab4b8a

stripe/_encode.py
anniel-stripe: Done in https://github.com/stripe/stripe-python/pull/1200/commits/458614d69533618d18ed0dc6629caaca31ab4b8a

stripe/__init__.py
anniel-stripe: It was also used in `_ensure_default_http_client()`,  but I moved this import back to its original location.

stripe/_api_requestor.py
anniel-stripe: I considered it when implementing but since the `MultipartDataGenerator` needs params and the header value needs the generator's boundary value, I decided it made more sense to set the header here than in request headers.

stripe/_api_requestor.py
anniel-stripe: In general, I only made new parameters keyword-only if they are required or have the potential to become required, but it would be disruptive to insert it into the middle of a list of positional parameters.

tests/test_util.py
anniel-stripe: No it's not required, I'll keep the test as is

tests/test_util.py
anniel-stripe: It never is :)

tests/api_resources/abstract/test_test_helpers_api_resource.py
anniel-stripe: A private `_request` already exists. I plan to remove this public method in a follow-up PR, since it can be replaced with `_request_and_refresh` where it's used.

stripe/_stripe_object.py
```diff
@@ -111,13 +114,35 @@ def __init__(
         self._retrieve_params = params
         self._previous = None
 
-        object.__setattr__(self, "api_key", api_key)
-        object.__setattr__(self, "stripe_version", stripe_version)
-        object.__setattr__(self, "stripe_account", stripe_account)
```
anniel-stripe, line 116: Just realized that because I've made these properties, they aren't settable in the same way as before. Looking into how to retain the previous setting behavior.

anniel-stripe: https://github.com/stripe/stripe-python/pull/1201

pakrym-stripe (approved):
Nice work!

stripe/__init__.py
```diff
@@ -36,6 +42,10 @@
     os.path.dirname(__file__), "data", "ca-certificates.crt"
 )
 
+# Lazily initialized stripe.default_http_client
+default_http_client = None
+_default_proxy = None
```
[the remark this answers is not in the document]
anniel-stripe: yep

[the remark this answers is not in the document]
anniel-stripe: Looks like `default_http_client` is already initialized [here](https://github.com/stripe/stripe-python/blob/5c002d3ebfbfd65f365191dd5173edcb3f08b761/stripe/__init__.py#L31) and this is a re-declaration, so I've removed the extra one too in https://github.com/stripe/stripe-python/pull/1200/commits/c50f5f1c010e6215f94158e688fb55a0918921bb

stripe/_stripe_client.py
```diff
@@ -0,0 +1,254 @@
...
+        self.setup_attempts = SetupAttemptService(self._requestor)
+        self.setup_intents = SetupIntentService(self._requestor)
+        self.shipping_rates = ShippingRateService(self._requestor)
+        self.sources = SourceService(self._requestor)
+        self.subscriptions = SubscriptionService(self._requestor)
+        self.subscription_items = SubscriptionItemService(self._requestor)
+        self.subscription_schedules = SubscriptionScheduleService(
+            self._requestor
+        )
+        self.tax = TaxService(self._requestor)
```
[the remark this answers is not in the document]
anniel-stripe: Screenshot 2024-01-23 at 9 41 08 AM
Yes!

stripe/_stripe_service.py
```diff
@@ -0,0 +1,13 @@
+from stripe._api_requestor import APIRequestor
+from stripe._client_options import _ClientOptions
+
+from typing import Optional
+
+
+class StripeService(object):
+    _requestor: APIRequestor
+    _options: Optional[_ClientOptions]
```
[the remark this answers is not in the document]
anniel-stripe: `_options` is used by `OAuthService` to access the `client_id` client option. I suppose we could just define `_options` or `client_id` specifically on `OAuthService`, but I figured I'd keep `_options` on the service in case more services will need access to client options. Do you have a preference?

stripe/_list_object.py
```diff
@@ -260,14 +204,11 @@ def previous_page(
                 "Unexpected: element in .data of list object had no id"
             )
 
-        params_with_filters = self._retrieve_params.copy()
+        params_with_filters = dict(self._retrieve_params).copy()
```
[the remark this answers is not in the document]
anniel-stripe: Very good point

stripe/_stripe_object.py
```diff
@@ -111,13 +114,35 @@ def __init__(
...
+            )
+            if _requestor is None
+            else _requestor
+        )
 
         if id:
             self["id"] = id
 
+    @property
+    def requestor(self):
```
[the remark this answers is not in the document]
anniel-stripe: Nice catch, removing this property. There's no reason to make it public.

stripe/_stripe_client.py
```diff
@@ -0,0 +1,254 @@
...
+                proxy=proxy, verify_ssl_certs=verify_ssl_certs
+            )
+
+        self._requestor = APIRequestor(
+            options=requestor_options,
+            client=http_client,
+        )
+
+        self._options = _ClientOptions(
+            requestor_options=requestor_options,
```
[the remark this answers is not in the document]
anniel-stripe: I kept it on because in my mind client options should be a superclass of requestor options, but you're right that there's no functional need to keep it on the client options class. I'll remove it

stripe/_stripe_service.py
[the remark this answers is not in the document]
anniel-stripe: Removed in https://github.com/stripe/stripe-python/pull/1200/commits/e5b202715ac8b173bd686d7b32d83f2b9e67808e

stripe/__init__.py
```diff
@@ -210,105 +232,288 @@ def __getattr__(name):
...
 )
 from stripe._reversal import Reversal as Reversal
 from stripe._review import Review as Review
+from stripe._review_service import ReviewService as ReviewService
 from stripe._setup_attempt import SetupAttempt as SetupAttempt
+from stripe._setup_attempt_service import (
+    SetupAttemptService as SetupAttemptService,
+)
 from stripe._setup_intent import SetupIntent as SetupIntent
+from stripe._setup_intent_service import (
```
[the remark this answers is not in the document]
anniel-stripe: Added in https://github.com/stripe/stripe-python/pull/1200/commits/532cbe469d04faff55534ae761f113052dec356b

PR #1201, opened 2024-01-22
Allow setting api_key, stripe_version, and stripe_account on StripeObject

anniel-stripe:
Re: https://github.com/stripe/stripe-python/pull/1200/files#r1462439802

While looking through https://github.com/stripe/stripe-python/pull/1200 for breaking changes, I discovered that changing ["api_key", "stripe_account", and "stripe_version" attributes to _properties_ ](https://github.com/stripe/stripe-python/pull/1200/files#diff-fa6c706090a75f1b4f7a788687aeaa3d36d89fa1b9fc34b98794094984881345R130-R145)on the StripeObject (and APIRequestor) would break anyone who might be setting those attributes on an API resource in a confusing way.

The following code snippet uses "new_key" in the current version of stripe-python master, but uses "old_key" in the implementation in https://github.com/stripe/stripe-python/pull/1200:
```python
stripe.api_key = "old_key"

cus = stripe.Customer.retrieve("cus_123")
cus.api_key = "new_key"
cus.delete()  # Should use new_key?
```
Even more confusingly, the implementation in https://github.com/stripe/stripe-python/pull/1200 causes `"api_key"` to be set on the underlying dictionary of the StripeObject:
```python
>>> cus
<Customer customer id=cus_123 at 0x105c6c900> JSON: {
  "address": null,
  "api_key": "sk_test_123",
  "balance": 0,
  ...
}
```

### Alternatives considered
Keep the behavior as is in https://github.com/stripe/stripe-python/pull/1200, knowing that "api_key", etc. will no longer be settable on the StripeObject's requestor. However, I think it's worth supporting this setting behavior so users don't encounter the silent behavioral change outlined above.

### DIFF
--- a/tests/test_stripe_object.py
+++ b/tests/test_stripe_object.py
@@ -413,3 +413,19 @@ class TestStripeObject(object):
         assert orig_requestor is not new_requestor
         assert obj.api_key == "newkey"
         assert orig_requestor.api_key == "origkey"
+
+    def test_can_update_api_key(self, http_client_mock):
+        obj = stripe.stripe_object.StripeObject("id", "key")
+
+        http_client_mock.stub_request(
+            "get",
+            path="/foo",
+        )
+
+        obj.api_key = "key2"
+        obj.request("get", "/foo")
+
+        http_client_mock.assert_requested(
+            api_key="key2",
+            stripe_account=None,
+        )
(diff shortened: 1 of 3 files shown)


tests/test_stripe_object.py
```diff
@@ -413,3 +413,21 @@ def test_refresh_from_creates_new_requestor(self):
         assert orig_requestor is not new_requestor
         assert obj.api_key == "newkey"
         assert orig_requestor.api_key == "origkey"
+
+    def test_can_update_api_key(self, http_client_mock):
```
anniel-stripe, line 417: This test also passes in master

pakrym-stripe (approved):

PR #1203, opened 2024-01-23
Deprecate `StripeObject.request`

anniel-stripe:
And remove internal usages. This was only used in `Charge.mark_as_safe / Charge.mark_as_fraudulent` and `Source.detach` to accomplish the same behavior already provided by `_request_and_refresh`.

### DIFF
--- a/stripe/_stripe_object.py
+++ b/stripe/_stripe_object.py
@@ -383,6 +383,27 @@ class StripeObject(Dict[str, Any]):
     def api_base(cls) -> Optional[str]:
         return None
 
+    @_util.deprecated(
+        "This will be removed in a future version of stripe-python."
+    )
+    def request(
+        self,
+        method: Literal["get", "post", "delete"],
+        url: str,
+        params: Optional[Dict[str, Any]] = None,
+        *,
+        base_address: BaseAddress = "api",
+        api_mode: ApiMode = "V1",
+    ) -> "StripeObject":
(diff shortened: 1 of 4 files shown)


stripe/_stripe_object.py
```diff
@@ -383,6 +383,9 @@ def _refresh_from(
     def api_base(cls) -> Optional[str]:
         return None
 
+    @_util.deprecated(
```
pakrym-stripe, line 386: Aren't we already removing a bunch of other stuff? Or do you think this one is too popular?

pakrym-stripe: Also, what happens when someone asks for an alternative?

pakrym-stripe (approved):

stripe/_stripe_object.py
anniel-stripe: This method seemed more likely to be used IMO. But we did remove something similar in stripe-java (https://github.com/stripe/stripe-java/blob/6c3dc053cb7945c8a8fa20ffb828b0a4cd7225fa/CHANGELOG.md?plain=1#L255)

anniel-stripe: Hmm, good point. I guess the alternative is `APIRequestor.request(...)` with the StripeObject's request options. But that goes to show that this really doesn't make sense to use. I'll remove it like we did for stripe-java

PR #1204, opened 2024-01-23
Remove more deprecated methods

anniel-stripe:
Did an audit of methods / classes with `@_util.deprecated` decorators. I removed ones that were for internal stripe-python use only (like `ListObject.empty_list()`), as well as interfaces that will already be broken since we made request options keyword-only in https://github.com/stripe/stripe-python/pull/1200.

## Changelog
* ⚠️ Remove `APIRequestor.format_app_info()`. This method was intended for internal stripe-python use only.
* ⚠️ Remove `StripeError.construct_error_object()`. This method was intended for internal stripe-python use only.
* ⚠️ Remove `ListObject.empty_list()`. This method was intended for internal stripe-python use only.
* ⚠️ Remove `SearchResultObject.empty_search_result()`. This method was intended for internal stripe-python use only.
* ⚠️ Remove `StripeObject.ReprJSONEncoder`. This class was intended for internal stripe-python use only.
* ⚠️ Remove `StripeObject.api_base`. This property was defunct and returned `None`.

### DIFF
--- a/stripe/_api_requestor.py
+++ b/stripe/_api_requestor.py
@@ -43,7 +43,6 @@ from stripe._app_info import AppInfo
 
 from stripe._base_address import BaseAddress
 from stripe._api_mode import ApiMode
-from stripe import _util
 
 if TYPE_CHECKING:
     from stripe._stripe_object import StripeObject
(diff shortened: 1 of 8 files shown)


pakrym-stripe (approved):

stiliyanaa (commented):
[]()

PR #1207, opened 2024-01-24
Remove APIRequestor from public API

anniel-stripe:
## Changelog
* ⚠️ Removed public access to `APIRequestor`. `APIRequestor`'s main use is internal, and we don't have a good understanding of its external use cases. We had to make several breaking changes to its interface as part of this update, so rather than leaving it public we made it private. If you have a use case for `APIRequestor`, please open up a Github issue describing it. We'd rather you rely on something specifically designed for your use case than having to reach into the library's internals.


pakrym-stripe (approved):

richardm-stripe (approved):
LGTM!

As discussed, we could change the internal `_global_with_options` etc. to not have underscores now that the entire module is private -- but that is non-breaking, can happen later, and I think it is better to keep diffs small.

Issue #327, opened 2017-06-12
Feature request: asyncio support


stephenmuss:
Hi folks,

I was wondering what it would take to get this library to support asyncio. This would mean it would be possible to use Python 3.5+ async/await coroutines with a number of async frameworks.

Looking at the code it would be possible to create a new Http Client with say aiohttp, but it looks like the code would need to be modified in other places in order to support it.

Has the Stripe team considered async support?

brandur-stripe:
Hi Stephen,

I really like this idea!

The bad news is that I don't know if we're going to be able to do it in the immediate future :/ The library is currently primarily targets Python 2, and we use "2to3" for compatibility across both major versions. I just ran some analytics against the Stripe platform, and requests from Python 2 outnumber requests from Python 3 by a full 10 to 1, so for the time being, this setup still makes sense for the project.

It might be possible to invert the primary major version by using something like "3to2" instead, but I suspect that it might not support really fancy Python 3+ constructs like async/await. Also, for now, [the asyncio docs seem to suggest avoiding the package if wide version support is necessary](https://docs.python.org/3/library/asyncio-task.html):

> The async def type of coroutine was added in Python 3.5, and is recommended if there is no need to support older Python versions.

Given that Python's 2 to 3 chasm seems to be a problem that might never be solved, I suspect that a more plausible way to get this in would be start maintaining a separate Python 3 fork that can leverage the richer features of 3+. Unfortunately, I can't promise this anytime soon because we'd really need to ramp up out internal language teams to make it happen.

Even so, I'm going to leave this open and tag it as "future", and hopefully it'll become tractable given a little more time.

FranzPoize:
Any news on this ?

ob-stripe:
@FranzPoize @brandur-stripe's comment above is still largely valid. While we no longer use 2to3, we still need to support Python 2 for the foreseeable future (more than half of Stripe's Python users are on 2.7), and we currently don't have the bandwidth to start a new Py3-only library.

glyph:
Just food for thought: one way to make asyncio-compatible code with python 2 support is to support Twisted :).

nathanielcompton:
Leaving a +1 on this issue to further raise awareness.

Now that we are within 18 months of Python 2's [end of lifecycle support](https://www.python.org/dev/peps/pep-0373/),  I imagine there will be an ever-increasing trend of production-ready `stripe-python` implementations switching from Python 2 to 3.

jamiesonbecker:
[...]

PR #1210, opened 2024-01-25
Beta: better support for trio in HTTPClientAsync

richardm-stripe:
## Changelog
  * Fixes support for `trio` on HttpClientAsync.

## Details
Acts on suggestions by @zac-hd on https://github.com/stripe/stripe-python/issues/327#issuecomment-1907298722.
  * Requires that the `anyio` module be available alongside `httpx` as a prerequisite for using the async HTTPX client. HTTPX has a dependency on anyio anyway, so there is no downside to this.
  * Changes the usage of `asyncio.sleep` to `anyio.sleep` -- this means that users of `trio` and `asyncio` will both be able to use the httpx client without any additional configuration.
  * Starts running all the async test cases with against both `asyncio` and `trio` backends. This is accomplished by adding the `pytest.mark.anyio` marker on all the asynchronous tests (where previously I was using the pytest-asyncio marker). The pytest integration comes bundled in with `anyio` (which I have added to the test dependencies), and is described at https://anyio.readthedocs.io/en/stable/testing.html. The default behavior of this marker is to run everything on all supported backends (so both asyncio and trio). I can confirm that tests are indeed running with trio, because I had errors when I only did `pip install anyio` instead of `pip install anyio[trio]`.

##

### DIFF
--- a/stripe/_http_client.py
+++ b/stripe/_http_client.py
@@ -986,8 +986,16 @@ class HTTPXClient(HTTPClientAsync):
     ):
         super(HTTPXClient, self).__init__(**kwargs)
 
-        assert httpx is not None
-        assert anyio is not None
+        if httpx is None:
+            raise ImportError(
+                "Unexpected: tried to initialize HTTPXClient but the httpx module is not present."
+            )
+
+        if anyio is None:
+            raise ImportError(
+                "Unexpected: tried to initialize HTTPXClient but the anyio module is not present."
+            )
+
         self.httpx = httpx
         self.anyio = anyio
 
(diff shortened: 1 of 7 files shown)


stripe/_http_client.py
```diff
@@ -986,7 +987,9 @@ def __init__(
         super(HTTPXClient, self).__init__(**kwargs)
 
         assert httpx is not None
```
pakrym-stripe, line 989: Consider throwing an exception with a useful error message.

pakrym-stripe (approved):

stripe/_http_client.py
richardm-stripe: Done! [Better assertions](https://github.com/stripe/stripe-python/pull/1210/commits/b2bc5a5d4ded80e2dd5555bb98c0f3f0e35528c0)

PR #1212, opened 2024-01-26
Refactors for beta

richardm-stripe:
When merging master into beta today (https://github.com/stripe/stripe-python/pull/1211) @anniel-stripe and I encountered some merge conflicts because `beta` branch had some refactoring that `master` branch did not. 

- Inside a test there is a `TestHandler` helper class. Python doesn't like that this begins with `Test` so in the beta branch this was renamed to `MyTestHandler`.
- There is a lot of code duplication between `APIRequestor.request_raw` and `APIRequestor.request_raw_async`. In the beta branch, a lot of the common logic was factored out. We undid this when resolving the merge conflict, but here is the same factoring out. You can see the full effect in the [beta branch](https://github.com/stripe/stripe-python/pull/new/richardm-refactors-for-beta-in-beta).

### DIFF
--- a/stripe/_api_requestor.py
+++ b/stripe/_api_requestor.py
@@ -366,18 +366,17 @@ class _APIRequestor(object):
 
         return headers
 
-    def request_raw(
+    def _args_for_request_with_retries(
         self,
         method: str,
         url: str,
         params: Optional[Mapping[str, Any]] = None,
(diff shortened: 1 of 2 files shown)


anniel-stripe (approved):
LGTM!

PR #1215, opened 2024-01-26
Standardize method implementations: sanitize_id

richardm-stripe:
## Summary
* Always use `_util.sanitize_id` instead of `urllib.parse.quote_plus` (the former is an alias for the latter).
* Consistently pass `params=params` as a keyword argument (not a positional argument) into `cls._static_request` in api resources.
* Formatting changes

These changes should be no-ops and allow us to simplify the code in our generator.


helenye-stripe (approved):
lgtm!

PR #1232, opened 2024-02-10
Improve types in _http_client.py

richardm-stripe:
Adds more type annotations to `_http_client.py`.

Mostly just making types narrower / describe what exists. I was really hoping to be able to make the type for streaming endpoints more specific than `Any` but, because `stripe.default_http_client` can change dynamically, you really can't do much better.

### DIFF
--- a/stripe/_http_client.py
+++ b/stripe/_http_client.py
@@ -234,7 +234,9 @@ class HTTPClientBase(object):
         sleep_seconds *= 0.5 * (1 + random.uniform(0, 1))
         return sleep_seconds
 
-    def _add_telemetry_header(self, headers):
+    def _add_telemetry_header(
+        self, headers: Mapping[str, str]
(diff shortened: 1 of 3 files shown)


stripe/_http_client.py
```diff
@@ -533,16 +607,32 @@ def __init__(
         assert urlfetch is not None
         self.urlfetch = urlfetch
 
-    def request(self, method, url, headers, post_data=None):
+    def request(
+        self, method, url, headers, post_data=None
```
anniel-stripe, line 611: nit: more types for arguments?

anniel-stripe (approved):

stripe/_http_client.py
richardm-stripe: Done in https://github.com/stripe/stripe-python/pull/1232/commits/35322442b9b0bec00137ff48e11db29f96b79868

PR #1233, opened 2024-02-13
Beta: async streaming

richardm-stripe:
## Summary
Provides an implementation for `request_stream_async` for `HTTPClientAsync`, so e.g. `await stripe.Quote.pdf_async(...)` will work.

```python
chunks = []
result = await stripe.Quote.pdf_async("qt_123")
async for chunk in result.stream:
    chunks.append(str(chunk, "utf-8"))
```

## Details

* There's a new public type [StripeStreamResponseAsync](https://github.com/stripe/stripe-python/pull/1233/files#diff-0c0a06a2ab83d966f0031559bfbd6de9efa84df9ffc5ebd085ea1db7f555383bR49), which is the async analog of StripeStreamResponse. (This is the type that will be returned to the user when calling `stripe.Quote.pdf_async`). The difference from StripeStreamResponse is that the stream is represented by `stream: AsyncIterable[bytes]` instead of `io: IOBase`. 
  * FYI: `IOBase` implements `Iterable[bytes]` -- so there's more symmetry between `StripeStreamResponse` and `StripeResponse` than appears at first glance.
* APIRequestor has been modified to use/produce it accordingly.
* The return type of `HTTPClientAsync.request_stream_async` is now `Tuple[AsyncIterable[bytes], ...]` and an implementation is provided for HTTPXHttpClient.

I was worried about a resource leak -- inside HTTPXHttpClient it calls `response = httpx.ASyncClient().send(...)` and then takes `content = response.aiter_bytes()` and only returns `content` to the caller. Does `response` get closed properly? I manually tested, and yes, after you do an `async for` through the bytes in `content` the response gets closed automatically.

### DIFF
--- a/stripe/_api_requestor.py
+++ b/stripe/_api_requestor.py
@@ -298,8 +298,6 @@ class _APIRequestor(object):
             _usage=_usage,
         )
         resp = await self._interpret_streaming_response_async(
-            # TODO: should be able to remove this cast once self._client.request_stream_with_retries
-            # returns a more specific type.
             stream,
             rcode,
             rheaders,
--- a/stripe/_api_requestor.py
+++ b/stripe/_api_requestor.py
@@ -3,6 +3,7 @@ import json
 import platform
 from typing import (
     Any,
+    AsyncIterable,
     Dict,
     List,
     Mapping,
@@ -12,7 +13,12 @@ from typing import (
     cast,
     ClassVar,
 )
-from typing_extensions import TYPE_CHECKING, Literal, NoReturn, Unpack
+from typing_extensions import (
+    TYPE_CHECKING,
+    Literal,
+    NoReturn,
+    Unpack,
+)
 import uuid
 from urllib.parse import urlsplit, urlunsplit
 
@@ -33,7 +39,11 @@ from stripe._encode import (
     _api_encode,
     _json_encode_date_callback,
 )
-from stripe._stripe_response import StripeResponse, StripeStreamResponse
+from stripe._stripe_response import (
+    StripeResponse,
+    StripeStreamResponse,
+    StripeStreamResponseAsync,
+)
 from stripe._request_options import RequestOptions, merge_options
 from stripe._requestor_options import (
     RequestorOptions,
@@ -276,7 +286,7 @@ class _APIRequestor(object):
         base_address: BaseAddress,
         api_mode: ApiMode,
         _usage: Optional[List[str]] = None,
-    ) -> StripeStreamResponse:
+    ) -> StripeStreamResponseAsync:
         stream, rcode, rheaders = await self.request_raw_async(
             method.lower(),
             url,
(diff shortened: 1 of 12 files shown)


stripe/_api_requestor.py
```diff
@@ -838,6 +862,7 @@ def _interpret_streaming_response(
                 raise NotImplementedError(
                     "HTTP client %s does not return an IOBase object which "
                     "can be consumed when streaming a response."
+                    % self._get_http_client().name
```
richardm-stripe, line 865: drive-by bugfix in this error message

stripe/_api_requestor.py
```diff
@@ -266,7 +276,7 @@ def request_stream(
         )
         return resp
 
-    async def request_stream_async(
+    async def _request_stream_async(
```
anniel-stripe, line 279: nit method naming consistency: Was there a specific reason to prefix this with an underscore? `APIRequestor` is not publically exposed now so every method is "private", and the other `request_...` methods in the class are not prefixed.

stripe/_http_client.py
```diff
@@ -462,17 +463,45 @@ async def sleep_async(cls: Type[Self], secs: float) -> Awaitable[None]:
...
-        method,
-        url,
-        headers,
+        method: str,
+        url: str,
+        headers: Mapping[str, str],
         post_data,
-        is_streaming,
-        max_network_retries,
+        is_streaming: Literal[False],
```
anniel-stripe, line 473: Why do we need this overload? Is this not the same as `is_streaming: bool` in the original method definition?

tests/test_api_requestor.py
```diff
@@ -324,9 +332,9 @@ async def test_empty_methods_streaming_response_async(
             http_client_mock_streaming_async.assert_requested(
                 meth, post_data=post_data
             )
-            assert isinstance(resp, StripeStreamResponse)
+            assert isinstance(resp, StripeStreamResponseAsync)
 
-            assert resp.io.getvalue() == b"thisisdata"
+            assert b"".join([x async for x in resp.stream()]) == b"thisisdata"
```
anniel-stripe, line 337: Do you think it would be helpful to add a `read()` method that does this for the user?

anniel-stripe (approved):
LGTM! Left a few nits / questions but nothing blocking

tests/test_api_requestor.py
richardm-stripe: Hmm, I'll consider in a follow-up.

stripe/_http_client.py
richardm-stripe: pyright complains if you only have a single overload. The other overload has a more narrow return type so if you want that one you have to include this one as well.

stripe/_api_requestor.py
richardm-stripe: No reason, just sloppiness/forgetfulness from me. Will revert.

richardm-stripe: Fixed in [Revert unnecessary underscore](https://github.com/stripe/stripe-python/pull/1233/commits/f02f30fe6f2fd70a593e1369a201e34b451e431c)

PR #1242, opened 2024-02-16
Testing: unify http client mock

richardm-stripe:
## Summary

It adds a lot of complexity / opportunity to get things wrong in our testing framework that you always have to specify `http_client_mock_streaming` vs `http_client_mock` (and in the beta branch, it balloons and you have `http_client_mock_async` and `http_client_mock_streaming_async`). This PR changes [HTTPClientMock](https://github.com/stripe/stripe-python/compare/master...richardm-unify-http-client-mock#diff-d5ac8afe42dceff6fd78320deaa49597671c11971999548cc0417d75379a9837R214), instead of constructing it with `is_streaming` and this determining which `self.func` was set, we instead have `self.funcs` and combine the mocking functionality to operate over multiple mocked functions.

### DIFF
--- a/tests/api_resources/abstract/test_custom_method.py
+++ b/tests/api_resources/abstract/test_custom_method.py
@@ -109,8 +109,8 @@ class TestCustomMethod(object):
 
         assert ids == ["cus_1", "cus_2", "cus_3"]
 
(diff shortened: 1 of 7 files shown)


anniel-stripe (approved):
Thanks so much for doing this!!

PR #1247, opened 2024-02-22
Make `ListObject.auto_paging_iter()` implement `AsyncIterator`

richardm-stripe:
## Summary
This PR gets rid of `.auto_paging_iter_async()` and makes `.auto_paging_iter()` capable of asynchronous iteration.

**Not included:** the analog for SearchResult. Want feedback on this approach before I apply it everywhere.

## Motivation
We got some feedback from internal users that it error-prone figuring out how to migrate sync code that used `.auto_paging_iter`, e.g.
```diff
- print(", ".join([cus.id for cus in stripe.Customer.list().auto_paging_iter()]))
+ print(", ".join([cus.id async for cus in stripe.Customer.list_async().auto_paging_iter_async()]))
```

There are 3 things you have to change, and so 2^3 - 2 = 6 ways to incorrectly/partially migrate this (minus one for the correct migration, minus one for making no change at all). Here are 3:
```python
# This throws
print(", ".join([cus.id async for cus in stripe.Customer.list_async().auto_paging_iter()]))
# This succeeds but will make synchronous calls
print(", ".join([cus.id for cus in stripe.Customer.list_async().auto_paging_iter()]))
# This succeeds but makes a synchronous call.
print(", ".join([cus.id async for cus in stripe.Customer.list().auto_paging_iter_async()]))
```

I experimented a little bit with [introducing ListObjectAsync and SearchResultAsync](https://github.com/stripe/stripe-python/compare/richardm-async-list-object?expand=1), this added *a lot* of complexity and would have resulted in

```python
# This throws and gives a type error
print(", ".join([cus.id async for cus in stripe.Customer.list_async().auto_paging_iter()]))
# This throws and gives a type error
print(", ".join([cus.id for cus in stripe.Customer.list_async().auto_paging_iter()]))
# This succeeds but makes a synchronous call.
print(", ".join([cus.id async for cus in stripe.Customer.list().auto_paging_iter_async()]))
```

This PR, on the other hand, makes it so that it is not a mistake to use `async for` along with `.auto_paging_iter()`. So it reduces the possible ways to migrate incorrectly to 2^2 - 2 = 2.

```python
# This succeeds but makes synchronous calls
print(", ".join([cus.id for cus in stripe.Customer.list_async().auto_paging_iter()]))
# This succeeds but makes a synchronous call.
print(", ".join([cus.id async for cus in stripe.Customer.list().auto_paging_iter()]))
```

As a follow-up, we could consider adding (and recommending) an option like `stripe.default_http_client = stripe.HTTPXClient(disable_sync=True)` that would cause these to begin throwing exceptions if the user wanted.

### DIFF
--- a/stripe/_list_object.py
+++ b/stripe/_list_object.py
@@ -16,27 +16,11 @@ from typing import (
 from stripe._api_requestor import (
     _APIRequestor,  # pyright: ignore[reportPrivateUsage]
 )
+from stripe._any_iterator import AnyIterator
 from stripe._stripe_object import StripeObject
 from stripe._request_options import RequestOptions, extract_options_from_dict
 
 from urllib.parse import quote_plus
-
-
-TIter = TypeVar("TIter", bound=StripeObject)
-
-
-class SyncAsyncIterator(Iterator[TIter], AsyncIterator[TIter]):
-    def __init__(
-        self, iterator: Iterator[TIter], async_iterator: AsyncIterator[TIter]
-    ) -> None:
-        self._iterator = iterator
-        self._async_iterator = async_iterator
-
-    def __next__(self) -> TIter:
-        return self._iterator.__next__()
-
-    async def __anext__(self) -> Any:
-        return await self._async_iterator.__anext__()
 
 
 T = TypeVar("T", bound=StripeObject)
@@ -141,8 +125,8 @@ class ListObject(StripeObject, Generic[T]):
     def __reversed__(self) -> Iterator[T]:  # pyright: ignore (see above)
         return getattr(self, "data", []).__reversed__()
 
-    def auto_paging_iter(self) -> SyncAsyncIterator[T]:
-        return SyncAsyncIterator(
+    def auto_paging_iter(self) -> AnyIterator[T]:
+        return AnyIterator(
             self._auto_paging_iter(),
             self._auto_paging_iter_async(),
         )
--- a/stripe/_list_object.py
+++ b/stripe/_list_object.py
@@ -21,6 +21,24 @@ from stripe._request_options import RequestOptions, extract_options_from_dict
 
 from urllib.parse import quote_plus
 
+
+TIter = TypeVar("TIter", bound=StripeObject)
+
+
+class SyncAsyncIterator(Iterator[TIter], AsyncIterator[TIter]):
+    def __init__(
+        self, iterator: Iterator[TIter], async_iterator: AsyncIterator[TIter]
+    ) -> None:
+        self._iterator = iterator
+        self._async_iterator = async_iterator
+
+    def __next__(self) -> TIter:
+        return self._iterator.__next__()
+
+    async def __anext__(self) -> Any:
+        return await self._async_iterator.__anext__()
+
+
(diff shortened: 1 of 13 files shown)


anniel-stripe:
> ```python
>  # This succeeds but makes synchronous calls
>  print(", ".join([cus.id for cus in stripe.Customer.list_async().auto_paging_iter()]))
>  ```

Do we want this to succeed? If the user wanted to use all async methods, this seems like something they could easily miss and unintentionally introduce synchronous calls.

richardm-stripe:
@anniel-stripe 
> Do we want this to succeed? If the user wanted to use all async methods, this seems like something they could easily miss and unintentionally introduce synchronous calls.

I think ideally, no. But the only way to get this to fail is via an approach like https://github.com/stripe/stripe-python/compare/richardm-async-list-object?expand=1 which has a lot of downsides, basically you have to have 2 separate versions of ListObject (which is kind of at odds with the design decision we've made *not* to have mirror versions of resources), and you also have to flow a `prefer_async_versions` property through the entire request-making infrastructure in order to persist the knowledge of whether a particular list_object was obtained through the result of an asynchronous request or not.

So instead of that, for users who want to make sure that they've eliminated all synchronous requests, I propose a bigger hammer
```
stripe.default_http_client = stripe.HTTPXClient(disable_sync=True)
```
this unfortunately won't help the typechecker find synchronous method calls but it could cause runtime errors when they run tests.

anniel-stripe:
Ahh, I didn't parse the follow-up about `stripe.default_http_client = stripe.HTTPXClient(disable_sync=True)`.

So if I'm understanding correctly:
```
# No migration, all synchronous calls
print(", ".join([cus.id for cus in stripe.Customer.list().auto_paging_iter()]))
# 1 asynchronous call (through `list_async` method) + synchronous pagination calls
print(", ".join([cus.id for cus in (await stripe.Customer.list_async()).auto_paging_iter()]))
# 1 synchronous call (through the `list` method) + asynchronous pagination calls
print(", ".join([cus.id async for cus in stripe.Customer.list().auto_paging_iter()]))
# Correct migration, all asynchronous calls
print(", ".join([cus.id async for cus in (await stripe.Customer.list()).auto_paging_iter()]))
```

Overall, I do think this change would give a better experience!

stripe/_list_object.py
```diff
@@ -123,7 +141,13 @@ def __len__(self) -> int:
     def __reversed__(self) -> Iterator[T]:  # pyright: ignore (see above)
         return getattr(self, "data", []).__reversed__()
 
-    def auto_paging_iter(self) -> Iterator[T]:
+    def auto_paging_iter(self) -> SyncAsyncIterator[T]:
+        return SyncAsyncIterator(
+            self._auto_paging_iter(),
+            self._auto_paging_iter_async(),
+        )
```
anniel-stripe, line 148: nit: might be worth adding a comment here or above on the `SyncAsyncIterator` class explaining how this supports multiple syntaxes

anniel-stripe (approved):

anniel-stripe (approved):

PR #1248, opened 2024-02-23
Beta: disable HTTPX connection pooling on the default client

richardm-stripe:
A user [reported some friction](https://github.com/stripe/stripe-python/issues/327#issuecomment-1939530867) using `async` methods with `unittest` which tries to run each test in an isolated event loop.

This is because HTTPX attempts to re-use connections by default, and re-using a connection across two different event loops triggers an error.

We can avoid the error by disabling connection pooling. This causes requests to succeed in more circumstances, at the expense of making requests more wasteful and hurting performance. Users who wish to avoid this should initialize their own http client, e.g. `stripe.default_http_client = stripe.HTTPXClient()`

### DIFF
--- a/stripe/_http_client.py
+++ b/stripe/_http_client.py
@@ -120,11 +120,9 @@ def new_default_http_client(*args: Any, **kwargs: Any) -> "HTTPClient":
 
 def new_http_client_async_fallback(*args: Any, **kwargs: Any) -> "HTTPClient":
     if httpx:
-        impl = HTTPXClient
+        return HTTPXClient(*args, **kwargs, connection_pooling=False)
     else:
-        impl = NoImportFoundAsyncClient
-
-    return impl(*args, **kwargs)
+        return NoImportFoundAsyncClient(*args, **kwargs)
 
 
 class HTTPClient(object):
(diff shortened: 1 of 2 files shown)


pakrym-stripe:
Are we sacrificing real-world performance to fix unit tests?

richardm-stripe:
In a sense? I don't consider our own test cases to be a part of the "real world", but I do think that users writing tests is an important real world use case for us. The tradeoff here is to reduce friction for (some) users writing tests, at the expense of requiring more work from users who want to minimize resource usage (and changing the default to be more wasteful for users who haven't thought about either).

HTTPX makes a similar design decision itself, i.e. if you use the globals and go `httpx.request(...)` it won't use connection pooling, it only uses connection pooling when you `client = httpx.Client(); client.request(...)`.

richardm-stripe:
Maybe we can do something similar to how we handle connection pooling for `Requests` though, where we make a thread-local connection pool.
https://github.com/stripe/stripe-python/blob/341b23fe2a620fd2d1c95a76ee70851c23f7feda/stripe/_http_client.py#L478-L480

Maybe there is some concept of a "event-loop-local" variable.

richardm-stripe:
Oh interesting, I just read https://github.com/encode/httpx/issues/2058#issuecomment-1960608385 where @Zac-HD is recommending that httpx do that very thing upstream.

richardm-stripe:
I'm going to go ahead and close unmerged, I think it's premature & so far it looks like the user having the event loop troubles is happy to work around it. There's no real reason we couldn't address this later.

PR #1253, opened 2024-02-23
Beta: Support 'allow_sync_methods=False' in HTTPXClient

richardm-stripe:
## Summary
Allows initializing HTTPXClient so that synchronous method calls raise an exception.

## Motivation
I expect this to be useful for users who want to completely migrate their Stripe integrations to use async control flow. It's easy to miss migrating a method or two, especially if it's a pagination request, and then you have blocking io that you didn't intend.

Brought this up https://github.com/stripe/stripe-python/issues/327#issuecomment-1960571444 and at least one user has indicated it would be useful.

It's definitely not impossible for users to subclass HTTPXClient and implement this themselves so maybe this is feature bloat, but I think eventually suggesting this strategy in our README will be helpful, and it addresses some of the migration experience agonies that motivated https://github.com/stripe/stripe-python/pull/1247.

### DIFF
--- a/stripe/_http_client.py
+++ b/stripe/_http_client.py
@@ -1215,7 +1215,7 @@ class HTTPXClient(HTTPClient):
     def __init__(
         self,
         timeout: Optional[Union[float, "HTTPXTimeout"]] = 80,
-        allow_sync_methods=True,
+        allow_sync_methods=False,
         **kwargs
     ):
         super(HTTPXClient, self).__init__(**kwargs)
@@ -1273,7 +1273,7 @@ class HTTPXClient(HTTPClient):
     ) -> Tuple[bytes, int, Mapping[str, str]]:
         if self._client is None:
             raise RuntimeError(
-                "HTTPXClient was initialized with allow_sync_methods=False, "
+                "Stripe: HTTPXClient was initialized with allow_sync_methods=False, "
                 "so it cannot be used for synchronous requests."
             )
         args, kwargs = self._get_request_args_kwargs(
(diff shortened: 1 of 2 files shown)


stripe/_http_client.py
```diff
@@ -1209,8 +1210,13 @@ def close(self):
 class HTTPXClient(HTTPClient):
     name = "httpx"
 
+    _client: Optional["HTTPXClientType"]
+
     def __init__(
-        self, timeout: Optional[Union[float, "HTTPXTimeout"]] = 80, **kwargs
+        self,
+        timeout: Optional[Union[float, "HTTPXTimeout"]] = 80,
+        allow_sync_methods=True,
```
anniel-stripe, line 1218: Should this be the default? It seems like unintentionally introducing synchronous methods (especially via a pagination request) is easier to do, so maybe we should guard against that?

anniel-stripe (approved):

stripe/_http_client.py
richardm-stripe: Hmm, that does make sense! It's also less of a trapdoor, if people report getting confused/inconvenienced by sync methods being disabled, it's non-breaking to change this default from False to True but not the other way around.

richardm-stripe: (Will change)

pakrym-stripe: Are we sure? Sync is a fully supported feature for HTTPX. Maybe customers want to use https for other reasons than async methods.

richardm-stripe: That's a fair point. We never got a feature request for HTTPX outside of async support context, though, and I see "httpx-but-not-async" as a fairly niche use case. 

I am willing to accept requiring users who `stripe.default_http_client = stripe.HTTPXClient()`and **do** intend to use sync methods to apply an easy-to-discover remedy, as a fair price for preventing users who `stripe.default_http_client = stripe.HTTPXClient()` and **don't** intend to use sync methods from making a hard-to-discover mistake. Especially since this decision is the easier to reserve.

PR #1254, opened 2024-02-24
Beta: add AIOHTTP http client

richardm-stripe:
Like it says on the tin. 
Things to note: aiohttp only supports asyncio, not trio. Otherwise this is not too different from the HTTPXClient logic.

### DIFF
--- a/stripe/_http_client.py
+++ b/stripe/_http_client.py
@@ -6,7 +6,11 @@ import time
 import random
 import threading
 import json
+import asyncio
+import ssl
 from http.client import HTTPResponse
+
+from aiohttp import TCPConnector
 
 # Used for global variables
 import stripe  # noqa: IMP101
@@ -60,6 +64,13 @@ try:
 except ImportError:
     httpx = None
     anyio = None
+
+try:
+    import aiohttp
+    from aiohttp import ClientTimeout as AIOHTTPTimeout
+    from aiohttp import StreamReader as AIOHTTPStreamReader
+except ImportError:
+    aiohttp = None
 
 try:
     import requests
@@ -121,6 +132,8 @@ def new_default_http_client(*args: Any, **kwargs: Any) -> "HTTPClient":
 def new_http_client_async_fallback(*args: Any, **kwargs: Any) -> "HTTPClient":
     if httpx:
         impl = HTTPXClient
+    elif aiohttp:
+        impl = AIOHTTPClient
     else:
         impl = NoImportFoundAsyncClient
 
@@ -1261,7 +1274,6 @@ class HTTPXClient(HTTPClient):
         url: str,
         headers: Mapping[str, str],
         post_data=None,
-        timeout: float = 80.0,
     ) -> Tuple[bytes, int, Mapping[str, str]]:
         args, kwargs = self._get_request_args_kwargs(
             method, url, headers, post_data
@@ -1282,7 +1294,6 @@ class HTTPXClient(HTTPClient):
         url: str,
         headers: Mapping[str, str],
         post_data=None,
-        timeout: float = 80.0,
     ) -> Tuple[bytes, int, Mapping[str, str]]:
         args, kwargs = self._get_request_args_kwargs(
             method, url, headers, post_data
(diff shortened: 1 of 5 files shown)


stripe/_http_client.py
```diff
@@ -1353,6 +1364,113 @@ async def close_async(self):
...
+        self, timeout: Optional[Union[float, "AIOHTTPTimeout"]] = 80, **kwargs
+    ):
+        super(AIOHTTPClient, self).__init__(**kwargs)
+
+        if aiohttp is None:
+            raise ImportError(
+                "Unexpected: tried to initialize AIOHTTPClient but the aiohttp module is not present."
+            )
+
+        self.httpx = httpx
```
pakrym-stripe, line 1380: ?

stripe/_http_client.py
```diff
@@ -1353,6 +1364,113 @@ async def close_async(self):
...
+            )
+            kwargs["connector"] = TCPConnector(ssl=ssl_context)
+
+        self._session = aiohttp.ClientSession(**kwargs)
+        self._timeout = timeout
+
+    def sleep_async(self, secs):
+        return asyncio.sleep(secs)
+
+    def request(self) -> Tuple[bytes, int, Mapping[str, str]]:
```
pakrym-stripe, line 1395: Do we want to implement sync fallback?

stripe/_http_client.py
```diff
@@ -1353,6 +1364,113 @@ async def close_async(self):
...
+            "AIOHTTPClient does not support synchronous requests."
+        )
+
+    def _get_request_args_kwargs(
+        self, method: str, url: str, headers: Mapping[str, str], post_data
+    ):
+        args = (method, url)
+        kwargs = {}
+        if self._proxy:
+            if self._proxy["http"] != self._proxy["https"]:
```
pakrym-stripe, line 1406: is this comparing by-value?

tests/test_http_client.py
```diff
@@ -1505,3 +1524,418 @@ def test_timeout(self):
...
+
+            class Result:
+                def __init__(self):
+                    self.content = Content()
+                    self.status = code
+                    self.headers = {}
+
+            result = Result()
+
+            request_mock.ClientSession().request = AsyncMock(
```
pakrym-stripe, line 1557: Is it expected that while testing HTTP client we are not making actual requests using the HTTP client?

pakrym-stripe: I was under impression that we are starting an in-process http server.

tests/test_integration.py
```diff
@@ -351,25 +372,29 @@ class MockServerRequestHandler(MyTestHandler):
...
-        )
         # If we set HTTPX's generic timeout the test is flaky (sometimes it's a ReadTimeout, sometimes its a ConnectTimeout)
         # so we set only the read timeout specifically.
-        stripe.default_http_client._async_fallback_client._timeout = (
-            httpx.Timeout(None, read=0.01)
-        )
+        hc = stripe.default_http_client
+
+        expected_message = ""
+        if isinstance(hc, stripe.HTTPXClient):
```
pakrym-stripe, line 390: Should we unify timeout errors? Makes for a leaky abstraction when to switch HTTP client customers need to rewrite their exception handling.

pakrym-stripe (changes requested):
Few questions.

stripe/_http_client.py
richardm-stripe: I don't think so. Async fallback is mostly just so the default client can use it. And initializing a sync client like Requests and setting AIOHTTP as the async fallback achieves the same effect, except for the case where you want to use AIOHTTP for async requests but HTTPX for sync requests. If the user wants this, this should implement their own HTTPClient.

stripe/_http_client.py
richardm-stripe: What are you trying to say? Are you worried these won't be strings?

tests/test_http_client.py
richardm-stripe: Yes `test_integration` makes end-to-end requests and I have expanded it in this PR. `test_http_client` mocks the underlying http client implementation but exercises the code in the http client "wrapper".

richardm-stripe: My take is that the integration tests are way more useful, but I've cargo-culted these and extended them to the new http clients, mostly begrudgingly, 80% of the time when these fail it's something dumb with the mock instead of a real problem with the implementation, but I wouldn't say they're completely useless.

tests/test_integration.py
richardm-stripe: What do you mean by "unify"? The existing logic is that timeout errors will throw a `stripe.APIConnectionError` but have messages derived from the original exception thrown by the underlying client. There's logic in this test that makes assertions specific to text inside the exception because we are testing timeout behavior and we want to be careful that the underlying exception represents a timeout and not some other sort of failure, but I wouldn't expect users to parse `message` in their own exception handling code.

stripe/_http_client.py
richardm-stripe: Copy pasta removed in [Remove dead code](https://github.com/stripe/stripe-python/pull/1254/commits/385f1f197ee4acf65bd33a868f444506294ad1b4)

richardm-stripe:
All feedback has a response

stripe/_http_client.py
pakrym-stripe: For some reason I though these are not strings.

pakrym-stripe (approved):

PR #1256, opened 2024-02-27
Beta: document async support in README.md

richardm-stripe:
We're going to tweet about the async support and mention it in the Stripe developer digest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,7 @@ deserialized_resp = stripe.deserialize(response)
 
 ### Async
 
-Asynchronous versions of most request-making methods are available by suffixing the method name
+Asynchronous versions of request-making methods are available by suffixing the method name
 with `_async`.
 
 ```python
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -277,6 +277,54 @@ response = stripe.raw_request(
 deserialized_resp = stripe.deserialize(response)
 ```
 
+### Async
+
+Asynchronous versions of most request-making methods are available by suffixing the method name
```
pakrym-stripe, line 282: why most? should we clarify what's missing and whether customers should expect it in the future?

richardm-stripe: Note about `.save` a little below

pakrym-stripe (approved):

README.md
richardm-stripe, line 282:
```suggestion
Asynchronous versions of request-making methods are available by suffixing the method name
```

PR #1259, opened 2024-02-27
Add helper to add beta version

helenye-stripe:
along with https://github.com/stripe/stripe-python/pull/1260

Curious whether you think I should make a new test file as well.


richardm-stripe:
This works for me! The APIRequestor test is fine and it feels right this test would be next to the test for setAppInfo. Would also be open to a `stripe_test.py` or something, but it would get irritating to have that only in beta but not in master.

richardm-stripe (approved):
LGTM but needs format

richardm-stripe (approved):

PR #1266, opened 2024-02-29
Beta: record usage of async interface

richardm-stripe:


stripe/_http_client.py
```diff
@@ -502,7 +502,7 @@ async def _request_with_retries_internal_async(
         *,
         _usage: Optional[List[str]] = None
     ) -> Tuple[Any, int, Mapping[str, str]]:
-        self._add_telemetry_header(headers)
```
richardm-stripe, line 505: this was a bug. This meant that telemetry wasn't being recorded at all for async requests

anniel-stripe (approved):

stripe/_http_client.py
anniel-stripe: Nice catch! Looks like there's some tests that need updating after fixing this bug.

richardm-stripe: Fixed in [Fix tests](https://github.com/stripe/stripe-python/pull/1266/commits/45f16ebc99b30b9d07e453601ef5d410c1c845ba)

tests/test_integration.py
```diff
@@ -329,7 +330,7 @@ def do_request(self, req_num):
         assert "last_request_metrics" in telemetry
 
         usage = telemetry["last_request_metrics"]["usage"]
-        assert usage == ["stripe_client"]
+        assert usage == ["stripe_client", "async"]
```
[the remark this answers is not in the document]
richardm-stripe: I'm ok with `test_stripe_client_py`, integration tests are a bit slower and don't need to have complete coverage.

PR #1271, opened 2024-03-14
Support AIOHTTPClient init without running event loop

OldSneerJaw:
Lazily instantiates and caches an aiohttp `ClientSession` at the points in time that it is actually used so that `AIOHTTPClient` can be initialized without a running event loop (e.g. when setting `stripe.default_http_client = stripe.AIOHTTPClient()` at the top level of a module).

Related to issue #327.

### DIFF
--- a/stripe/_http_client.py
+++ b/stripe/_http_client.py
@@ -1399,6 +1399,8 @@ class AIOHTTPClient(HTTPClient):
 
     @cached_property
     def _session(self):
+        assert aiohttp is not None
+
         kwargs = {}
         if self._verify_ssl_certs:
             ssl_context = ssl.create_default_context(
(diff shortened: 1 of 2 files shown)


OldSneerJaw:
I fixed the typing errors and removed the usage of the `cached_property` decorator since it's not available in Python 3.7.

OldSneerJaw:
@richardm-stripe: Bump.

richardm-stripe:
This looks good to me! It will be in Thursday's beta release.

Sorry for the silence on this. Thank you for teaching me about about `@cached_property` (too bad we can't use it yet 😞  ) and handling the backwards compatibility/typing errors.

richardm-stripe (approved):

PR #1294, opened 2024-04-04
CHANGELOG.md: slight format fix

richardm-stripe:


richardm-stripe:
Oh, didn't realize that #1292 had not been merged, I'll push a commit over there instead.

PR #1297, opened 2024-04-05
Use stdlib AsyncMock when available

mgorny:
As indicated in the `asyncmock` project description:

> Mock 4.0+ (included within Python 3.8+) now includes an awaitable mock
> `mock.AsyncMock`. This is recommended for new projects.

Appropriately, update the test code to use `unittest.mock.AsyncMock` in Python 3.8+, and update the test requirements to pull in `mock >= 4.0` (rather than the obsolete `asyncmock` package) for older Python versions.


richardm-stripe:
Hello @mgorny. Thank you for the contribution.

I agree with this change. I think when the time comes and we've dropped support for python < 3.8 having the code like this will make it easier to make the appropriate update.

Can you sign the CLA in the github comment from @CLAassistant?

richardm-stripe (approved):

mgorny:
Thanks!

PR #1304, opened 2024-04-16
Stripe deprecate param util for adding a decoractor

prathmesh-stripe:
Adds decorator for warning the user when they use a deprecated parameter. Works at runtime.

### DIFF
--- a/stripe/_util.py
+++ b/stripe/_util.py
@@ -464,4 +464,5 @@ def stripe_deprecate_parameter(key):
             return original_function(params=params, *args, **kwargs)
 
         return stripe_deprecate_param
+
     return stripe_deprecate_param_decorator


stripe/_util.py
```diff
@@ -433,3 +433,33 @@ def _wrapper(*args, **kwargs):
                 return class_method(*args, **kwargs)
 
         return _wrapper
+
+
+def stripe_deprecate_parameter(key):
+    def stripe_deprecate_param_decorator(original_function):
+        def stripe_deprecate_param_inner(params: Mapping[str, Any], parts: List[str]):
```
richardm-stripe, line 440: nit: you can unnest `stripe_deprecate_param_inner` and move it to the top level, since it only uses its own arguments and doesn't need `original_function` to be in scope.

richardm-stripe (approved):
LGTM

ramya-stripe:
@prathmesh-stripe I'll skip this from the changelog because there is no net change to user experience until we start using the decorator

PR #1325, opened 2024-05-09
Removed deprecated parameter util

prathmesh-stripe:
We surveyed and found out the developers rarely read warning logs. They might not find this as helpful as IDE warnings. Hence rolling it back. 

No usages in code. Does this need a changelog?


anniel-stripe (approved):
LGTM!

> Does this need a changelog?

The title is descriptive enough IMO, and that'll get picked up in the Changelog.

PR #1326, opened 2024-05-09
Add lint rule to ensure `async` functions are named `X_async`

xavdid-stripe:
This came from a discussion in https://github.com/stripe/stripe-python/pull/1165#discussion_r1427214415

I exempted 1 existing function (`StripeStreamResponseAsync.read`). We could rename it, but I don't have a good sense of how breaking that is, so it's not urgent. 

This should catch future things though!

### DIFF
--- a/stripe/_stripe_response.py
+++ b/stripe/_stripe_response.py
@@ -61,5 +61,5 @@ class StripeStreamResponseAsync(StripeResponseBase):
     def stream(self) -> AsyncIterable[bytes]:
         return self._stream
 
-    async def read(self) -> bytes:
+    async def read(self) -> bytes:  # noqa: ASY100
         return b"".join([chunk async for chunk in self._stream])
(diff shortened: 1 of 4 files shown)


stripe/_stripe_response.py
```diff
@@ -61,5 +61,5 @@ def __init__(
     def stream(self) -> AsyncIterable[bytes]:
         return self._stream
 
-    async def read(self) -> bytes:
+    async def read(self) -> bytes:  # noqa: ASY100
```
anniel-stripe, line 64: nit: consider filing a ticket and/or adding a TODO (MAJOR) comment to fix this naming in the next major.

anniel-stripe (approved):

stripe/_stripe_response.py
xavdid-stripe: sure! added a `TODO (MAJOR)`.

anniel-stripe (approved):

PR #1345, opened 2024-06-12
adjust default # of retries and statuses to retry on

xavdid-stripe:


xavdid-stripe:
Shelving for now, but will revisit

PR #1365, opened 2024-07-19
Do not share single RequestorOptions between APIRequestor instances

akx:
Given `RequestorOptions` is internally mutable, it's probably not the intent that a single `RequestorOptions` is shared between all `APIRequestor`s that aren't explicitly passed one.


akx:
cc @xavdid-stripe :)

xavdid-stripe (approved):

Issue #43, opened 2013-08-25
AssertionError: No api proxy found for service "urlfetch"


jgirdner:
We are using Google Compute engine on our project and we are also using Google SQL for our DB and Google Cloud Storage.  We have to have Google App Engine installed to be able to connect to these other services.  I ran into this problem when trying to run the init_plans.py command.

AssertionError: No api proxy found for service "urlfetch"

Here is the traceback.

Traceback (most recent call last):
  File "/home/code/virtualenvs/***_/local/lib/python2.7/site-packages/django/core/management/base.py", line 222, in run_from_argv
    self.execute(_args, **options.**dict**)
  File "/home/code/virtualenvs/****/local/lib/python2.7/site-packages/django/core/management/base.py", line 255, in execute
    output = self.handle(_args, *_options)
  File "/home/code/virtualenvs/****/local/lib/python2.7/site-packages/payments/management/commands/init_customers.py", line 14, in handle
    Customer.create(user=user)
  File "/home/code/virtualenvs/****/local/lib/python2.7/site-packages/payments/models.py", line 378, in create
    email=user.email
  File "/home/code/virtualenvs/****/local/lib/python2.7/site-packages/stripe/**init**.py", line 703, in create
    response, api_key = requestor.request('post', url, params)
  File "/home/code/virtualenvs/****/local/lib/python2.7/site-packages/stripe/**init**.py", line 233, in request
    rbody, rcode, my_api_key = self.request_raw(meth, url, params)
  File "/home/code/virtualenvs/****/local/lib/python2.7/site-packages/stripe/**init**.py", line 292, in request_raw
    rbody, rcode = self.urlfetch_request(meth, abs_url, headers, params)
  File "/home/code/virtualenvs/****/local/lib/python2.7/site-packages/stripe/**init**.py", line 432, in urlfetch_request
    result = urlfetch.fetch(**args)
  File "/home/code/virtualenvs/***_/local/lib/python2.7/site-packages/google/appengine/api/urlfetch.py", line 267, in fetch
    rpc = create_rpc(deadline=deadline)
  File "/home/code/virtualenvs/**__/local/lib/python2.7/site-packages/google/appengine/api/urlfetch.py", line 223, in create_rpc
    return apiproxy_stub_map.UserRPC('urlfetch', deadline, callback)
  File "/home/code/virtualenvs/_**_/local/lib/python2.7/site-packages/google/appengine/api/apiproxy_stub_map.py", line 413, in **init**
[...]

PR #1378, opened 2024-08-16
Prioritize requests as preferred HTTP library

nanewalt:
Prioritize the `requests` library over Google app engine's `urlfetch` as http provider.

`_http_client.py` [declares](https://github.com/stripe/stripe-python/blob/master/stripe/_http_client.py#L42-L45) that "Requests is the preferred HTTP library"; however, `urlfetch` is chosen before `requests` when creating a default http client.

Additionally, the current ordering throws the error `No api proxy found for service "urlfetch"` when `appengine-python-standard` is installed. It appears that the recommended [solution](https://github.com/stripe/stripe-python/issues/43) is to uninstall the `appengine-python-standard` package.

### DIFF
--- a/stripe/_http_client.py
+++ b/stripe/_http_client.py
@@ -116,10 +116,10 @@ def _now_ms():
 
 
 def new_default_http_client(*args: Any, **kwargs: Any) -> "HTTPClient":
-    if urlfetch:
-        impl = UrlFetchClient
-    elif requests:
+    if requests:
         impl = RequestsClient
+    elif urlfetch:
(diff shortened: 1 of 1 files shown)


richardm-stripe:
Hi @nanewalt thanks for taking the initiative and opening this. 

In this case I think we have to change the comment to reflect reality (#1381) rather than change reality to reflect the comment. Changing the precedence would cause everybody using urlfetch on Google App Engine to be suddenly switched to using Requests when upgrading the library, which seems like it could be pretty disruptive.

PR #1390, opened 2024-09-06
Moved raw_request functions from stripe module to StripeClient class

jar-stripe:
Context:
We are changing our SDKs to move "raw request" functions into the `StripeClient` class.


stripe/_stripe_client.py
```diff
@@ -280,3 +286,21 @@ def construct_event(
...
+    def raw_request(self, method_: str, url_: str, **params):
+        return _raw_request(method_, url_, **params)
+
+    async def raw_request_async(self, method_: str, url_: str, **params):
+        return await _raw_request_async(method_, url_, **params)
+
+    def deserialize(
+        self,
+        resp: Union[StripeResponse, Dict[str, Any]],
+        api_key: Optional[str] = None,
```
jar-stripe, line 299: one question I had here: since we have access to the API key as part of StripeClient, do we still want to take api_key as an argument?  `deserialize` in other languages (i.e. java) don't; here, it's used to construct a requestor that is used in object construction during deserialization (I can assume so that some objects can make subsequent API calls?)

tests/test_integration.py
```diff
@@ -361,10 +363,11 @@ class MockServerRequestHandler(MyTestHandler):
 
         stripe.api_base = "http://localhost:%s" % self.mock_server_port
 
-        resp = await stripe.raw_request_async(
+        client = StripeClient("sk_test")
```
ramya-stripe, line 366: I think the general pattern in tests is to create the client once and re-use it across the different test cases. Can you confirm that by looking at any other test file and make the changes appropriately?

stripe/_preview.py
```diff
@@ -1,4 +1,5 @@
-from stripe import raw_request, raw_request_async
+from stripe._raw_request import raw_request as raw_request
+from stripe._raw_request import raw_request_async as raw_request_async
```
ramya-stripe, line 2: We also do not want the preview layer anymore - so you can delete this file and the corresponding test file

stripe/_stripe_client.py
```diff
@@ -146,7 +152,7 @@ def __init__(
             **base_addresses,
         }
 
-        requestor_options = RequestorOptions(
+        self._requestor_options = RequestorOptions(
```
ramya-stripe, line 155: What was the reason for this change?

jar-stripe: That was so I could reference those api_key, stripe_version, and stripe_account down in deserialize without having to add several additional member variables.

tests/test_integration.py
jar-stripe: It seems to be a mixed bag: `test_generated_examples.py` and `test_integration.py` create a new client per test, `test_oauth.py` uses an in-class fixture, and a few other files (`test_webhook.py`, `test_file_upload.py`) use a fixture defined in `conftest.py`.  Do we have a preferred pattern?

stripe/_preview.py
jar-stripe: deleted the _preview.py and test_preview.py files and removed the imports from __init__.py

stripe/_stripe_client.py
```diff
@@ -19,10 +19,16 @@
...
+from stripe._stripe_response import StripeResponse
 from stripe._webhook import Webhook, WebhookSignature
 from stripe._event import Event
 
-from typing import Optional, Union, cast
+from typing import Any, Dict, Optional, Union, cast
+
+from stripe._raw_request import raw_request as _raw_request  # noqa
+from stripe._raw_request import raw_request_async as _raw_request_async  # noqa
+from stripe._raw_request import deserialize as _deserialize  # noqa
```
ramya-stripe, line 31: The above 3 methods use the global api requestor. I believe one of the key differentiators of the StripeClient model is that there is a dedicated api requestor per StripeClient that gets used instead of the global one - that is how we get to set different api-key/account per StripeClient. 

Therefore, I think we need a bigger refactor.

tests/test_integration.py
ramya-stripe: @xavdid-stripe Any preferences?

ramya-stripe: am ok with doing that at a later time

jar-stripe: Since I was adding StripeClient lines as part of this PR, if that's the ultimate direction we want to go I'm inclined to use the fixture here.

stripe/_stripe_client.py
jar-stripe: I think this is done.  I removed _raw_request.py and merged those methods directly into StripeClient so I could use its _requestor member, and updated the tests to match.

tests/test_integration.py
jar-stripe: quick update: I made the change to use the fixtures in test_raw_request, but for test_integration many of the tests override the default options in some way, so I kept the per-test client pattern and made sure the constructor args were correctly set up.

stripe/_stripe_client.py
prathmesh-stripe: Where do we use this in deserialize method?

stripe/_stripe_client.py
```diff
@@ -280,3 +284,62 @@ def construct_event(
...
+    def deserialize(
+        self,
+        resp: Union[StripeResponse, Dict[str, Any]],
+        params: Optional[Dict[str, Any]] = None,
+    ) -> StripeObject:
+        return _convert_to_stripe_object(
+            resp=resp,
+            params=params,
+            requestor=self._requestor,
+            api_mode="V1",
```
prathmesh-stripe, line 344: This needs to be compatible with v2. What do you think about passing this as a param?

stripe/_stripe_client.py
```diff
@@ -280,3 +284,62 @@ def construct_event(
...
+            api_mode=api_mode,
+            usage=["raw_request"],
+        )
+
+        return self._requestor._interpret_response(rbody, rcode, rheaders)
+
+    async def raw_request_async(self, method_: str, url_: str, **params):
+        params = params.copy()
+        options, params = extract_options_from_dict(params)
+        api_mode = params.pop("api_mode", None)
```
prathmesh-stripe, line 320: We won't need to pass api_mode for raw_request because this will be calculated automatically from the path by a util method in api_requestor. 

I think that was not added on the [request_raw method](https://github.com/stripe/stripe-python-next/blob/015a4930c0b90859cc20ed9416fa6e9621b639b9/stripe/_api_requestor.py#L670) but exists on the [request method](https://github.com/stripe/stripe-python-next/blob/015a4930c0b90859cc20ed9416fa6e9621b639b9/stripe/_api_requestor.py#L185).

prathmesh-stripe (changes requested):

stripe/_stripe_client.py
jar-stripe: good catch!  this is correct; will update this shortly

jar-stripe: resolved in https://github.com/stripe/stripe-python/pull/1399

stripe/_stripe_client.py
jar-stripe: resolved in https://github.com/stripe/stripe-python/pull/1399

PR #1397, opened 2024-09-24
Adding support for the new Usage Billing APIs

prathmesh-stripe:
Adding support for the new Usage Billing APIs


stripe/_api_requestor.py
```diff
@@ -316,15 +323,60 @@ def handle_error_response(self, rbody, rcode, resp, rheaders) -> NoReturn:
...
+                resp,
+                rheaders,
+                code,
+            )
+        # switchCases: The beginning of the section generated from our OpenAPI spec
+        elif type == "temporary_session_expired":
+            return error.TemporarySessionExpiredError(**error_args)
+        # switchCases: The end of the section generated from our OpenAPI spec
+
+        return self.specific_v1_api_error(
```
ramya-stripe, line 373: Shouldnt we look for code == 'invalid_fields' here before falling back on v1 errors?

PR #1399, opened 2024-09-25
Moved raw_request functions from stripe module to StripeClient class

jar-stripe:
Follow up to https://github.com/stripe/stripe-python/pull/1390 which was accidentally merged and then reverted.


stripe/_stripe_client.py
```diff
@@ -298,3 +302,62 @@ def parse_snapshot_event(
...
+    def deserialize(
+        self,
+        resp: Union[StripeResponse, Dict[str, Any]],
+        params: Optional[Dict[str, Any]] = None,
+    ) -> StripeObject:
+        return _convert_to_stripe_object(
+            resp=resp,
+            params=params,
+            requestor=self._requestor,
+            api_mode="V1",
```
prathmesh-stripe, line 362: Aren't you going to pass api_mode to this method?

jar-stripe: it would help if I pushed the commit that fixes this.  really I was just making sure you're paying attention 😅

PR #1401, opened 2024-09-27
Fix type hint for SearchResultObject.data

maxhollmann:
I think this is correct, a comment in the same file even mentions that data is of type `list[T]`.


xavdid-stripe (approved):

xavdid-stripe:
Thank you for this! Apologies for the extremely slow motion on our side- sometimes things get lost in the shuffle.

PR #1403, opened 2024-09-27
Fix v2 event object name

jar-stripe:
### Why?
The SDK uses object strings in the API response to decide what class to parse a resource returned from the Stripe API into.  Recently, the object string for the V2 event base class changed from "event" to "v2.core.event".  We have to update the SDK to match.

### What?
- Changed "event" to "v2.core.event" in _util._convert_to_stripe_object

Verified using the `stripe_webhook_handler.py` example with events triggered from and forwarded by the Stripe CLI

### See Also
http://go/j/DEVSDK-2202

### DIFF
--- a/examples/stripe_webhook_handler.py
+++ b/examples/stripe_webhook_handler.py
@@ -24,7 +24,6 @@ def webhook():
         # Fetch the event data to understand the failure
         event = client.v2.core.events.retrieve(thin_event.id)
         if isinstance(event, V1BillingMeterErrorReportTriggeredEvent):
-            # CHECK: fetch_object is present and callable, returning a strongly-typed object (without casting)
             meter = event.fetch_related_object()
             meter_id = meter.id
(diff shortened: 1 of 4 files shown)


examples/stripe_webhook_handler.py
```diff
@@ -27,6 +27,7 @@ def webhook():
             # CHECK: fetch_object is present and callable, returning a strongly-typed object (without casting)
```
xavdid-stripe, line 27: not part of this PR, but this line probably shouldn't be in examples?

xavdid-stripe (approved):

prathmesh-stripe (approved):

prathmesh-stripe:
Needs a test to be fixed too.

examples/stripe_webhook_handler.py
jar-stripe: correct, will fix.  thanks!

PR #1405, opened 2024-10-01
Add Custom requests section to README

jar-stripe:
### Why?
The new release of the SDK adds methods to issue custom requests to support using the SDK to access pre-release or undocumented API endpoints.  This PR updates the README to document how this is used.

### What?
- added Custom requests section to readme with an example


README.md
```diff
@@ -264,6 +264,20 @@ If your beta feature requires a `Stripe-Version` header to be sent, set the `str
...
 ```
 
+### Custom requests
+
+If you would like to send a request to an undocumented API (for example you are in a private beta), or if you prefer to bypass the method definitions in the library and specify your request details directly, you can use the `raw_request` method on `StripeClient`.
+
+```python
+client = StripeClient("sk_test_...")
+response = client.raw_request(
+    "post", "/v1/beta_endpoint", param=123, stripe_version="2022-11-15; feature_beta=v3"
```
prathmesh-stripe, line 274:
```suggestion
    "post", "/v1/beta_endpoint", param=123, stripe_version="2022-11-15; feature_beta=v3"
```

Did you try this request?
param should be `params` and you might be missing a quote somewhere.

jar-stripe: ah good call.  I did not, I just copied it over from beta.  let me take a look.

jar-stripe: ok, I looked at it again, and params (inside raw_request) is a `**kwargs` which means it will collect any keyword args after the named arguments into a dictionary.  its an unfortunate coincidence that param and params are named similar, but this will insert a key 'param' inside 'params' with a value of '123' (along side stripe_version, etc).  i did however notice that deserialize needs an api_mode argument, so I'm fixing this now.

prathmesh-stripe (approved):

PR #1409, opened 2024-10-03
Add livemode to ThinEvent

helenye-stripe:
## Why?

This field is being returned from the API. It should be typed in our SDKs.

## Changelog

* Update the class for `ThinEvent` to include `livemode`


prathmesh-stripe (commented):
Can we add a test to assert livemode with both true and false values?

tests/test_v2_event.py
```diff
@@ -18,6 +18,7 @@ def v2_payload_no_data(self):
                 "id": "evt_234",
                 "object": "v2.core.event",
                 "type": "financial_account.balance.opened",
+                "livemode": "false",
```
prathmesh-stripe, line 21:
```suggestion
                "livemode": False,
```

Same in the other method. Maybe change that one to True

ramya-stripe (approved):

PR #1410, opened 2024-10-03
Merge from master to beta

ramya-stripe:
There have been more changes in the master branch since the last time we merged from it to beta. Doing on last merge from master before this week's beta releases


stripe/v2/_event.py
```diff
@@ -132,6 +136,7 @@ def __init__(self, payload: str) -> None:
         self.id = parsed["id"]
         self.type = parsed["type"]
         self.created = parsed["created"]
+        self.livemode = parsed.get("livemode")
```
jar-stripe, line 139: we're missing reason; is that a blocker for this release?

ramya-stripe: scroll scroll scroll a bit... gently down the page

jar-stripe (approved):
👍

Issue #1432, opened 2024-12-12
Backwards compatibility broken from stripe v7 -> v8 transition


eduard-grigoryan-ext-bark:
### Describe the bug

We have noticed a strange bug that when upgrading from v7 -> v8.
StripeError -> http_body becomes None instead of valid dictionary

### To Reproduce

Fill STRIPE_GB_SECRET_KEY with valid test/dev key and run the script. 
You will see that `http_body` is None, while `json_body` is valid dict. 
This is starting from v8. Before v8 both http_body and json_body contain identical data. 

```python
import stripe
import os
from dotenv import load_dotenv
import shortuuid
import json

# Load environment variables from .env file
load_dotenv()

# Configure Stripe with your test API key
stripe.api_key = os.getenv('STRIPE_GB_SECRET_KEY')

def test_stripe_charge():
    # Test parameters
    metadata = {}
    metadata['moto_charge'] = False
    params = {
        "amount": 2000,  # $20.00 in cents
        "currency": "USD",
        "customer": "cus_xxx",  # Replace with a test customer ID
        "description": "Test charge from script",
        "idempotency_key": shortuuid.uuid(),
        "metadata": metadata
    }

    try:
        print("Attempting to create charge...")
        stripe_charge = stripe.Charge.create(**params)
        print("Charge successful!")
        print(f"Charge ID: {stripe_charge.id}")
        print(f"Amount: ${stripe_charge.amount/100}")
        print(f"Status: {stripe_charge.status}")
        return stripe_charge
        
    except stripe.error.InvalidRequestError as e:
        # Invalid parameters were supplied to Stripe's API
        print(f"Invalid Request Error: {str(e)}")
        print(f"Invalid Request Error: {str(e.http_body)}")
        print(f"Invalid Request Error: {str(e.json_body)}")
        print(json.dumps(e.json_body, indent=2))
        

if __name__ == "__main__":
    test_stripe_charge()
```

### Expected behavior

Suddenly http_body instead of giving dict starts giving None. I think this is backward compatibility issue. 

### OS

macOS

### Language version

3.8

### Library version

stripe-python v8.0.0

### API version

Not sure

jar-stripe:
Hi @eduard-grigoryan-ext-bark, thank you for reporting this!  I think I see the issue, and I've added this to the backlog to get fixed in an upcoming SDK release.  Is this issue preventing you from upgrading your SDK version, and could you upgrade to the latest version (v11) if it were fixed in that version?

[...]

PR #1435, opened 2024-12-16
Fix StripeError http_body

jar-stripe:
### Why?
See https://github.com/stripe/stripe-python/issues/1432.  There was a bug introduced in v7.1 of the SDK where the unprocessed http body passed into a Stripe error was not correctly set into the `http_body` property. 

### What?
- changed StripeError constructor to properly set http_body when a string http body is passed in.

## Changelog
- Fixes an issue where `http_body` may be null in a Stripe error even when `json_body` is a valid dictionary.

### DIFF
--- a/stripe/_error.py
+++ b/stripe/_error.py
@@ -40,7 +40,6 @@ class StripeError(Exception):
                     )
             elif isinstance(http_body, str):
                 body = http_body
-                pass
 
         self._message = message
         self.http_body = body
--- a/stripe/_error.py
+++ b/stripe/_error.py
@@ -26,14 +26,21 @@ class StripeError(Exception):
         super(StripeError, self).__init__(message)
 
         body: Optional[str] = None
-        if http_body and hasattr(http_body, "decode"):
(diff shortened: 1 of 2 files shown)


stripe/_error.py
```diff
@@ -26,14 +26,21 @@ def __init__(
...
+                try:
+                    body = cast(bytes, http_body).decode("utf-8")
+                except BaseException:
+                    body = (
+                        "<Could not decode body as utf-8. "
+                        "Please report to support@stripe.com>"
+                    )
+            elif isinstance(http_body, str):
+                body = http_body
+                pass
```
xavdid-stripe, line 43: can remove

xavdid-stripe (commented):
looks good! have an extra line you should pull before merge, but approved otherwise

stripe/_error.py
jar-stripe: ah thanks for catching that.  removed!

xavdid-stripe (approved):

Issue #1437, opened 2024-12-19
v11.4.0 SSL certificate verify failed with async http client


e-carlin:
### Describe the bug

With version 11.4.0 I am getting `httpx.ConnectError: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1129)`. If I downgrade to 11.3.0 the error goes away. 

This seems to only happen when running under asyncio. 

### To Reproduce

```bash
pip install stripe==11.3.0
python <<'EOF'
import asyncio
import stripe
import stripe.checkout
def main():
    stripe.default_http_client = stripe.HTTPXClient()
    stripe.api_key = "YOUR_API_KEY"
    price_id = "YOUR_PRICE_ID"
    asyncio.run(
        stripe.checkout.Session.create_async(
            customer_email="test@test.com",
            ui_mode="embedded",
            line_items=[
                {
                    "price": price_id,
                    "quantity": 1,
                },
            ],
            mode="subscription",
            return_url=f"http://localhost:3000/payment-complete?session_id={{CHECKOUT_SESSION_ID}}",
        )
    )
main()
EOF
# >> No error
pip install stripe==11.4.0
python <<'EOF'
import asyncio
import stripe
import stripe.checkout
def main():
    stripe.default_http_client = stripe.HTTPXClient()
    stripe.api_key = "YOUR_API_KEY"
    price_id = "YOUR_PRICE_ID"
    asyncio.run(
        stripe.checkout.Session.create_async(
            customer_email="test@test.com",
            ui_mode="embedded",
            line_items=[
                {
                    "price": price_id,
                    "quantity": 1,
                },
            ],
            mode="subscription",
            return_url=f"http://localhost:3000/payment-complete?session_id={{CHECKOUT_SESSION_ID}}",
        )
    )
main()
EOF
# >> see error
```

### Expected behavior

I expect there to be no error using version 11.4.0 of the library.

### OS

macOS and Fedora

### Language version

Python 3.9.15 and Python 3.12.2

### Library version

stripe-python==11.4.0

### API version

2024-12-18.acacia

sileht:
This seems related to this commit: https://github.com/stripe/stripe-python/commit/12637de43efa1af5e4a2cd5948c94ac9b5ec11e7

jar-stripe:
[...]

PR #1438, opened 2024-12-19
fix SSL context with local path

xavdid-stripe:
### Why?

https://github.com/stripe/stripe-python/pull/1431 fixed a deprecation warning, but introduced a bug because we weren't creating the SSL context correctly. We have a path to a file, so we should use `cafile`. I mistakenly thought that kwarg was the actual contents of a file. This now lines up with [how `httpx` treated the string we were passing before](https://github.com/encode/httpx/blob/26d48e0634e6ee9cdc0533996db289ce4b430177/httpx/_config.py#L45-L54). 

Reported in https://github.com/stripe/stripe-python/issues/1437.

### What?

- use `cafile` kwarg instead of `capath` for `ssl.create_default_context`.

### Testing

This script fails without the fix and passes with it:

```py
import stripe

def main():
    stripe.default_http_client = stripe.HTTPXClient(allow_sync_methods=True)
    stripe.api_key = "sk_test_1234"
    print(stripe.Customer.list())
    
main()
```

### See Also

- https://github.com/stripe/stripe-python/issues/1437
- https://github.com/stripe/stripe-python/pull/1431

### DIFF
--- a/stripe/_http_client.py
+++ b/stripe/_http_client.py
@@ -1245,7 +1245,7 @@ class HTTPXClient(HTTPClient):
         kwargs = {}
         if self._verify_ssl_certs:
             kwargs["verify"] = ssl.create_default_context(
-                capath=stripe.ca_bundle_path
+                cafile=stripe.ca_bundle_path
             )
         else:
             kwargs["verify"] = False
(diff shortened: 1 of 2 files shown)


jar-stripe:
replaced "Fixes" with "Reported in" in the description, so we don't automatically close the issue.

helenye-stripe (approved):

PR #1440, opened 2025-01-13
add justfile, update readme, remove coveralls

xavdid-stripe:
### Why?

In an effort to modernize and simplify our local tooling, we're moving our dev commands from makefiles to [justfiles](https://github.com/casey/just). This is intended to be mostly a drop-in replacement, but some command names may change per standardization efforts.

As part of simplifying the CI, I removed the `Makefile` and `tox` setup (since I had tweaked everything too far for them to be useful).

### What?

- added justfile
- removed makefile
- removed tox dependency and configuration files
- simplified CI (moved more steps into a single, statically-versioned step and more!)
- updated README
- removed coveralls

### See Also

[DEVSDK-2325](https://go/j/DEVSDK-2325)

## Why Not Tox?

`tox` is great at simultaneously managing multiple python versions, but we don't really do that much. GH Actions handles versions in CI and locally we all just work off the same one version (and mise can swap us if we need). So our pattern of make -> tox -> pytest (/ other tools) was way harder to reason about that it needed to be. 

Instead, each tool uses whatever default `python` is available and we ensure unit test coverage by switching out those versions. If we want to work/test against different versions locally, we manage our versions with [mise](https://mise.jdx.dev/lang/python.html) and use the scripts to reset the env.


helenye-stripe (approved):
lgtm! 🔥

PR #1441, opened 2025-01-13
testing pypy in CI

xavdid-stripe:
I have a hunch that our tests don't pass on older versions of pypy, so I wanted to test that in isolation.


xavdid-stripe:
Turns out tox is even more confusing than I thought!

PR #1445, opened 2025-01-16
minor justfile fixes & pin CI version

xavdid-stripe:
### Why?

During this project, we changed how the extra justfile was imported, so we need to tweak this one. We also pin the ubuntu version in CI so it doesn't auto-update on us.

### What?

- update import
- remove extra `@` from bash script

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ on:
 jobs:
   lint:
     name: Static Checks
-    runs-on: ubuntu-latest
+    runs-on: "ubuntu-24.04"
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ on:
 jobs:
   lint:
     name: Static Checks
-    runs-on: ubuntu-latest
+    runs-on: "ubuntu-24.04"
```
prathmesh-stripe, line 23: Any reason why we are pining this to a specific version of ubuntu?

prathmesh-stripe (approved):

PR #1456, opened 2025-02-13
upgrade ruff version

xavdid-stripe:
### Why?

Bumps our formatter (`ruff`) version to latest to take advantage of the newer python LSP ([docs](https://docs.astral.sh/ruff/editors/migration/)). Also just good to bump these once in a while. After bumping, I ran `just format`, which formatting one file slightly differently than before ([expected](https://astral.sh/blog/ruff-v0.9.0))

Tried to bump others, but hit linting errors I didn't want to fix now.

### What?

- bumps `ruff` version
- re-formats everything using `just format` (1 file changed)
- adds comment to other tools versions

### DIFF
--- a/stripe/_ephemeral_key.py
+++ b/stripe/_ephemeral_key.py
@@ -147,8 +147,7 @@ class EphemeralKey(
     def create(cls, **params):
         if params.get("stripe_version") is None:
             raise ValueError(
-                "stripe_version must be specified to create an ephemeral "
-                "key"
+                "stripe_version must be specified to create an ephemeral key"
             )
 
         url = cls.class_url()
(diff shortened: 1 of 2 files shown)


stripe/_ephemeral_key.py
```diff
@@ -147,8 +147,7 @@ async def delete_async(  # pyright: ignore[reportGeneralTypeIssues]
     def create(cls, **params):
         if params.get("stripe_version") is None:
             raise ValueError(
-                "stripe_version must be specified to create an ephemeral "
-                "key"
```
xavdid-stripe, line 151: autoformatted using the newer algorithm

prathmesh-stripe (approved):

PR #1474, opened 2025-03-26
[BREAKING] rename `StripeStreamResponseAsync`'s `.read()` to `read_async()` for consistency

xavdid-stripe:
### Why?

We added a linter ages ago to enforce that all async functions have the `_async` suffix. This consistency ensures that users are properly `await`ing them (failure to do so can be the source of tricky bugs). 

There was one exception to that rule, which I'm removing here.

While it's a breaking change for 100% of users, it will be a loud one. The method is renamed and they'll get both typecheck and runtime failures when they use the old method. Also, it only affects a single user-facing method.

This was the one method in our codebase that didn't comply with the convention, so it's being updated

### What?

- renamed `StripeStreamResponseAsync.read()` to `StripeStreamResponseAsync.read_async()`
- fixed test

### See Also

- https://github.com/stripe/stripe-python/pull/1326

## Changelog

- ⚠️ Rename `StripeStreamResponseAsync.read()` to `read_async()`
    - This brings the method name in line with the conventions used by every other async method in the package, ensuring consistent `async` usage.
    - You'll need to update your code if you call `Quote.pdf_async().read()` method. A typechecker will alert you to this change.


stripe/_stripe_response.py
```diff
@@ -61,6 +61,5 @@ def __init__(
     def stream(self) -> AsyncIterable[bytes]:
         return self._stream
 
-    # TODO (MAJOR): rename this to `read_async`
-    async def read(self) -> bytes:  # noqa: ASY100
+    async def read_async(self) -> bytes:
```
mbroshi-stripe, line 64: Question: Is there a reason you don't do a two-step in the rename?
1. Add `@deprecated` to `read`, add the new `read_async` and have `read` call `read_async`
2. Remove `read` in the next release

xavdid-stripe: We could if it were a bigger lift or affected more endpoints, but this is a fairly small change in difficulty/scope. IMO It falls under the category of "it's a breaking change, please update your code". I'd just as soon get it off our plate.

helenye-stripe (approved):
the nittiest of nits: can we use ⚠️ in the changelog for consistency?

PR #1476, opened 2025-03-27
Update add_beta_version logic

mbroshi-stripe:
### Why?

We decided to allow `add_beta_version` to overwrite an existing beta version

### What?

- Validates that `beta_version` is of the form "v" + a number
- Allows overwriting existing `beta_name`. The new behavior is that the higher `beta_version` is accepted instead of raising an `Exception`.

## Changelog
- `stripe.add_beta_version` will use the highest version number used for a beta feature instead of raising an `Exception` on a conflict as it had done previously.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -129,21 +129,25 @@ def add_beta_version(
     beta_version: str,
 ):
     # Validate beta_version format
-    if not re.match(r"v\d+", beta_version):
+    if not beta_version.startswith("v") or not beta_version[1:].isdigit():
         raise ValueError(
             f"Invalid beta_version format: {beta_version}. Expected format is 'v' followed by a number (e.g., 'v3')."
         )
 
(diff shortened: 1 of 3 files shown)


stripe/__init__.py
```diff
@@ -638,6 +653,7 @@ def __getattr__(name):
 from stripe._treasury_service import TreasuryService as TreasuryService
 from stripe._v2_services import V2Services as V2Services
 from stripe._webhook_endpoint import WebhookEndpoint as WebhookEndpoint
+import re
```
helenye-stripe, line 656: Is there any way to not require this?

helenye-stripe (dismissed):

stripe/__init__.py
mbroshi-stripe: Agreed. I'll rewrite without `re`

helenye-stripe (approved):

PR #1483, opened 2025-04-02
Fix event names and related objects

ramya-stripe:
### Why?
Manually creating PRs as automation is currently failing
Re-generating after the latest set of fixes in the codegen around fixing event names and related objects

### What?
Re-run code gen


stripe/_invoice.py
```diff
@@ -5620,7 +5620,6 @@ class VoidInvoiceParams(RequestOptions):
     The status of the invoice, one of `draft`, `open`, `paid`, `uncollectible`, or `void`. [Learn more](https://stripe.com/docs/billing/invoices/workflow#workflow-overview)
     """
     status_transitions: StatusTransitions
-    subscription: Optional[ExpandableField["Subscription"]]
```
xavdid-stripe, line 5623: is this intentional?

stripe/_subscription.py
```diff
@@ -2363,7 +2363,7 @@ class SearchParams(RequestOptions):
     """
     A date in the future at which the subscription will automatically get canceled
     """
-    cancel_at_period_end: Optional[bool]
```
xavdid-stripe, line 2366: also this one

xavdid-stripe (approved):
event changes look good. There are a couple of unrelated changes that I wanted to double check were supposed to be included.

stripe/_invoice.py
ramya-stripe: This should have gone in as part of prior PRs - not sure why it got left out - yes this is intentional - it is one of the overrides we added

stripe/_subscription.py
ramya-stripe: This should have gone in as part of prior PRs - not sure why it got left out - yes this is intentional - it is one of the overrides we added

ramya-stripe:
> There are a couple of unrelated changes that I wanted to double check were supposed to be included.

These should have gone in as part of prior PRs - not sure why they got left out - yes these are intentional - these are from the two overrides we added

PR #1493, opened 2025-04-08
Update readme with clarification on resource based call patterns

ramya-stripe:
### Why?
The readme uses StripeClient while rest of Stripe docs still use the older resource based call patterns. Users will need some clarity until all the Stripe docs get updated in the coming months

### What?
Added a section in the readme just like we did in PHP a few years ago


jar-stripe (approved):
LG!

PR #1494, opened 2025-04-08
switch build systems

xavdid-stripe:
### Why?

We've been having CI failures (such as [this](https://github.com/stripe/stripe-python/actions/runs/14344596118/job/40211702000?pr=1493)) because we're using a deprecated approach to building python packages (calling `python setup.py`) and trying to target a deprecated format (eggs). Rather than figure out exactly what broke, I've swapped us to using [build](https://github.com/pypa/build), the current industry standard in favor of `setuptools`.

Separately, I added better logging to our dependency installation when running in CI, which will help us understand better why something might have broken.

Note: this has no user-facing impact. We're just building our package for distribution in a more modern way.

### What?

- swap `python setup.py` call to `python -m build`
- swap `setuptools` dependency for `build`
- add more logging to dependency installation when running in CI.

### See Also

- https://setuptools.pypa.io/en/latest/deprecated/commands.html
- https://github.com/stripe/stripe-python/actions/runs/14344596118/job/40211702000?pr=1493

### DIFF
--- a/deps/build-requirements.txt
+++ b/deps/build-requirements.txt
@@ -1,4 +1,5 @@
 # packages needed to package & release
 
-twine
-setuptools
+# pinned to latest as of 2025-04-08; nothing special about these versions
+twine == 6.1.0
+build == 1.2.2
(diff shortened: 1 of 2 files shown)


jar-stripe (approved):
All for more reliable builds!

PR #1499, opened 2025-04-18
Encode bools with lower case

mbroshi-stripe:
### Why?

We discovered that in V2, the Stripe API is more strict on boolean query parameters, and `True`/`False` are no longer accepted as boolean values, rather instead they need to be `true`/`false`. This will cause V2 Python SDK requests with boolean values to fail.

### What?

- Sets boolean query parameter values to `true`/`false` before sending to the Stripe API.


jar-stripe (approved):
LG!

PR #1518, opened 2025-05-16
build SDK w/ new spec

xavdid-stripe:
DON'T MERGE. This is just to see the output of snapshots built w/ openapi v2.

CI may fail because of changes to `DeletedObject`, which is expected


stripe/v2/billing/_meter_event_session.py
```diff
@@ -34,3 +34,7 @@ class MeterEventSession(StripeObject):
     """
     Has the value `true` if the object exists in live mode or the value `false` if the object exists in test mode.
     """
+
+    @classmethod
+    def class_url(cls):
```
xavdid-stripe, line 39: I don't think this should be here

stripe/_api_requestor.py
```diff
@@ -201,6 +201,7 @@ def request(
             params=params,
             requestor=requestor,
             api_mode=api_mode,
+            v2_deleted_object=method == "delete" and api_mode == "V2",
```
xavdid-stripe, line 204: this is the runtime code that ensures we turn delete responses into `V2DeletedObject` instead of `EventDestination`

We're not merging this second, but it's what we'll do when we _are_ ready to merge.

stripe/_util.py
```diff
@@ -198,6 +198,12 @@ def get_thin_event_classes():
     return THIN_EVENT_CLASSES
 
 
+def get_deleted_object():
```
xavdid-stripe, line 201: busting a circular import

stripe/v2/_deleted_object.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+from stripe._stripe_object import StripeObject
+
+
+class DeletedObject(StripeObject):
```
xavdid-stripe, line 5: manually written class, references to which will get spit out by codegen

tests/test_api_requestor.py
```diff
@@ -320,6 +322,55 @@ async def test_empty_methods_async(self, requestor, http_client_mock):
...
+
+            resp = requestor.request(method, path, {}, base_address="api")
+
+            post_data = None
+
+            http_client_mock.assert_requested(method, post_data=post_data)
+            if path == self.v1_path:
+                assert isinstance(resp, Customer)
+            else:
+                assert isinstance(resp, DeletedObject)
```
xavdid-stripe, line 343: actual tests to ensure that delete responses work as expected. Tests are green!

PR #1534, opened 2025-06-30
Fix the link to releases page from readme

ramya-stripe:
### Why?
The current link goes to the .NET release page

### What?
Fix the link


jar-stripe (approved):
👍

PR #1537, opened 2025-07-11
Fix timeout type hint in RequestsClient

abe-101:
### Why?

The `RequestsClient`'s `__init__` method incorrectly typed the `timeout` parameter as `int`. The underlying `requests` library accepts a float for a total timeout or a `(connect_timeout, read_timeout)` tuple.

### What?

This commit updates the type hint to `Union[float, Tuple[float, float]]` to accurately reflect the timeout in the requests library.

### See Also

https://github.com/psf/requests/blob/91a3eabd3dcc4d7f36dd8249e4777a90ef9b4305/src/requests/sessions.py#L538

### DIFF
--- a/stripe/_http_client.py
+++ b/stripe/_http_client.py
@@ -595,7 +595,7 @@ class RequestsClient(HTTPClient):
 
     def __init__(
         self,
-        timeout: int = 80,
+        timeout: Union[float, Tuple[float, float]] = 80,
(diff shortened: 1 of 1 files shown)


abe-101:
Good job bot (@cla-assistant) but why 3 duplicate message?

helenye-stripe:
@abe-101 Thanks for this change!

PR #1538, opened 2025-07-11
Add strongly typed Thin Event push payloads

xavdid-stripe:
### Why?

As part of the Next Gen Event Handling project, we're adding fully typed push payloads to thin events. These describe the full union of thin events a user could receive. 

### What?

- [DEVSDK-2662](https://go/j/browse/DEVSDK-2662)

### DIFF
--- a/stripe/_stripe_client.py
+++ b/stripe/_stripe_client.py
@@ -27,9 +27,10 @@ from stripe._stripe_response import StripeResponse
 from stripe._util import _convert_to_stripe_object, get_api_mode, deprecated  # noqa: F401
 from stripe._webhook import Webhook, WebhookSignature
 from stripe._event import Event
-from stripe.v2._event import ThinEvent
+from stripe.v2._event import UnknownThinEvent
 
 from typing import Any, Dict, Optional, Union, cast
+from typing_extensions import TYPE_CHECKING
 
 # Non-generated services
 from stripe._oauth_service import OAuthService
@@ -110,6 +111,9 @@ from stripe._treasury_service import TreasuryService
 from stripe._webhook_endpoint_service import WebhookEndpointService
 from stripe._v2_services import V2Services
 # services: The end of the section generated from our OpenAPI spec
+
+if TYPE_CHECKING:
+    from stripe.events._event_classes import All_PUSHED_THIN_EVENTS
 
 
 class StripeClient(object):
@@ -274,7 +278,7 @@ class StripeClient(object):
         sig_header: str,
         secret: str,
         tolerance: int = Webhook.DEFAULT_TOLERANCE,
-    ) -> ThinEvent:
+    ) -> "All_PUSHED_THIN_EVENTS":
         payload = (
             cast(Union[bytes, bytearray], raw).decode("utf-8")
             if hasattr(raw, "decode")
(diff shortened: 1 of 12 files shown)


flake8_stripe/flake8_stripe.py
```diff
@@ -35,6 +35,7 @@ class TypingImportsChecker:
         "Unpack",
         "Awaitable",
         "Never",
+        "override",
```
xavdid-stripe, line 38: This is fine because we're dropping python 3.6, so we get access to a newer version of `typing_extensions`

stripe/v2/_event.py
```diff
@@ -115,37 +116,119 @@ class ThinEvent:
     """
     created: str
     """
-    Livemode indicates if the event is from a production(true) or test(false) account.
+    Time at which the object was created.
```
xavdid-stripe, line 132: bugfix - comments go below their property in python

xavdid-stripe:
(ignore red CI; docker hub is having a problem so all tests are down. Everything besides compat was good before i merged master back into this and trigger the CI run.)

stripe/__init__.py
```diff
@@ -100,8 +100,6 @@ def _warn_if_mismatched_proxy():
 # StripeClient
 from stripe._stripe_client import StripeClient as StripeClient  # noqa
 
-from stripe.v2._event import ThinEvent as ThinEvent  # noqa
```
jar-stripe, line 103: don't we need to to import event_notification?  or did our strategy change here?

stripe/_stripe_client.py
```diff
@@ -238,6 +250,9 @@ def raw_request(self, method_: str, url_: str, **params):
 
         stripe_context = params.pop("stripe_context", None)
 
+        # we manually pass usage in event internals, so use those if available
+        usage = params.pop("usage", ["raw_request"])
```
jar-stripe, line 254: what does this do if there are multiple usage strings, and what happens if usage exists in both params and usage when supplied to request_raw?

stripe/v2/_event.py
```diff
@@ -115,37 +129,119 @@ class ThinEvent:
...
+            f"/v2/core/events/{self.id}",
+            stripe_context=self.context,
+            usage=["pushed_event_pull", "pushed_event_pull_async"],
+        )
+        return cast(Event, self._client.deserialize(response, api_mode="V2"))
+
+
+class UnknownEventNotification(EventNotification):
+    """
+    Represents a Thin Event payload that the SDK doesn't have types for. May have a related object.
```
jar-stripe: "Thin Event" should be Event Notification

stripe/__init__.py
xavdid-stripe: The intention is that users write `from stripe.v2 import EventNotification` if they need it. It's the same way they'd access a v2 `Event`

stripe/_stripe_client.py
xavdid-stripe: > what does this do if there are multiple usage strings

It gets the `list[str]` out of params, if present. So if there are multiple usage strings, it'll grab them. If it's missing, then it defaults to `['raw_request']`, which is what we had before.

Had to add this because we're calling `raw_request` from `fetch_related_object`. I don't want that usage counted towards raw_request.

> what happens if usage exists in both params and usage when supplied to request_raw?

this removes usage from params, so `request_raw` won't get both

PR #1543, opened 2025-08-06
add support for v2 deleted objects

xavdid-stripe:
### Why?

The typing of delete calls has always been a little confusing in V1: SDKs say they return an entire object (like a `Customer`) but with the `deleted` property set to `True`. That's true for some APIs but not for others. For consistency, we just always say you get a "real" object.

V2 APIs are more strict about the data they return: there's only ever `id` and `object`. To better represent this, we're now generating a `V2.DeletedObject` class that is returned from all v2 delete calls. This PR includes that generated class, but APIRequestor-related code to use it when appropreate.

> Note: this PR is generated using the V2 OpenAPI spec rather than the protos so that the `DeletedObject` class is generated. 

### What?

- add param to `_convert_to_stripe_object`
- provide that param from the API requestor
- add tests

### See Also

[DEVSDK-2588](https://go/j/DEVSDK-2588)


helenye-stripe:
I'm peaceful with this but won't the generated changes get removed by CI since we'll go back to using the proto spec to generate the SDK?

xavdid-stripe:
> I'm peaceful with this but won't the generated changes get removed by CI since we'll go back to using the proto spec to generate the SDK?

the manual changes won't. Plus I don't think we're running codegen against `sdk-release/next-major`. Worst case scenario, CI fails beacuse it's using this code against the proto generation, but I think it won't (because of the branch). The ordering is definitely tricky there though!

helenye-stripe (approved):

PR #1549, opened 2025-08-14
Introduce V1 namespaces in StripeClient

prathmesh-stripe:
### What?
We are introducing a V1 namespace to improve code organization in Stripe SDKs.

## Changelog
- Introduced V1 service in StripeClient. All the existing V1 StripeClient services(eg. customers, products) are copied under the new V1 service. Service accessors living directly under StripeClient will be marked as deprecated in the next major release and will be removed in a future release. Eg.
```diff
client = StripeClient("sk_test...")

# Accessing V1 Stripe services on a StripeClient should be through the V1 namespace
- client.customers.list() 
+ client.v1.customers.list()
```
Refer to the [migration guide](https://github.com/stripe/stripe-python/wiki/v1-namespace-in-StripeClient) for help upgrading.

### DIFF
--- a/stripe/__init__.py
+++ b/stripe/__init__.py
@@ -567,6 +567,7 @@ from stripe._transfer_reversal_service import (
 )
 from stripe._transfer_service import TransferService as TransferService
 from stripe._treasury_service import TreasuryService as TreasuryService
+from stripe._v1_services import V1Services as V1Services
 from stripe._v2_services import V2Services as V2Services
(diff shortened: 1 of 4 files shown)


helenye-stripe (commented):
Same as Java, should we mark changelog with ⚠️

prathmesh-stripe:
We have used that for breaking changes in the past. This is not a breaking change.

Issue #1427, opened 2024-11-25
Stripe import is very slow


silviogutierrez:
### Describe the bug

I looked around, searched the issues and even googled. Surprised this hasn't come up.

Just importing the library adds 500ms to startup time. There's got to be a way to lazy load parts.

This is on a top of the line Mac with SSD.

### To Reproduce

```python
# Takes about 20 ms
time python -c "print('ok')";

# Takes bout 500ms
time python -c "print('ok'); import stripe"
```

### Expected behavior

For serverless setups, startup time is pretty important.

Even for local development, django's runserver or uvicorn's reload are both much slower with imports like these.

Of course you can do the import locally everywhere stripe is used, but this can be problematic.

At the very least, some guidance in the docs on how to do imports and have a lazy-loaded client would help. Something like a central `get_stripe_client() -> Stripe` or the like. And cover thread-safety in here.

### OS

macOS

### Language version

3.12.1

### Library version

11.3.0

### API version

2020-08-27

silviogutierrez:
For production, yes, it's mostly an issue with cold starts. If you use python's `importtime` you can see Stripe takes up a big, big chunk. See more here: https://adamj.eu/tech/2023/03/02/django-profile-and-improve-import-time/

Cold start platforms: Google Cloud Run in my case.

But it also has local dev implications. Django and FastAPI rely on an autoreloading server. Unlike frontend tools like vite, there's no concept of hot module reloading. So this library — with no usage, just importing — is responsible for a 500ms lag in restarts.

Imagine a few other libraries like this (Google storage does the same thing), and it starts to add up.

prathmesh-stripe:
Thanks for reporting this. We will be looking into SDK optimizations in the new year.

bufke:
I took a peek with memray. I see that it imports both aiohttp and requests, lovely. Those were installed in my testing environment.  _test_helpers_service looks questionable for importing. https://gitlab.com/-/snippets/4802529

stianjensen:
Some of the import time regression could be inside requests itself, depending on which version of requests you're getting with the different versions of stripe you tested:
https://github.com/psf/requests/issues/6790

rsinger86:
I had to increase the capacity of my Fargate ECS service after adding this SDK to my project. Some optimizations would be much appreciated.

chris-mui:
[...]

PR #1553, opened 2025-08-19
Import available http libraries more efficiently

xavdid-stripe:
### Why?

As discussed in https://github.com/stripe/stripe-python/issues/1427, the way we decide which HTTP library we should use involves a lot of unnecessary imports.

This PR improves that by making all the imports lazy and by stopping looking once we've found a library that works. It's backwards compatible.

Basically, it moves all the type checking in an `if TYPE_CHECKING` block. At runtime, it returns as soon as its found _any_ HTTP library (keeping the same precedence). There's no user-facing impact of this change.

It required retooling the tests a lot, since we're not putting http modules at the root level anymore. Also cleaned up a bunch of other things while I was in there.

This resulted in ~ a 50% speedup for importing `_api_requestor.py`, which is a good starting place!

### What?

- lazily import http libraries
- adjust tests
- (unrelated) fix justfile so single test running works correctly

### See Also

- [DEVSDK-2708](https://go/browse/DEVSDK-2708)


justfile
```diff
@@ -10,9 +10,10 @@ _default:
     just --list --unsorted
 
 # ⭐ run all unit tests
+[positional-arguments]
 test *args: install-test-deps
     # configured in pyproject.toml
-    pytest {{ args }}
+    pytest "$@"
```
jar-stripe, line 16: mostly curiosity, but why is this better than {{ args }} ?

stripe/_http_client.py
```diff
@@ -32,113 +31,103 @@
...
+    """
+    This method creates and returns a new HTTPClient based on what libraries are available. It uses the following precedence rules:
 
-    return impl(*args, **kwargs)
+    1. Urlfetch (this is provided by Google App Engine, so if it's present you probably want it)
+    2. Requests (popular library, the top priority for all environments outside Google App Engine, but not always present)
+    3. Pycurl (another library, not always present, not as preferred as Requests but at least it verifies SSL certs)
+    4. urllib with a warning (basically always present, a reasonable final default)
+
+    For performance, it only imports what it's actually going to use. But, it re-calculates every time its called, so probably save its result instead of calling it multiple times.
```
jar-stripe, line 76: is this a todo?

stripe/_http_client.py
```diff
@@ -32,113 +31,103 @@
...
+    """
+    try:
+        from google.appengine.api import urlfetch  # type: ignore # noqa: F401
```
jar-stripe, line 79: are we optimizing for the google app engine or requests?  does this leave the url lib folks no better off?

stripe/_http_client.py
```diff
@@ -610,8 +601,12 @@ def __init__(
         self._session = session
         self._timeout = timeout
 
-        assert requests is not None
-        self.requests = requests
+        if _lib is None:
+            import requests
```
jar-stripe, line 597: do we need these imports if new_default_http_client will import the right package before creating this class?  same applies for the others

stripe/_http_client.py
```diff
@@ -1103,6 +1110,7 @@ def close(self):
         pass
 
 
+# todo(major): rename this, urllib2 isn't called that anymore
```
jar-stripe, line 1105: an idea: we can rename it in this release and just declare a global variable as an alias

jar-stripe: otherwise, we should create a ticket and add it to the next major epic and add it to the todo

jar-stripe (approved):
some non blocking comments.  The test file looks like a lot of related but not directly applicable cleanup/ refactor so I skimmed those parts but what I read looked good!

stripe/_http_client.py
xavdid-stripe: no, it's a warning to users. Like, they can call this over and over, but they're better off saving the result instead. We already do that.

stripe/_http_client.py
xavdid-stripe: App engine. And yeah, correct

stripe/_http_client.py
jar-stripe: Oh gotcha; that makes more sense 😅

stripe/_http_client.py
xavdid-stripe: I went back and forth. Ultimately, I think we want the clients to be self-sufficient (so they need to be able to import their respective libs) in case people subclass them.

So ultimately we're doing the import to check if it exists and then picking the class accordingly. Looks a little funny, but works (and is backwards compatible).

stripe/_http_client.py
jar-stripe: Maybe not for this PR but I wonder if theres a less expensive way to check if a library is present.  Probably not, given Python's dynamic nature 🫤

stripe/_http_client.py
jar-stripe: Ah that makes sense!

justfile
xavdid-stripe: explained in https://github.com/casey/just?tab=readme-ov-file#positional-arguments, but it affected us because if you did `just test -k "a and b"` before (to tell pytest to run tests with both "a" and "b" in the name), then you'd get `file not found: and` because `just` was merging everything and ignoring quotes.

We do the same thing in Node! we just hadn't done it here yet, apparently.

PR #1563, opened 2025-08-31
fix: paginate backwards if `starting_after == None`

### LINKED ISSUE #1562: `auto_paging_iter` does not paginate backwards if `starting_after` is `None`
### Describe the bug

I was working with the `auto_paging_iter` to implement Relay pagination over invoices, and I encountered an interesting quirk in the way it is implemented. The `auto_paging_iter` code currently checks if `ending_before` is in the retrieve params and `starting_after` is not to decide to reverse the results and paginate backwards:

```python
def _auto_paging_iter(self) -> Iterator[T]:
    page = self

    while True:
        if (
            "ending_before" in self._retrieve_params
            and "starting_after" not in self._retrieve_params
        ):
            for item in reversed(page):
                yield item
            page = page.previous_page()
        else:
            for item in page:
                yield item
            page = page.next_page()

        if page.is_empty:
            break
```

In my code I am passing along all input args because, at least in the services model, the SDK correctly ignores and drops any `params` passed as `None`:

```python
for stripe_invoice in stripe_client.v1.invoices.list(
    params={
        "limit": 100,
        "customer": dynamodb_alarm_user.customer_id,
        "starting_after": parsed_input.after,
        "ending_before": parsed_input.before,
        "expand": ("data.customer",),
    },
    options=merge_default_options(
        {"stripe_account": dynamodb_jurisdiction.account_id}
    ),
).auto_paging_iter():
```

This leads to the following params being shown in the request logs on Stripe's side:

```json
{
  "customer": "cus_SyBoRWyoh9GtxR",
  "ending_before": "in_1S2FQhBARyXc9bEw3n0eQQwj",
  "expand": {
    "0": "data.customer"
  },
  "limit": "100"
}
```

However, because `starting_after` is present in `self._retrieve_params` even though it is `None`, the `auto_paging_iter` logic falls into the else cases and does not reverse the results and calls `page.next_page()` instead of `page.previous_page()`.

This leads to an incorrect pagination after the first page of results, because it uses the wrong cursors and arguments to retrieve the next page.

### To Reproduce

1. create N invoices to be able to paginate over under the same stripe account.
2. use `stripe_client.v1.invoices.list(params={"limit": 1, "starting_after": None, "ending_before": invoices[-1].id}).auto_paging_iter()` to try and iterate the results.

### Expected behavior

The auto paging iter should iterate backwards over the invoices list from the end to the beginning, excluding the invoice specified as `ending_before`.

### Code snippets

```Python

```

### OS

Ubuntu 22.04.5 LTS

### Language version

Python 3.13.3

### Library version

12.5.0

### API version

2025-08-27.basil

### Additional context

_No response_

ryancausey:
### Why?
This fixes the case where `starting_after` is present in `_retrieve_params` but is `None` so that the `auto_paging_iter` correctly paginates backwards.

### What?

- This fixes the case where `starting_after` is present in `_retrieve_params` but is `None` so that the `auto_paging_iter` correctly paginates backwards.
- adds tests to confirm this behavior

### See Also
closes #1562

### DIFF
--- a/stripe/_list_object.py
+++ b/stripe/_list_object.py
@@ -132,8 +132,8 @@ class ListObject(StripeObject, Generic[T]):
 
         while True:
             if (
-                "ending_before" in self._retrieve_params
-                and "starting_after" not in self._retrieve_params
+                self._retrieve_params.get("ending_before") is not None
+                and self._retrieve_params.get("starting_after") is None
             ):
                 for item in reversed(page):
                     yield item
@@ -151,8 +151,8 @@ class ListObject(StripeObject, Generic[T]):
 
         while True:
             if (
-                "ending_before" in self._retrieve_params
-                and "starting_after" not in self._retrieve_params
+                self._retrieve_params.get("ending_before") is not None
+                and self._retrieve_params.get("starting_after") is None
             ):
                 for item in reversed(page):
                     yield item
--- a/tests/api_resources/test_list_object.py
+++ b/tests/api_resources/test_list_object.py
@@ -331,7 +331,9 @@ class TestAutoPaging:
             self.pageable_model_response(["pm_125", "pm_126"], True), "mykey"
         )
         lo._retrieve_params = {
-            "foo": "bar", "ending_before": "pm_127", "starting_after": None
+            "foo": "bar",
+            "ending_before": "pm_127",
+            "starting_after": None,
         }
 
         http_client_mock.stub_request(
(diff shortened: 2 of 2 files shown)


mbroshi-stripe:
@ryancausey Looks great! Can you run `just format` locally and commit? There are lint errors

ryancausey:
Thanks for formatting and merging!

PR #1570, opened 2025-09-12
Don't use mutable default arguments

xavdid-stripe:
### Why?

Mutable default function arguments are a [classic python footgun](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments). Though we haven't gotten any issues related to this, it's a bug waiting to happen. 

### What?

- add linter to catch use of mutable default args
- set param defaults to `None` instead of `{}`. The function(s) we're calling also take optional params, so we'll leave it up to them to pick a default
- similar manual fixes
- add new linter to catch these in the future

### See Also

[DEVSDK-2139](https://go/j/DEVSDK-2139)

## Changelog

- Service methods now correctly set `None` as the default function argument instead of `{}`


prathmesh-stripe:
> This shouldn't be a breaking change, since we weren't modifying these values internally. But, it's worth calling out in the changelog.

We can get rid of this line from the changelog.

Issue #1558, opened 2025-08-29
Async methods fail with ImportError while sync methods work out of the box


geminimir:
### Describe the bug

Noticed small inconsistency in dependency requirements between sync and async methods... Sync methods work immediately after installing the stripe package, while async methods fail with an ImportError unless additional dependencies are installed... 

### To Reproduce

1. Install stripe: pip install stripe
2. Try to use any async method without installing httpx/aiohttp
3. Observe ImportError

### Expected behavior

Both sync and async methods should either:

- Work out of the box with the base installation, OR
- Clearly document that async support requires additional dependencies

### Code snippets

```Python
import stripe

client = stripe.StripeClient("sk_test_...")

# This works fine
customer = client.customers.retrieve("cus1")  

# This fails with ImportError
customer = await client.customers.retrieve_async("cus1")
```

### OS

all

### Language version

3.9+

### Library version

Latest

### API version

2025-08-29

mbroshi-stripe:
👋 Hi @geminimir! Thanks for reporting this issue! I will start investigating and circle back as needed.

mbroshi-stripe:
I was able to reproduce the reported issue. Did you see our README section on using async requests: https://github.com/stripe/stripe-python?tab=readme-ov-file#async? I'm wondering if that clarifies the intended usage for you, or if we should be more explicit there about imports?

meet-vasita:
Hi, I’d like to work on this issue as my first contribution.

xavdid-stripe:
Hey @meet-vasita - thanks for the offer! Luckily, this is just a small docs update for now, so we'll handle it on our end. There's some other work we may do to make this experience better though!

xavdid-stripe:
This is fixed in 13.0.1: you can now `pip install stripe[async]` and it'll work out of the box.

Thanks for flagging!

PR #1572, opened 2025-09-19
migrate from `setup.py` to `pyproject.toml`

xavdid-stripe:
### Why?

tldr: the `pyproject.toml` file is the modern way to declare a python package. By moving to it (from `setup.py`), we'll be able to simplify our dependencies and more cleanly tackle user issues like https://github.com/stripe/stripe-python/issues/1558. 

This change is backwards compatible as long as you're using a package manager released in the last 7 years. There are no user-facing changes otherwise.

If that's enough information for your review, then great! Otherwise, read on.

### More than you've ever wanted to know about Python packages

Python packaging has been long been a sore spot in the ecosystem. It's complicated and there are a _lot_ of ways to do it (many of which are considered "legacy" or "only for really specific circumstances"). Good news though: there's been a lot of work to write real standards and move packages towards them. There are also now multiple tools that implement these standards with minimal configuration.

All packages must have a manifest file that declare the name, version, dependencies, etc (roughly equivalent to JS's `package.json`). Originally, this was handled in `setup.py`, a Python file that called some functions from the `setuptools` library to register static information about the package. This created a big headache for building and distributing package. Because dependencies were declared in a call to a Python function, you couldn't use them in that file before they were installed. But, you didn't know what to install before you executed the file. It was a big mess that you can read about in [great detail here](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

The modern alternative is the `pyproject.toml` file, as introduced in [PEP 518](https://peps.python.org/pep-0518). It can store much of the same information as `setup.py`, but can be parsed without executing anything (and in a language-agnostic way).

Our Python package was written before this standardization happened, so it's still declared with the legacy (but [still supported](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/#setup-py-deprecated)) `setup.py` file. It's worked well enough, but lacks some of the modern packaging features we'd like to use. So, we moved!

The migration itself was straightforward. There's even a [nice guide](https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#where-to-start). Most of the keys in `setup.py` have analogues in `pyproject.toml`. Plus, we can dynamically reference the contents of files and our build backend (more on that soon) will read them and sort everything out for us.

Here's what happened to each property from `setup.py`:

| property                        | outcome                        | notes                                                                                                                                                             |
| ------------------------------- | ------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `name`                          | migrated                       |                                                                                                                                                                   |
| `version`                       | made into a string             |                                                                                                                                                                   |
| `description`                   | migrated                       |                                                                                                                                                                   |
| `long_description`              | removed                        | It's what shows up on the PyPI page. Most projects just use the readme (and so should we)                                                                         |
| `long_description_content_type` | removed                        | no longer needed                                                                                                                                                  |
| `author`                        | migrated                       |                                                                                                                                                                   |
| `author_email`                  | migrated                       |                                                                                                                                                                   |
| `url`                           | moved to `project.urls.Source` |                                                                                                                                                                   |
| `license`                       | migrated                       |                                                                                                                                                                   |
| `keywords`                      | migrated                       |                                                                                                                                                                   |
| `packages`                      | removed                        | was only used to exclude the tests directory, which we included elsewhere anyway                                                                                  |
| `package_data`                  | removed                        | not needed anymore, everything in the package directory gets included                                                                                             |
| `zip_safe`                      | removed                        | only relevant for `egg` files, which are removed ([docs](https://setuptools.pypa.io/en/latest/deprecated/zip_safe.html))                                          |
| `install_requires`              | migrated to `dependencies`     |                                                                                                                                                                   |
| `python_requires`               | migrated to `requires-python`  |                                                                                                                                                                   |
| `project_urls`                  | migrated to `project.urls`     | keys updated to match their following their [well-known labels](https://packaging.python.org/en/latest/specifications/well-known-project-urls/#well-known-labels) |
| `classifiers`                   | migrated                       |                                                                                                                                                                   |
| `setup_requires`                | removed                        | using `wheel` directly is no longer needed ([docs](https://github.com/pypa/wheel?tab=readme-ov-file#historical-note))                                             |

The last thing our `pyproject.toml` needs is a `build-system`. 

There are 2 parts to building a Python package. The simpler half is the [build frontend](https://packaging.python.org/en/latest/glossary/#term-Build-Frontend). It's responsible for parsing the `pyproject.toml`, creating an isolated virtualenv, installing the build backend, and running it. The backend takes a directory of Python files and produces a `whl` (a fancy `zip` that package managers can install; [docs](https://realpython.com/python-wheels/)).

Ever since https://github.com/stripe/stripe-python/pull/1494, we've used [build](https://github.com/pypa/build), which is a minimal frontend maintained by the Python Packaging Authority (the group behind PyPI itself). When a `build-system` isn't defined, it falls back to `setuptools.build_meta:__legacy__` as its backend ([docs](https://build.pypa.io/en/stable/differences.html#fallback-backend)), which is what we've been using.

For the backend, we could have kept using `setuptools`, which is still actively maintained. But, its longevity also means its takes a lot of configuration to get exactly what we need. Instead, I went with [flit](https://flit.pypa.io/en/stable/). It's much smaller, more focused, and opinionated. We've got a pure Python package, so we can use it without any configuration. It also means we can remove a bunch of now-unused configuration files like `MANIFEST.in`, `setup.cfg`, and `setup.py` itself.

Put it all together... and we're about back where we started. But with an easier to maintain package and some new opportunities on the horizon!

### What?

- remove `setup.py` from `.flake8`. 
- simplify `.flake8` config to use `extend-ignore` instead of `ignore` (which doesn't require re-ignoring default rules)
- remove outdated installation instructions from README
- move all necessary `setup.py` configuration into `pyproject.toml` (see table above)
- delete `setup.py`, `setup.cfg`, `LONG_DESCRIPTION.rst`, `MANIFEST.in`,  and `.coveragerc`, as they're no longer needed
- add a static `version` to `pyproject.toml` and update the `update-version` justfile recipe to update it (similar to what we do in other languages)

### Testing

I built the package locally and compared it against the one from PyPI. Here's what I found:

- we dropped `py2` from the name of the wheel file (which is expected). We don't support Python 2, so we don't need it in the filename
- the contents of the `.whl` files are identical (minus the change to the version declaration above)
- Our sdist only includes production files (`pyproject.toml` and the contents of the `stripe` directory) rather than all the tests and examples too. This is more correct

I also installed the package directly from each of the distribution methods and verified that everything worked as expected in the REPL.

## See Also

- [DEVSDK-2765](https://go/j/DEVSDK-2765)

## Changelog

- ⚠️ The package has swapped from `setup.py` to `pyproject.toml`. As a result, we're dropping support for `pip < 10.0.0` (released April 2018). Most users won't notice this change.
- Additionally, we're no longer shipping tests or examples in our sdist now, which should offer a small size reduction for the package if installed without the wheel (approx. 2.5MB unzipped)


.coveragerc
```diff
@@ -1,11 +0,0 @@
-[paths]
```
xavdid-stripe, line 1: this should have been removed in https://github.com/stripe/stripe-python/pull/1440, but I missed it somehow

LONG_DESCRIPTION.rst
```diff
@@ -1,34 +0,0 @@
-Official Stripe Bindings for Python
```
xavdid-stripe, line 1: there's nothing in here that's better than what's in the readme, so I ditched it

justfile
```diff
@@ -41,6 +41,7 @@ reset: clean && venv
 
 # build the package for upload
 build: install-build-deps
+    rm -rf dist
```
xavdid-stripe, line 44: Everything will get uploaded to pypi, so we want to make sure this directory only has what we've just built

justfile
```diff
@@ -78,3 +79,5 @@ venv:
 update-version version:
     echo "{{ version }}" > VERSION
     perl -pi -e 's|VERSION = "[.\d\w]+"|VERSION = "{{ version }}"|' stripe/_version.py
+    perl -pi -e 's|^version = "[.\d\w]+"|version = "{{ version }}"|' pyproject.toml
```
xavdid-stripe, line 82: have to update the version in two places now, like most of the other langauges

jar-stripe:
> More than you've ever wanted to know about Python packages

Love this!  Can we pull this into a more durable doc?  This feels like useful information to capture somewhere other than a PR description.

jar-stripe:
How does this help us solve issues like https://github.com/stripe/stripe-python/issues/1558?

xavdid-stripe:
ah, probably should have mentioned 😅

1558 boils down to "I ran `pip install stripe`, but `client.customers.list_async()` fails with an import error while `client.customers.list()` works. what gives?"

`pyproject.toml` supports `optional-dependencies` ([docs](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#dependencies-optional-dependencies)), which will let us declare an explicit dep on `httpx` (our default async client) that users can opt into. So in a follow-up PR, i'll add the dep and instructions to the readme that say something to the effect of:

> If you don't already include an async-capable HTTP client, run `pip install stripe[async]` instead of `pip install stripe`

That'll install our normal deps + anything in the `async` group and we'll be able to close that out.

jar-stripe (approved):

PR #1596, opened 2025-09-25
⚠️ Unify resource and service method parameters into one class

helenye-stripe:
### Why?

Decrease the size of the library for unnecessary duplicated types

### What?

Moves all params (service and resource) to the top level and are shared. 

Generated changes only.

## Changelog
* ⚠️ Resource and service request parameter types have been moved to the top-level and are shared, prepended with their related resource/service
  * For example, `_stripe._account.Account.CreateParams` and `_stripe._account_service.CreateParams` have moved to `_stripe.params._account_create_params.AccountCreateParams`
  * This change only affects users who explicitly refer to params types. No migration is necessary for users otherwise


stripe/_account_capability_service.py
```diff
@@ -1,43 +1,25 @@
 # -*- coding: utf-8 -*-
 # File generated from our OpenAPI spec
+from stripe._account_capability_list_params import AccountCapabilityListParams
```
xavdid-stripe: these should all probably go in an `if TYPE_CHECKING` block. Won't need changes to the main body of the service, since the types are already in a string

xavdid-stripe (approved):

xavdid-stripe:
i've updated `param` -> `params`

PR #1602, opened 2025-09-26
Move `V2.Event` API resources to `V2.Core.Events`

xavdid-stripe:
### Why?

In `2024-09-30.acacia` ([docs](https://docs.stripe.com/changelog/acacia)), we released our new V2 Event system, found under the `/v2/core/events` API endpoint. When we put them in our in SDKs, for presentation reasons, we put these new resources in the `Stripe.V2.Event` namespace.

We've realized that dropping `Core` from the SDK namespace adds more confusion than good, so we're putting the v2 `Event` resource into the `Stripe.V2.Core` namespace, where you'd expect to find it in the first place.

There's no change to specific event classes, like `V1BillingMeterErrorReportTriggeredEvent`. Those are, and will continue to be, under `Stripe.Events`.

### What?

- moved all resources from `V2.Events` to `V2.Core.Events`
- updated tests
- moved some other imports around for consistency

## Changelog

- ⚠️ Move all V2 Event-related resources (`Event`, `RelatedObject`, etc) from `stripe.v2` to `stripe.v2.core`.They now correctly match their API path and are in line with all other resources. To update your code:

```diff
-from stripe.v2 import Event
+from stripe.core.v2 import Event
```

- Existing event types (like `stripe.events.V1BillingMeterErrorReportTriggeredEvent` haven't moved).


examples/event_notification_webhook_handler.py
```diff
@@ -13,6 +13,14 @@
 
 import os
 from stripe import StripeClient
+from stripe import Event
+from stripe.v2.core import Event as V2Event, EventNotification
+from stripe.events import (
+    V2CoreEventDestinationPingEventNotification,
```
xavdid-stripe, line 19: unused, but checks that all of our imports are where we expect them

tests/test_generated_examples.py
```diff
@@ -9599,14 +9599,14 @@ def test_customers_sources_post_service_non_namespaced(
         client.customers.payment_sources.update(
             "cus_123",
             "card_123",
-            {"account_holder_name": "Kamil"},
+            {"name": "Kamil"},
```
xavdid-stripe, line 9602: this was an autochange from master

PR #1603, opened 2025-09-28
remove many deprecated backcompat exports

xavdid-stripe:


tests/test_v2_error.py
```diff
@@ -55,47 +55,47 @@ def test_raises_v2_error(
...
-            assert e.error.message == "bank account is invalid"
-        else:
-            assert False, "Should have raised a InvalidUsBankAccountError"
-
-        http_client_mock.assert_requested(
-            method,
-            path=path,
-            api_key="keyinfo_test_123",
-        )
+    # @pytest.mark.skip("python doesn't have any errors with invalid params yet")
```
xavdid-stripe, line 58: this is pre-emptive test for a class of errors that don't exist yet. It was giving linting errors now that our imports were better, so I commented it out

examples/oauth.py
```diff
@@ -1,11 +1,15 @@
...
-import stripe
+from stripe import StripeClient, OAuth
+from stripe.oauth_error import OAuthError
+
 from flask import Flask, request, redirect
 
 
-stripe.api_key = os.environ.get("STRIPE_SECRET_KEY")
-stripe.client_id = os.environ.get("STRIPE_CLIENT_ID")
+client = StripeClient(
```
prathmesh-stripe, line 9: Does this change work with directly calling Oauth resource?

xavdid-stripe: ah no, you're probably right

xavdid-stripe: good catch!

examples/webhooks.py
```diff
@@ -16,13 +21,11 @@ def webhooks():
     received_sig = request.headers.get("Stripe-Signature", None)
 
     try:
-        event = stripe.Webhook.construct_event(
-            payload, received_sig, webhook_secret
-        )
+        event = Webhook.construct_event(payload, received_sig, webhook_secret)
```
prathmesh-stripe, line 24: This uses stripe.api_key to construct the event so this would have the same issue as Oauth.

prathmesh-stripe (approved):

examples/webhooks.py
xavdid-stripe: Ah, good catch. i'll fix in a follow-up PR!

PR #1604, opened 2025-09-29
Fixed InvoiceLineItem method definition

prathmesh-stripe:
### Why?
Codegen was producing incorrect method definition in resources for InvoiceLineItem and the [snippet](https://docs.stripe.com/api/invoice-line-item/update?lang=python&api-version=2025-03-31.basil) on APIRef was incorrect.

### What?
- Add an override for InvoiceLineItem.modify. Unfortunately this removes the type hints for the params but I think it is okay.
- Service methods are fine and don't need a change. 

### See Also
https://jira.corp.stripe.com/browse/DEVSDK-2597

## Changelog
* ⚠️ `InvoiceLineItem.modify` and `InvoiceLineItem.modify_async` now require `invoice` and `line_item_id` as method parameters.
  * Removed `InvoiceLineItem.ModifyParam` class.


stripe/_invoice_line_item.py
```diff
@@ -234,243 +227,6 @@ class TaxRateDetails(StripeObject):
         """
         _inner_class_types = {"tax_rate_details": TaxRateDetails}
 
-    class ModifyParams(RequestOptions):
```
helenye-stripe, line 237: Why does the params object get deleted here?

prathmesh-stripe: They were automatically generated from the `modify` method. Now that I manually add that method, there is nothing to use as the base to generate these type hints.

helenye-stripe: approving but it would be awesome if we could get it back after the release

PR #1606, opened 2025-09-29
drop support for python 3.6 and clarify version policy

xavdid-stripe:
### Why?

We've got a new Python version support policy! add docs to that, plus 

### What?

- update CI
- update docs to link to our version support policy & clarify supported versions
- bump `requires-python`

### See Also

- [DEVSDK-2772](https://go/j/DEVSDK-2772)

## Changelog

- ⚠️ Drop support for Python 3.6. Please upgrade to Python 3.7+ to use the latest version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,7 @@ jobs:
       fail-fast: false
       matrix:
         python_version:
+          # https://endoflife.date/python
           - "3.7"
           - "3.8"
           - "3.9"
@@ -87,6 +88,7 @@ jobs:
           - "pypy-3.8"
           - "pypy-3.9"
           - "pypy-3.10"
+          - "pypy-3.11"
     name: Test (${{ matrix.python_version }})
     steps:
       - uses: extractions/setup-just@v2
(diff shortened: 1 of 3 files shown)


examples/webhooks.py
```diff
@@ -1,14 +1,10 @@
...
-from stripe import StripeClient, Webhook, SignatureVerificationError
-
+from stripe import Webhook, SignatureVerificationError
 from flask import Flask, request
 
-
-client = StripeClient(
-    api_key=os.environ["STRIPE_SECRET_KEY"],
-    client_id=os.environ.get("STRIPE_CLIENT_ID"),
-)
```
prathmesh-stripe, line 11: minor nit: We do read stripe.api_key in the [Webhook.construct_event](https://github.com/stripe/stripe-python/blob/e10c1203079d9ece98498636cbc4e51882b1041d/stripe/_webhook.py#L31). 
Although this script would work because we don't use the requestor to parse the event, any subsequent interactions that require a call to stripe will fail. Not needed but would be good to tell the users to set the API Key: 
```python
stripe.api_key = 'sk_test...'
```

xavdid-stripe: ah, good call! v1 webhooks are weird.

In the interest of having fewer examples that set the global API key, i'm going to pass it as a kwarg, which `construct_event` accepts.

PR #1622, opened 2025-10-02
Improve webhook example

xavdid-stripe:
### Why?

I'm going back and sprucing up the webhook handling examples. This will also go onto docs.stripe.com.


examples/event_notification_webhook_handler.py
```diff
@@ -1,26 +1,24 @@
 """
-event_notification_webhook_handler.py - receive and process thin events like the
-v1.billing.meter.error_report_triggered event.
+event_notification_webhook_handler.py - receive and process thin events like the v1.billing.meter.error_report_triggered and v1.billing.meter.no_meter_found events.
```
jar-stripe: change thin events to event notifications

examples/event_notification_webhook_handler.py
```diff
@@ -37,22 +35,58 @@ def webhook():
...
+            # these lines are optional, but will give you more accurate typing in this block
+            from typing import cast
+
+            event_notif = cast(UnknownEventNotification, event_notif)
+
+            # continue matching on the type property
+            # from this point on, the `related_object` property _may_ be None (depending on the event type)
+            if event_notif.type == "some.new.event":
+                if event_notif.related_object:
+                    obj = event_notif.fetch_related_object()
```
jar-stripe: ooc what happens if the user calls fetch_related_object without checking for a non-null related_object?  it might be worth adding a comment here about that case

jar-stripe (approved):
Couple small comments but otherwise looks good!

examples/event_notification_webhook_handler.py
xavdid-stripe: we do that check for them and it just returns `None` (and the function signature reflects this). I could swap the example to always call it, but that felt more confusing.

examples/event_notification_webhook_handler.py
xavdid-stripe: good catch. going to mention both, since most of our docs still say thin event (and will for a while)

examples/event_notification_webhook_handler.py
jar-stripe: i think this is fine as is if that's the pattern we want to promote; i think if i were a dev i would be asking the question to maybe avoid a null check in my code, but if people can look at the type or the method comment and see the answer that should be good enough

xavdid-stripe: yeah the real logic is that certain types will always have a related object and the rest won't, the SDK just doesn't know at this point. I was trying to teach that "if `notif.related_object` is present, the fetch will get you something". But unfortunately, the type of `obj` is `StripeObject | None` anyway, since we can't correlate the property being present with a non-null return.

I think folks will just figure it out!

PR #1633, opened 2025-10-03
add httpx to optional deps and only ship tests in sdist

xavdid-stripe:
### Why?

This is two small changes in 1 PR:

- walk back some changes from https://github.com/stripe/stripe-python/pull/1627 that haven't shipped yet to ship only what people need
- a fix for https://github.com/stripe/stripe-python/issues/1558. Users will be able to `pip install stripe[async]` and they'll get everything they need for async requests.

### What?

- added optional dependency group to `pyproject.toml`
- remove everything but `tests` from our sdist

## Changelog

- Add new `async` optional dependency. Now, `pip install stripe[async]` gets you everything you need to make async HTTP calls out of the box.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,8 @@ The default HTTP client uses `requests` for making synchronous requests but
 you to explicitly initialize your own http client and pass it to StripeClient
 or set it as the global default.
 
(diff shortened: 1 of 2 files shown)


helenye-stripe:
build seems to be failing

xavdid-stripe:
pesky toml tables! should be clear now

PR #1646, opened 2025-10-17
Prathmesh/merge python beta

prathmesh-stripe:
### Why?
Ran `just merge python-beta` and `just gen python --beta`.

### DIFF
--- a/stripe/v2/core/__init__.py
+++ b/stripe/v2/core/__init__.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+
+from typing_extensions import TYPE_CHECKING
 from stripe.v2.core._event import (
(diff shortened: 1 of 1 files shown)


stripe/v2/core/__init__.py
```diff
@@ -7,21 +7,61 @@
 )
 
 # The beginning of the section generated from our OpenAPI spec
-from stripe.v2.core import accounts as accounts, vault as vault
```
xavdid-stripe, line 10: Something went funny with the merge- you're missing a manual `from typing_extensions import TYPE_CHECKING` that's present in `master`

prathmesh-stripe: That was it.

PR #1669, opened 2025-11-11
Update v2 array parameter serialization to use indexed format

mbroshi-stripe:
### Why?
Stripe has updated how array parameters are handled in `GET` and  `DELETE` methods to `/v2` endpoints. Previously, the SDKs were serializing them with repeated parameter format (e.g., `?include=foo&include=bar`) since that is was the API expected. While that format is still accepted, it is deprecated. Instead, we will serialize array parameters to use indexed format (e.g., `?include[0]=foo&include[1]=bar`), which is the preferred method going forward. This aligns v2 behavior with v1 for consistency.

### What?
- Modified `stripe/_encode.py` to always use indexed array format for all API modes
- Removed `api_mode` argument to `_api_encode` function and all references to it
- Updated test expectations in `tests/test_http_client.arraysAsRepeated` to reflect new behavior

## Changelog
- `Retrieve` and `List` calls for `/v2` endpoints now use indexed format (e.g., `?include[0]=foo&include[1]=bar`) instead of repeated parameter format (e.g., `?include=foo&include=bar`) when communicating with the Stripe API. This may break any unit tests that expect the latter behavior when setting up a mock server. Instead, they should now expect the former.


stripe/_encode.py
```diff
@@ -27,26 +27,25 @@ def _json_encode_date_callback(value):
     return value
 
 
-def _api_encode(
-    data, api_mode: Optional[str]
-) -> Generator[Tuple[str, Any], None, None]:
+def _api_encode(data) -> Generator[Tuple[str, Any], None, None]:
```
mbroshi-stripe, line 30: Is this a breaking change? I'm assuming because it starts with a `_` we're OK, but I'm not too sure how we treat that from a semver point of view

jar-stripe: in python, an _ prefix conventionally means "private" and most linters will at least complain if someone is trying to access a method prefixed with an _.  I think this is not a breaking change here.

PR #1687, opened 2025-11-25
Fix DeprecationWarning when encoding StripeObject metadata (fixes #1651)

### LINKED ISSUE #1651: Direct update metadata causes DeprecationWarning
### Describe the bug

Our code is trying to update the metadata (without casting/converting to the `dict`). When we are `modify` the instance, the data is getting `encoded` using the internal sources and causes DeprecationWarning when it tries to check for `stripe_id` attribute on the object.

Because this is internal usage, the warning should not be raised.

### To Reproduce

1. Fetch an invoice with metadata filled
2. Get the metadata directly
3. Update metadata
4. Update the invoice with new metadata
5. Observe the DeprecationWarning

```
args = (<StripeObject at 0xffff6827fcf0> JSON: {
  "newkey": "value",
},), kwargs = {}

    @functools.wraps(__arg)
    def wrapper(*args, **kwargs):
>       warnings.warn(
            __msg, category=category, stacklevel=stacklevel + 1
        )
E       DeprecationWarning: For internal stripe-python use only. The public interface will be removed in a future version.
```

### Expected behavior

There should not be validation error happening while updating metadata

NOTE: the issue can be "silenced" by forcely casting the metadata to `dict` as it will not be StripeObject anymore and will not cause DeprecationWarning during encoding...

### Code snippets

```Python
invoice = stripe.Invoice.retrieve(...)
metadata = invoice.metadata or {}
metadata['newkey'] = 'value'
stripe.Invoice.modify(invoice.id, metadata=metadata)
```

### OS

Linux

### Language version

Python 3.13

### Library version

stripe-python v11.6.0

### API version

2017-08-15

### Additional context

We are using slightly obsolete versions... but I have checked main, the code for `encode.py` is the same and the check for attr `stripe_id` is there. I am pretty sure the same issue should happen on the latest versions.

praniketkw:
Fixes #1651

### Why?
When updating metadata on resources like invoices, users encountered a DeprecationWarning even though they were using the public API correctly. The warning was triggered internally when the encoder checked for the deprecated stripe_id property on StripeObject instances. While the deprecation warning serves an important purpose for direct user access to stripe_id, it should not be shown for internal library usage.

### What?
- Modified the _api_encode function in stripe/_encode.py to suppress DeprecationWarning when internally accessing the stripe_id property.
- Added warnings.catch_warnings() context managers around both the hasattr check and the property access to prevent warnings from bubbling up to users during internal encoding operations.
- Preserved the deprecation warning for direct user access to stripe_id while silencing it only for internal library usage.
- Updated tests to verify that no deprecation warnings are raised during encoding operations while maintaining all existing encoding behavior.
- The approach maintains the intended deprecation strategy while fixing the user experience issue where legitimate API usage triggered internal warnings.


stripe/_encode.py
```diff
@@ -31,8 +31,13 @@ def _api_encode(data) -> Generator[Tuple[str, Any], None, None]:
     for key, value in data.items():
         if value is None:
             continue
-        elif hasattr(value, "stripe_id"):
-            yield (key, value.stripe_id)
```
ramya-stripe, line 35: Hey @praniketkw 

https://github.com/stripe/stripe-python/pull/1101/ has the context around the deprecation warning that was added to the `stripe_id`. There were a a bunch of other attributes and methods that were also marked as deprecated in that PR because they all were meant for internal use. I would recommend using the `with warnings.catch_warnings():` clause to avoid the deprecation warning being bubbled to the user due to this internal usage.

The next time we ship a major version of the SDK, we will clean all of this up

praniketkw: Thanks for the context @ramya-stripe! That makes perfect sense. I understand the deprecation warning was intentionally added as part of the broader cleanup in #1101.
Thanks for the suggestion about using warnings.catch_warnings()! Given that this conflicts with the #1101 cleanup strategy, would you still like me to implement that approach?

ramya-stripe: Yes please go ahead
The broader cleanup will only get shipped end of March next year. So there is still value in you fixing this now

praniketkw: Thanks @ramya-stripe! I've updated the PR to use warnings.catch_warnings() as you suggested.

The fix now suppresses deprecation warnings only during internal usage in _api_encode() while preserving them for direct user access to stripe_id. This resolves the metadata update issue without interfering with the broader cleanup strategy.

I've tested it with the exact scenario from the issue and confirmed it works correctly.

ramya-stripe: Thank you!

ramya-stripe:
@praniketkw Can you also update the PR description based on the new approach?

praniketkw:
@ramya-stripe Done!

stripe/_encode.py
```diff
@@ -31,25 +32,35 @@ def _api_encode(data) -> Generator[Tuple[str, Any], None, None]:
...
```
ramya-stripe, line 35: Hey @praniketkw 

Just thought of an alternative with a smaller footprint. Can you try this and see if it works?

```
elif hasattr(value, "stripe_id")
   yield (key, getattr(value, "id"))
```

praniketkw: Thanks for the suggestion @ramya-stripe! I implemented your approach using getattr(value, "id") instead of value.stripe_id.

I had to wrap the hasattr(value, "stripe_id") check in warnings.catch_warnings() since hasattr itself triggers the deprecation warning. Also added a check for getattr(value, "id") is not None to handle edge cases properly.

The implementation now uses your approach while ensuring no warnings bubble up during internal encoding. Let me know if this looks good!

ramya-stripe: Oh! I was hoping that this approach could result in skipping the ignoring of warnings entirely. Adding both the warning and checking `hasattr(value, "id") and getattr(value, "id") is not None` is redundant.

How about replacing the original
```
elif hasattr(value, "stripe_id"):
            yield (key, value.stripe_id)
```
with
```
elif hasattr(value, "id"):
            yield (key, getattr(value, "id"))
```

Another question - any particular reason you are having the additional check of `getattr(value, "id") is not None` - this check was not there before

praniketkw: You're absolutely right @ramya-stripe! Thank you for the feedback! Your approach is much cleaner and avoids the warning entirely. I've implemented exactly what you suggested:

```python
elif hasattr(value, "id"):
    yield (key, getattr(value, "id"))
```
Regarding the ```getattr(value, "id") is not None``` check, you're correct that it wasn't in the original code. I added it thinking it would be safer, but I see now that it's unnecessary and changes the original behavior. I've removed it to match your suggestion exactly.

I really appreciate you taking the time to guide me!

ramya-stripe (approved):

ramya-stripe:
@praniketkw Thanks for your work on this. Can you update the PR description to reflect the final approach we have taken here?

boris-savic:
Hi. This merge broke certain projects, because you have changed the behavior. Before it has checked the objects `stripe_id`  property and now ti checks `id` - which can and indeed broke things. 

image

stripe/_encode.py
boris-savic: This change from stripe_id to id broke a lot of other projects where objects are being passed from database and the "id" is the database id, while the  "stripe_id" was stored in appropriate field.

ramya-stripe:
@boris-savic Internally, the `stripe_id` property reads from the `id` property. See https://github.com/praniketkw/stripe-python/blob/master/stripe/_stripe_object.py#L555 

So both these properties should have the same value.
Can you expand as to how this is breaking for you?

boris-savic:
In 14.1.0, if an object was passed in PaymentIntent.create for the `customer=some_object` the stripe library read the `stripe_id`  property stored in  `some_object.stripe_id`. After 14.2.0 this is no longer true as it reads the `some_object.id` and obviously breaks.

prathmesh-stripe:
@boris-savic I was looking at the [PaymentIntent.create API](https://docs.stripe.com/api/payment_intents/create?lang=python&architecture-style=services#create_payment_intent-customer) and the customer key requires a string and not an object.
Our types also ask the `customer` to be a string and not an object. 
Can you please provide an example with code sample of what you are trying to do just so that we can understand what breaks.

PR #1703, opened 2025-12-17
Update ci to run on Python 3.14 as well

ramya-stripe:
### Why?
Python 3.14 is now GA!

### What?
Add 3.14 to the matrix of Python versions we test for


mbroshi-stripe (approved):
Thanks!

PR #1713, opened 2026-01-06
Delete API_VERSION file as it is no longer needed

ramya-stripe:
### Why?
This file contained information about our code generator. The name `API_VERSION` was misleading. Since we now have a more accurately named file `CODEGEN_VERSION`, we can now delete this file


prathmesh-stripe:
Shouldn't this PR be built using the automerge tool so that the commit history lines up and we don't get conflicts in the generated code PRs?

jar-stripe (approved):

PR #1756, opened 2026-03-12
Add runtime support for V2 int64 string-encoded fields

mbroshi-stripe:
### Why?

V2 APIs represent int64 fields as JSON strings on the wire (e.g. `"12345"` instead of `12345`) to avoid precision loss. The SDK should transparently convert between native Python `int` and this wire format so users don't need to handle string encoding themselves.

### What?

- Added `_coerce_v2_params()` and `_coerce_value()` to `stripe/_encode.py` for request-side coercion (int → str), handling scalars, nested objects, and arrays
- Added `_field_encodings` class variable and `_coerce_field_value()` method to `StripeObject` for response-side coercion (str → int) during `_refresh_from`
- Added tests covering top-level, nested, array, nullable, and mixed-field scenarios for both request and response paths

### DIFF
--- a/stripe/_encode.py
+++ b/stripe/_encode.py
@@ -70,8 +70,13 @@ def _coerce_int64_string(value: Any, *, encode: bool) -> Any:
     to_type = str if encode else int
 
     if isinstance(value, list):
-        return [to_type(v) if isinstance(v, from_type) else v for v in value]
-    if isinstance(value, from_type):
+        return [
+            to_type(v)
+            if isinstance(v, from_type) and not isinstance(v, bool)
+            else v
+            for v in value
+        ]
+    if isinstance(value, from_type) and not isinstance(value, bool):
         return to_type(value)
     return value
 


stripe/_stripe_object.py
```diff
@@ -628,3 +631,23 @@ def _get_inner_class_type(
...
+        """
+        encoding = self._field_encodings.get(field_name)
+        if encoding is None or value is None:
+            return value
+
+        if encoding == "int64_string":
+            if isinstance(value, str):
+                return int(value)
+            if isinstance(value, list):
+                items: List[Any] = value
```
xavdid-stripe: do we need this line? type assertions don't have any runtime impact, so it doesn't seem like it's doing much

stripe/_encode.py
```diff
@@ -27,6 +27,65 @@ def _json_encode_date_callback(value):
...
+    for key, value in params.items():
+        field_schema = schema.get(key)
+        if field_schema is not None:
+            result[key] = _coerce_value(value, field_schema)
+        else:
+            result[key] = value
+    return result
+
+
+def _coerce_value(value: Any, schema: _SchemaNode) -> Any:
```
xavdid-stripe, line 84: why don't we use this function in `_coerce_field_value`?

xavdid-stripe (commented):
This makes sense, but will likely conflict with https://github.com/stripe/stripe-python/pull/1762 (at a git and/or logical level). Also I don't think your fancier function actually gets used, so either it's not needed or should be wired up (or I misread)

stripe/_stripe_object.py
mbroshi-stripe: Yep, good point. I was straight-up porting. I'll remove.

stripe/_encode.py
mbroshi-stripe: Fixed. `coerce_field_value` and `coerce_value` both delegate to `_coerce_int64_string`.

stripe/_encode.py
```diff
@@ -27,6 +27,80 @@ def _json_encode_date_callback(value):
...
+    if value is None:
+        return None
+
+    from_type = int if encode else str
+    to_type = str if encode else int
+
+    if isinstance(value, list):
+        return [to_type(v) if isinstance(v, from_type) else v for v in value]
+    if isinstance(value, from_type):
+        return to_type(value)
```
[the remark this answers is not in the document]
mbroshi-stripe: Fair point. Added guard for bools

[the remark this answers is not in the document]
mbroshi-stripe: The server should never return the wrong type. If it does, it's OK to fail loudly

stripe/_encode.py
```diff
@@ -27,6 +27,80 @@ def _json_encode_date_callback(value):
...
+        # Nested object schema
+        if isinstance(value, list):
+            # Array of objects with int64_string fields
+            return [
+                dict(_coerce_v2_params(v, schema) or {})
+                if isinstance(v, dict)
+                else v
+                for v in value
+            ]
+        if isinstance(value, dict):
```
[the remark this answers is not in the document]
mbroshi-stripe: This is OK in practice. We only have Mappings

PR #1790, opened 2026-04-04
Add Ref[T] type for typed v2 object references

jar-stripe:
### Why?

Custom objects need typed references to other Stripe objects. A reference has the wire shape `{type, id, url}` and in the SDK becomes `Ref[T]` — a container that knows the target type and can fetch it.

### What?

- `stripe.v2.Ref[T]` class (`Generic[T]`, not a StripeObject subclass)
- Three attributes: `type`, `id`, `url`
- Stores `_requestor: Optional[_APIRequestor]` (not StripeClient)
- `fetch()` calls `self._requestor.request("get", self.url, base_address="api")` — same pattern as `Event.fetch_related_object()`
- `fetch_async()` for async support
- `_construct_from` classmethod matching `StripeObject._construct_from` signature so the deserialization pipeline can instantiate Ref via `_inner_class_types`
- Requestor is threaded through `_convert_to_stripe_object` → `_construct_from` automatically
- Special handling in `_util.py` for `issubclass(klass, Ref)` to call Ref's `_construct_from` (which skips `last_response`)

### See Also

Generated with [Claude Code](https://claude.com/claude-code)


stripe/v2/_ref.py
```diff
@@ -0,0 +1,99 @@
...
+
+from typing_extensions import TYPE_CHECKING
+
+from stripe._api_mode import ApiMode
+
+if TYPE_CHECKING:
+    from stripe._api_requestor import _APIRequestor
+    from stripe._stripe_response import StripeResponse
+
+T = TypeVar("T")
```
xavdid, line 13: This should probably be `bound` to `StripeObject` because that's the root of everything that'll go in here.

stripe/v2/_ref.py
```diff
@@ -0,0 +1,99 @@
...
+
+    url: str
+    """The URL path that can be used to fetch the referenced resource."""
+
+    _requestor: Optional["_APIRequestor"]
+
+    def __init__(
+        self,
+        parsed_body: Dict[str, Any],
+        requestor: Optional["_APIRequestor"],
```
xavdid, line 39: same as ruby - in what situations is the `Ref` useful when the requestor isn't present `None`?

stripe/v2/_ref.py
```diff
@@ -0,0 +1,99 @@
...
+from stripe._api_mode import ApiMode
+
+if TYPE_CHECKING:
+    from stripe._api_requestor import _APIRequestor
+    from stripe._stripe_response import StripeResponse
+
+T = TypeVar("T")
+
+
+class Ref(Generic[T]):
```
xavdid, line 16: Similar to ruby, we _probably_ want this to be a `StripeObject`?

xavdid (commented):
Similar comments to the other langs

jar-stripe:
Changing direction for now.

PR #1810, opened 2026-05-14
Add WebhookSignature.generate_test_header_string helper

### LINKED ISSUE #1697: Provide a public helper to sign webhook data
### Is your feature request related to a problem? Please describe.

The Node SDK provides a [`generateTestHeaderString()` utility](https://github.com/stripe/stripe-node?tab=readme-ov-file#testing-webhook-signing) to manually sign payloads for testing. It would be great to have a similar utility in the Python SDK.

We currently copy over this [internal test fixture](https://github.com/stripe/stripe-python/blob/785e75f96fc81deb0342a4521200709a4c697f05/tests/test_webhook.py#L18-L30), but the signing logic is an internal detail that can change at any point in time.

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

vishwakt:
## Summary

Adds a public `WebhookSignature.generate_test_header_string()` helper that mirrors stripe-node's [`generateTestHeaderString()`](https://github.com/stripe/stripe-node?tab=readme-ov-file#testing-webhook-signing), so users can sign test payloads without reaching for the underscore-prefixed `_compute_signature` internal.

Fixes #1697.

### Implementation notes
- New method lives on `WebhookSignature` (next to `EXPECTED_SCHEME`, `_compute_signature`, and `verify_header`), accessed as `stripe.WebhookSignature.generate_test_header_string(...)`.
- Python `Optional` keyword arguments are used instead of the Node SDK's options object, matching Python conventions while preserving the full Node parameter surface (`payload`, `secret`, `timestamp`, `scheme`, `signature`).
- Defaults: `timestamp` → `int(time.time())`, `scheme` → `EXPECTED_SCHEME`, `signature` → computed from `payload` + `secret`. Passing `signature` explicitly inserts it verbatim, mirroring Node.
- The internal `generate_header` helper in `tests/test_webhook.py` was refactored to delegate to the new public method. This proves the new API covers every existing test scenario and removes the duplicated signing logic from tests.
- New `TestGenerateTestHeaderString` class covers: defaults, round-trip through `verify_header`, custom timestamp/scheme, signature pass-through, and bad-secret failure.

### Example usage
```python
import stripe

header = stripe.WebhookSignature.generate_test_header_string(
    payload=payload_json,
    secret="whsec_test_secret",
)
event = stripe.Webhook.construct_event(payload_json, header, "whsec_test_secret")
```

## Test plan
- [x] `just test` — 4245 passed, 6 skipped
- [x] `just lint` — clean
- [x] `just format-check` — clean
- [x] `just typecheck` — only pre-existing error in `stripe/_http_client.py:681` (present on `master`, unrelated to this change)


stripe/_webhook.py
```diff
@@ -52,6 +53,38 @@ def _compute_signature(payload, secret):
...
+    def generate_test_header_string(
+        cls,
+        payload: str,
+        secret: str,
+        timestamp: Optional[int] = None,
+        scheme: Optional[str] = None,
+        signature: Optional[str] = None,
+    ) -> str:
+        """
+        Generates a value for the `Stripe-Signature` header that can be used
```
Viicos, line 52: The docstring uses a mix of markdown and RST syntax. I would suggest sticking to markdown to be consistent with the rest of the code base. You can specify parameters as:

```
Args:
    payload: ...
    secret: ...
```

vishwakt: switched to a markdown `Args:` block. Thanks for the catch!

stripe/_webhook.py
```diff
@@ -52,6 +53,40 @@ def _compute_signature(payload, secret):
         )
         return mac.hexdigest()
 
+    @classmethod
+    def generate_test_header_string(
```
mbroshi-stripe, line 44: Question: Did you consider `Webhook.generate_test_header_string`? stripe-node puts the helper on the primary `webhooks` object. I'm curious why you chose `WebhookSignature` instead of `Webhook`

vishwakt: You're right that `Webhook` is the better home. My original reasoning was internal cohesion (co-locating with `_compute_signature` and `verify_header`), but that's outweighed by:

- **Cross-SDK consistency.** stripe-node, stripe-go, and stripe-ruby all put their test header helper on the primary webhooks object, so a user hopping between SDKs knows where to look.
- **Discovery via use case.** The helper exists to feed `Webhook.construct_event` in tests; sitting next to it is a stronger signal than sitting next to the primitive it wraps.
- **Public-surface symmetry.** `Webhook` is the user-facing entry point; `WebhookSignature` is closer to plumbing. A *testing* helper belongs on the surface being tested.

Moved it to `Webhook.generate_test_header_string`, with the implementation still delegating to `WebhookSignature._compute_signature` internally. Tests updated accordingly. Thanks for pushing back on this.

jar-stripe:
Hi @vishwakt just wanted to give a quick update here.  We're pretty backed up at the moment and are doing our best to triage internal and external reviews along with our normal work.  Which is to say, we have not forgotten about you and very much appreciate your patience here!

xavdid:
Thank you for the contribution! As luck (or not) would have it, this got swept up into some other events-related work I've been doing (https://github.com/stripe/stripe-python/pull/1855), so I'll close this one out. Apologies for the hassle!

PR #1813, opened 2026-05-21
Emit warning when `stripe-notify` header is present in response

xavdid:
### Why?

We've noticed a trend where users (and their agents) will sometimes reach for suboptimal API endpoints/methods because of outdated information. For example, using `/v1/accounts` when `/v2/core/accounts` is available and has a superset of the functionality from v1.

In an effort to help steer users towards best practices, we're introducing a new header to provide feedback during the development process. The SDK is responsible for surfacing this header, if present.

Once enabled internally, the header will be returned by the server for:

- all testmode calls
- livemode calls made by an agent

We may tweak that going forwards. But no matter the conditions, the SDKs will act as a thin pipe.

### What?

- Emit warning when we see the `Stripe-Notice` header in a response
- add tests

### See Also

- https://go/agent-steering
- [DEVSDK-2430](https://go/j/browse/DEVSDK-2430)

### DIFF
--- a/stripe/_api_requestor.py
+++ b/stripe/_api_requestor.py
@@ -73,7 +73,8 @@ _default_proxy: Optional[str] = None
 
 
 def _maybe_emit_stripe_notice(rheaders: Mapping[str, str]) -> None:
-    if notice := rheaders.get("Stripe-Notice"):
+    notice = rheaders.get("Stripe-Notice")
+    if notice:
         import warnings
 
         warnings.warn(notice)
--- a/stripe/_api_requestor.py
+++ b/stripe/_api_requestor.py
@@ -72,6 +72,13 @@ HttpVerb = Literal["get", "post", "delete"]
 _default_proxy: Optional[str] = None
 
 
+def _maybe_emit_stripe_notice(rheaders: Mapping[str, str]) -> None:
+    if notice := rheaders.get("Stripe-Notice"):
+        import warnings
+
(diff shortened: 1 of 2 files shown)


stripe/_api_requestor.py
```diff
@@ -72,6 +72,13 @@
 _default_proxy: Optional[str] = None
 
 
+def _maybe_emit_stripe_notice(rheaders: Mapping[str, str]) -> None:
+    if notice := rheaders.get("Stripe-Notice"):
```
prathmesh-stripe: [Typechecker](https://github.com/stripe/stripe-python/actions/runs/26259580131/job/77289878164?pr=1813#step:8:50) is failing in CI. `Operator ":=" requires Python 3.8 or newer`. We'll have to update the the list of minor versions we test against here: https://github.com/stripe/stripe-python/blob/41a2ece934c8d365b9fc88253bc3c4675913d09e/.github/workflows/ci.yml#L49

prathmesh-stripe (changes requested):

stripe/_api_requestor.py
xavdid: fixed! i've opened [RUN_DEVSDK-2440](https://go/j/RUN_DEVSDK-2440) to do that separately

ksealmalaysia41-lang (commented):
what happen

prathmesh-stripe (approved):

PR #1814, opened 2026-05-22
Only typecheck against a single version

xavdid:
### Why?

We used to typecheck a range of versions, but this wasn't actually adding much coverage. Ideally, we just check our types against a single modern version, since we get stricter checks. Incompatible syntax will be caught by unit tests, not type cehcks.

### What?

- remove the range and just check against a modern python version
- remove arguments from `typecheck` justfile recipe
- bump version we typecheck against

### See Also

https://github.com/stripe/stripe-python/pull/1764


prathmesh-stripe (approved):
It would've been nicer to pick this from our list of supported versions but this is fine for now.

prathmesh-stripe:
There are some static checks that are [failing](https://github.com/stripe/stripe-python/actions/runs/26313203201/job/77466580804?pr=1814#step:8:62) in Python 3.13 that we need to fix before we merge this.

stripe/_http_client.py
```diff
@@ -149,7 +149,7 @@ def __init__(
         self._verify_ssl_certs = verify_ssl_certs
         if proxy:
             if isinstance(proxy, str):
-                proxy = {"http": proxy, "https": proxy}
+                proxy = HTTPClient._Proxy(http=proxy, https=proxy)
```
xavdid, line 152: this is more correct than the inline dict, which the typechecking on 3.13 caught!

xavdid:
@prathmesh-stripe good call! i've updated CI to only check against a single (modern) version. 

How does that look?

PR #1822, opened 2026-06-02
Disable trim_trailing_whitespace for CHANGELOG.md

jar-stripe:
Prevents editors from stripping intentional trailing whitespace in changelog entries.

Fixes: https://go/j/DEVSDK-2229

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,3 +15,6 @@ indent_size = 2
 
 [Makefile]
 indent_style = tab
+
+[CHANGELOG.md]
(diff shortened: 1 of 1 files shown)


xavdid (approved):
IMO the solution here is really to trim all the whitespace once (and honestly just run it through `prettier`) and then ignore that rev for git-blame purposes. Not necessarily blocking, but is an alternative (and potentially better, long-term) solution

xavdid (approved):

PR #1847, opened 2026-07-13
move publish job to trusted publishing

xavdid:
### Why?

Our current release process uses a long-lived PyPI token to upload releases. While still supported, it's no longer the favored way to do Python releases because if that token leaks, attackers can publish your package.

The current best practice is "trusted publishing", where github communicates with PyPI to mint a short-lived token with publishing permissions on demand. We start a release, get a token, use that token for the release, then revoke it. 

The process is straightforward. We make these changes to our CI (using the official GH action for python publishing) and then log into PyPI and hit a checkbox to enable the feature. After that, publishing will work like it always had and we'll be able to delete the `TWINE_PASSWORD` secret.

### What?

- changed CI workflow to use the standard trusted publishing action
  - follows the standard practices and uses the required fields, such as `environment` and `id-token`

### See Also

- https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/
- https://github.com/pypa/gh-action-pypi-publish

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,34 +111,20 @@ jobs:
       startsWith(github.ref, 'refs/tags/v')
     needs: [build, test, lint]
     runs-on: "ubuntu-24.04"
+    environment: publish
     permissions:
       contents: read
+      id-token: write
     steps:
-      - uses: extractions/setup-just@v2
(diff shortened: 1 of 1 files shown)


jar-stripe (approved):
LG!

PR #1853, opened 2026-07-23
Fix intermittent Windows test flake in async httpx integration tests

jar-stripe:
### Why?
BaseHTTPRequestHandler defaults to protocol_version "HTTP/1.0", which sets close_connection=True after every response. On Windows, the socket close can race with the async httpx client reading the response body, causing ReadError (Windows discards buffered TCP data on RST more aggressively than Linux).

Setting protocol_version to "HTTP/1.1" keeps the connection alive, eliminating the race. The 30s timeout is a backstop against the handler blocking on readline if the client never closes the connection.

Committed-By-Agent: claude

### What?
- Sets the `BaseHTTPRequestHandler` protocol version to HTTP/1.1, and timeout to 30s
- Updates tests to not assume we're dropping the connection between requests

### DIFF
--- a/tests/test_integration.py
+++ b/tests/test_integration.py
@@ -26,6 +26,19 @@ from http.server import BaseHTTPRequestHandler, HTTPServer
 MOCK_HOST = os.environ.get("STRIPE_MOCK_HOST", "localhost")
 
 
+class RequestSnapshot:
+    """Snapshot of request attributes, captured at handle time.
+
+    With HTTP/1.1 keep-alive, multiple requests reuse the same handler
+    instance whose attributes get overwritten on each request.
+    """
+
+    def __init__(self, handler):
+        self.command = handler.command
+        self.path = handler.path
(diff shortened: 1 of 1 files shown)


tests/test_integration.py
```diff
@@ -26,18 +26,34 @@
...
 
 
+class RequestSnapshot:
+    """Snapshot of request attributes, captured at handle time.
+
+    With HTTP/1.1 keep-alive, multiple requests reuse the same handler
+    instance whose attributes get overwritten on each request.
+    """
+
+    def __init__(self, handler):
```
xavdid: nit: add a type annotation to `handler`

xavdid (approved):
I'm not super familiar with this code, but it seems reasonable!

tests/test_integration.py
jar-stripe: Done!

PR #1885, opened 2026-08-22
Add `.pre_handle` method to `EventNotificationHandler`

xavdid:
### Why?

In final testing, users have expressed interest in being able to centrally run code before any other handler executes. It can be used for centralized logging, event deduplication, and more. I don't want to go full middleware, but this felt like a reasonable way to dip our toe into the waters.

Because users are responsible for [deduplicating events](https://docs.stripe.com/webhooks#handle-duplicate-events), this pre-handle hook (if present) should return a `bool`, signifying whether handling should continue.

If the function returns `true` (or is missing entirely), the handler proceeds like normal (running at-most-one other callback). If it returns `false`, no further callback is run and the original `.handle()` call finishes cleanly.

### What?

- add the `.pre_handle` method
- & associated tests
- update some error messages with correct verbs & consistency
- refactored some internal error handling
- updated example

### See Also

- http://go/j/DEVSDK-3249

### DIFF
--- a/examples/event_notification_handler_endpoint.py
+++ b/examples/event_notification_handler_endpoint.py
@@ -49,9 +49,9 @@ unverified_handler = client.notification_handler_without_verification(
 @unverified_handler.pre_handle
 def deduplicate_events(notif: EventNotification, client: StripeClient) -> bool:
     """
-    Runs before any registered handler or fallback callback. Returning False
+    Runs before any registered callback. Returning False
     here skips handling entirely for this delivery, which is useful for
-    deduplicating retried webhooks.
+    deduplicating webhooks.
     """
     if notif.id in processed_event_ids:
         print(f"Already processed {notif.id}, skipping.")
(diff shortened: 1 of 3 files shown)


examples/event_notification_handler_endpoint.py
```diff
@@ -39,6 +45,22 @@ def fallback_callback(
 )
 
 
+@handler.pre_handle
```
jar-stripe, line 48: my only concern is this assumes handler and unverified_handler are accessible at the scope of the function declaration, e.g. if it is a class member variable.  consider having another way e.g. `handler.register_pre_handle(...)` to also set the pre_handle callback

jar-stripe (approved):
Approved to unblock, but see the comment

examples/event_notification_handler_endpoint.py
xavdid: we chatted- this is working as intended, but i've updated some docs

PR #1887, opened 2026-08-25
add async event handler methods

xavdid:
### Why?

Previously the (as of yet unreleased in GA) event handlers only accepted synchronous callbacks. But given the popularity of async Python, it made sense to support async methods too. This means 2 new classes with a bunch of internal code reuse. I've added some decent module-level docstrings.

See the new example for full usage.

### What?

- refactor handler classes & add async variants
- add related client methods
- add tests
- add docs
- add example

### See Also

- https://github.com/stripe/stripe-python/pull/1881
- https://github.com/stripe/stripe-python/pull/1885
- [DEVSDK-3228](https://go/j/DEVSDK-3228)


jar-stripe (approved):

kattgrom-hub (commented):
A

PR #1888, opened 2026-08-25
widen some webhook types

xavdid:
### Why?

We were a little inconsistent with what string-like types our (now many) webhook methods took. So first off, we should be consistent. But, this also means we have a better integration with common web frameworks, because we natively accept (and handle) their `request.body` type without the user having to potentially decode! Same goes for a potentially missing header- we'll accept it and throw an error vs the user having to verify it themselves.

### What?

- ensure all webhook-related methods accept the new union types of `str | bytes | bytearray`
- ensure signature headers are typed as optional, but note that they throw errors when not present.
- many tests

## See also

- [DEVSDK-3228](https://go/j/DEVSDK-3228)

### DIFF
--- a/examples/async_event_notification_handler_endpoint.py
+++ b/examples/async_event_notification_handler_endpoint.py
@@ -95,8 +95,7 @@ async def webhook(request: Request):
 async def webhook_from_cloud_provider(request: Request):
     # no signature header to pass along; the channel already authenticated this event
     try:
-        body = await request.body()
-        await unverified_handler.handle_async(body.decode())
+        await unverified_handler.handle_async(await request.body())
         return Response(status_code=200)
     except Exception as e:
         return Response(content=str(e), status_code=500)
--- a/examples/async_event_notification_handler_endpoint.py
+++ b/examples/async_event_notification_handler_endpoint.py
@@ -95,7 +95,8 @@ async def webhook(request: Request):
 async def webhook_from_cloud_provider(request: Request):
     # no signature header to pass along; the channel already authenticated this event
     try:
-        await unverified_handler.handle_async(await request.body())
+        body = await request.body()
+        await unverified_handler.handle_async(body)
         return Response(status_code=200)
     except Exception as e:
         return Response(content=str(e), status_code=500)
--- a/examples/event_notification_handler_endpoint.py
+++ b/examples/event_notification_handler_endpoint.py
@@ -76,7 +76,9 @@ def handle_meter_error(
 @app.route("/webhook", methods=["POST"])
 def webhook():
     try:
-        handler.handle(request.data, request.headers.get("Stripe-Signature"))
+        webhook_body = request.data
+        sig_header = request.headers.get("Stripe-Signature")
+        handler.handle(webhook_body, sig_header)
         return jsonify(success=True), 200
     except Exception as e:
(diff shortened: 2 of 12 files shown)


examples/async_event_notification_handler_endpoint.py
```diff
@@ -95,8 +95,7 @@ async def webhook(request: Request):
 async def webhook_from_cloud_provider(request: Request):
     # no signature header to pass along; the channel already authenticated this event
     try:
-        body = await request.body()
-        await unverified_handler.handle_async(body.decode())
```
jar-stripe, line 99: small nit; think the original code that used a body variable is a bit cleaner, even its then just passed into the handler, but it's a very small nit.

jar-stripe (commented):
What was the issue that lead to wanting to make sig_header optional?

xavdid:
It came from some dogfooding I was doing on the handlers. Basically, when using flask starter code, a user with a type checker sees a bunch of errors:

```py
endpoint_secret = os.environ.get("STRIPE_WEBHOOK_SECRET")

client = stripe.StripeClient("sk_test_123")

# ERR: Argument of type "str | None" cannot be assigned to parameter "webhook_secret" of type "str"
handler = client.notification_handler(endpoint_secret, lambda x, y, z: None)

app = Flask(__name__)


@app.route("/webhook", methods=["POST"])
def webhook():
    # bytes; we purposefully accept them in `parse_event_notification` 
    # so user doesn't have to think about conversion (even if that should really only accept a `str`)
    payload = request.data 
    sig_header = request.headers.get("stripe-signature")

    try:
        # ERR: Argument of type "str | None" cannot be assigned to parameter "sig_header" of type "str"
        # ERR: Argument of type "str | None" cannot be assigned to parameter "secret" of type "str"
        event = client.parse_event_notification(payload, sig_header, endpoint_secret)
    except stripe.error.SignatureVerificationError as e:
        print("⚠️  Webhook signature verification failed." + str(e))
        return jsonify(success=False)

    ...

    return jsonify(success=True)
```

Now, all those errors are _correct_. `os.environ.get` could return `None` if the value isn't set, as can `request.headers.get`. But if a user knows they're set up correctly, they still have to add extra code to appease the type checker. Either we make _every_ user do all of the null checking themselves, or we widen our types and do that null-checking correctly for everyone.

We do a similar thing in the node webhook code: [stripe/stripe-node | Webhooks.ts#L10-L20](https://github.com/stripe/stripe-node/blob/cd421c5a7c19228fa80341e63d46fbe8e7d39555/src/Webhooks.ts#L10-L20). Basically we accept types that will throw errors at runtime because it means you can pass data directly from `express` into the SDK without doing additional checks. We know that a stripe webhook will only ever have one `stripe-signature` header value, so it _is_ a string at runtime (not a `string[]`). Either we make every user verify that, or we assume the server value is correct (and throw an error if we get unexpected data).

And, existing code / defensive coders can still do the same assertions up front/if they want to handle missing values. But for people who just want it to work, this is a smoother integration experience.

examples/async_event_notification_handler_endpoint.py
xavdid: fair! i can revert.

jar-stripe (approved):

PR #1896, opened 2026-08-28
Harden API requestor code against malicious URLs

xavdid:
### Why?

If an attacker got access to a user's webhook secret (or the user wasn't validating incoming webhook signatures) _and_ a user was calling the built-in `event_notification.fetch_related_object()` method, a malicious webhook payload could cause a user integration to make an authenticated request to an attacker-controlled domain (leaking the secret key)

To solve, we're being more careful during url construction to ensure user input can't be used to send requests to non-`stripe.com` domains (on a per-request basis).

### What?

- add path validation method and call it before the final URL is built
- Encode the `event.id` path param in `fetch_event()` to match the sanitation in the rest of the generated endpoints
- protect against treating malicious payloads as v2 iterables
- add tests

### See Also

- [RUN_DEVSDK-2808](https://go/j/RUN_DEVSDK-2808)

### DIFF
--- a/stripe/_util.py
+++ b/stripe/_util.py
@@ -355,18 +355,16 @@ def validate_path(path: str) -> None:
     Assert that a request path is origin-relative: that it begins with a single
     "/" and carries no scheme, authority or userinfo.
 
-
-
     The absolute URL is built by concatenating a base address onto this path, and
     no base address ends in a slash. A path like "@evil.example/v1/x" or
-    ".evil.example/v1/x" would therefore land inside the authority component and
-    send the request (Authorization header included) -- to a host of the path's
-    choosing. Some request paths originate in remote data (a webhook body's
-    related_object.url, a response's next_page_url), so the path cannot be
-    assumed to be well-formed.
+    ".evil.example/v1/x" would modify the resulting host and direct the request
+    (including the API key) to a non-Stripe host.
 
-    Stripe only ever issues plain paths, so anything else is tampering and is
-    rejected rather than sanitized.
+    Because some relative urls arrive from potentially untrusted sources (like
+    webhook bodies), we have to be a little defensive.
+
+    So, we require that a path starts with a leading slash and that urlsplit
+    finds no scheme or authority in it.
     """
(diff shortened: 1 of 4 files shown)


stripe/v2/core/_event.py
```diff
@@ -229,7 +229,7 @@ def fetch_event(self) -> Event:
     async def fetch_event_async(self) -> Event:
         response = await self._client.raw_request_async(
             "get",
-            f"/v2/core/events/{self.id}",
+            f"/v2/core/events/{sanitize_id(self.id)}",
             stripe_context=self.context,
             headers={"Stripe-Request-Trigger": f"event={self.id}"},
```
jar-stripe, line 234: does this need sanitization as well?

stripe/_util.py
```diff
@@ -346,6 +350,34 @@ def sanitize_id(id):
     return quotedId
 
 
+def validate_path(path: str) -> None:
+    """
+    Assert that a request path is origin-relative: that it begins with a single
```
jar-stripe, line 355: this feels like a lot of exposition for validate_path.  is there a better place to put this?

stripe/_util.py
```diff
@@ -346,6 +350,34 @@ def sanitize_id(id):
...
+    "/" and carries no scheme, authority or userinfo.
+
+    The absolute URL is built by concatenating a base address onto this path, and
+    no base address ends in a slash. A path like "@evil.example/v1/x" or
+    ".evil.example/v1/x" would modify the resulting host and direct the request
```
jar-stripe, line 360: this is pretty verbose for what validate_path does; consider if this info should live somewhere else, or if it can be pared down.

jar-stripe (approved):

stripe/v2/core/_event.py
xavdid: no, because it can't change the destination of the url or escape the `Stripe-Request-Trigger`. It could send us a bad value, but that would just mess up internal traceability for this event rather than have user impact

PR #1898, opened 2026-08-29
Use cryptographically secure boundaries for multipart file uploads

xavdid:
### Why?

Fixes a bug where the boundary markers in multi-part file uploads weren't cryptographically secure. As a result, it could be possible for an attacker-controlled file to specify additional properties in the upload request.

Similarly, a maliciously crafted filename could be used to add additional properties to the request.

This only affects the `files.create` method.

### What?

- use a cryptographically secure method of generating multipart file boundaries
- sanitize filename input before adding to requests
- add tests

### See Also

- [RUN_DEVSDK-2807](https://go/j/RUN_DEVSDK-2807)

### DIFF
--- a/stripe/_multipart_data_generator.py
+++ b/stripe/_multipart_data_generator.py
@@ -1,13 +1,29 @@
-import random
 import io
+import secrets
 
 from stripe._encode import _api_encode
+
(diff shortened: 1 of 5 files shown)


jar-stripe (approved):
LG!

PR #1920, opened 2026-09-18
Apply positional-only service path parameters, and update CHANGELOG with changes

mark-kogan-stripe:
### Why?
Service-method path parameters have been marked as positional-only. This changelog entry and migration guide communicate the breaking change to all users

### What?
Applied positional-only special parameter to service-path functions.

## Configuration

- [ ] `skip-changefile`: This PR is not a user-facing change, so there's no changefile.


.hark/migration-guides/v16.md
```diff
@@ -20,3 +20,21 @@ response = stripe_object.request("get", "/v1/example")
 client = stripe.StripeClient("sk_test_...")
 response = client.raw_request("get", "/v1/example")
 ```
+
+## Keyword arguments are no longer allowed in Stripe service methods
+
+Path parameters in all service methods must now be passed positionally instead of by keyword. This prevents parameter names derived from the API specification from becoming part of the public interface.
```
jar-stripe: this may be a nit but "This prevents parameter names derived from the API specification from becoming part of the public interface." feels unnecessary as part of the migration guide.  If we wanted to motivate the change, include that motivation in the .change.md file.

jar-stripe (commented):
See the attached comment.  I think it reads cleaner to include motivation in the changelog instead of the migration guide.

jar-stripe (approved):
Change looks good!  Please fix the PR description (it looks like some of the title may have bled into the top of the description) and then merge away!

